PR #184, opened 2019-07-11
Upgrade Typescript to 3.5

InventingWithMonster:


src/motion/functionality/dom.tsx
```diff
@@ -144,7 +144,7 @@ export function createDomMotionConfig<P>(
             style,
             isStatic
         ) => {
-            const activeComponents: ReactElement<P>[] = []
+            const activeComponents: ReactElement<any>[] = []
```
eelco: This looks like a weakening of type strictness. Can we avoid that?

InventingWithMonster: Yeah I've changed it to be a bit better, but it wasn't correct before

eelco: Interesting, I also managed to fix it by upgrading the React types, it’s a bit of PR creep, but would you like me to push that?

PR #186, opened 2019-07-11
Don't attempt to run useLayoutEffect when SSRing

### LINKED ISSUE #179: [BUG] useLayoutEffect warning for positionTransition
When rendering components with `positionTransition` server-side, React complains about `useLayoutEffect`.

We're not doing anything that would affect the first render but we do need `useLayoutEffect` for this specific use-case.

As this functionality is loaded in via a renderless component, we can add the presence of `window` to the check as to whether or not this component should be rendered.

Andarist:
fixes #179

### DIFF
--- a/src/motion/functionality/position.ts
+++ b/src/motion/functionality/position.ts
@@ -93,7 +93,8 @@ function usePositionAnimation(
 }
 
 export const Position: FunctionalComponentDefinition = {
-    shouldRender: (props: MotionProps) => !!props.positionTransition,
+    shouldRender: (props: MotionProps) =>
+        typeof window !== "undefined" && !!props.positionTransition,
     component: makeHookComponent(
         ({
             innerRef,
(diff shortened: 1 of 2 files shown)


src/motion/functionality/position.ts
```diff
@@ -101,12 +101,11 @@ export const Position: FunctionalComponentDefinition = {
             values,
             positionTransition,
         }: FunctionalProps) => {
-            return usePositionAnimation(
-                innerRef,
-                values,
-                controls,
-                positionTransition
-            )
+            if (typeof window === "undefined") {
```
InventingWithMonster: This check might live better on line 96?

```
shouldRender: (props: MotionProps) => !!props.positionTransition && typeof window !== "undefined",
```

?

Andarist: good point, changed

InventingWithMonster:
Awesome thanks for this!

PR #187, opened 2019-07-11
Validate AnimatePresence keys - warn if they are not unique

### LINKED ISSUE #166: Add error warning for multiple children of `AnimatePresence` with the same key.
Currently `AnimatePresence` doesn't warn if a component doesn't have a valid key. This is to allow toggling the visibility of single components without needing one.

We can either:
1) Require even for single-child examples and throw
2) Detect multiple of the same key if we have more than one child and throw

2 would miss the `Slideshow`- esque use-case where we have a single child with a changing key, but would enable the toggling of visibility of a single child without a key. I am leaning towards 1.

Andarist:
fixes #166

### DIFF
--- a/src/components/AnimatePresence.tsx
+++ b/src/components/AnimatePresence.tsx
@@ -79,8 +79,22 @@ function updateChildLookup(
     children: ReactElement<any>[],
     allChildren: Map<ComponentKey, ReactElement<any>>
 ) {
+    const seenChildren =
+        process.env.NODE_ENV !== "production" ? new Set<ComponentKey>() : null
+
     children.forEach(child => {
         const key = getChildKey(child)
+
+        if (process.env.NODE_ENV !== "production" && seenChildren) {
+            if (seenChildren.has(key)) {
+                console.warn(
+                    `Children of AnimatePresence require unique keys. "${key}" is a duplicate.`
+                )
+            }
+
+            seenChildren.add(key)
+        }
+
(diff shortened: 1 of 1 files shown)


src/components/AnimatePresence.tsx
```diff
@@ -79,8 +79,20 @@ function updateChildLookup(
     children: ReactElement<any>[],
     allChildren: Map<ComponentKey, ReactElement<any>>
 ) {
+    const seenChildren =
+        process.env.NODE_ENV !== "production" ? new Set<ComponentKey>() : null
```
Andarist: those env checks look slightly weird but they provide better DCE in production builds

src/components/AnimatePresence.tsx
```diff
@@ -79,8 +79,20 @@ function updateChildLookup(
...
+
     children.forEach(child => {
         const key = getChildKey(child)
+
+        if (process.env.NODE_ENV !== "production" && seenChildren) {
+            if (seenChildren.has(key)) {
+                console.warn("AnimatePresence's children need unique keys.")
```
InventingWithMonster:
```
console.warn(`Children of AnimatePresence require unique keys. ${key} is a duplicate.`)
```

Andarist: adjusted

InventingWithMonster:
Nice one thanks!

PR #189, opened 2019-07-11
More accurate type definition for SVG elements in `motion` object

Cryrivers:
This PR provides more accurate props definition for SVG elements in `motion`.

Before this PR, the ref of `<motion.text />` is typed as `SVGElement`
Screenshot 2019-07-12 at 12 18 11 AM

In this PR, the ref of `<motion.text />` is typed as `SVGTextElement`
Screenshot 2019-07-12 at 12 21 03 AM

### DIFF
--- a/api/framer-motion.api.md
+++ b/api/framer-motion.api.md
@@ -192,61 +192,61 @@ export type KeyframesTarget = ResolvedKeyframesTarget | [null, ...CustomValueTyp
 
 // @public
 export const motion: {
-    symbol: ForwardRefExoticComponent<SVGMotionProps & RefAttributes<SVGElement>>;
-    circle: ForwardRefExoticComponent<SVGMotionProps & RefAttributes<SVGElement>>;
-    clipPath: ForwardRefExoticComponent<SVGMotionProps & RefAttributes<SVGElement>>;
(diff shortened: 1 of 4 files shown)


Cryrivers:
`SVGElement` lacks methods of which the actual SVG elements have, for example, `SVGElement` doesn't have `getBBox` method.

Without this PR, we need to declare the React ref as `SVGElement` and do an unnecessary type narrowing, otherwise, TypeScript will either complain the type of React ref is not compatible or `SVGElement` doesn't have some methods.

With this PR, most likely everything works with intrinsic SVG elements should work with the `motion` counterparts.

InventingWithMonster:
Ah I see. So before a ref typed `SVGTextElement` passed to `motion.text` would throw? If so it'd be great to get a test in https://github.com/framer/motion/blob/master/src/motion/__tests__/component-svg.test.tsx with this scenario that would have failed before and now passes.

Cryrivers:
yeah correct. I’ll add tests later

Cryrivers:
added the test with the scenario that would have failed previously, please take a look :)

InventingWithMonster:
Excellent, thanks!

PR #195, opened 2019-07-13
feat(use-viewport-scroll): support passing element ref

### LINKED ISSUE #182: [FEATURE] use closest scroll container in useViewportScroll
`useViewportScroll` only works with the `body` element.

I'd like if `useViewportScroll` could optionally accept a ref to an element and use the closest scroll container.

I noticed that the docs suggest rolling this yourself, but I think it could be helpful to provide out of the box.

---

I'd be happy to submit a PR with this functionality, please let me know 😊.

souporserious:
This was a quick take on adding `ref` support to `useViewportScroll`, curious your thoughts on everything. Still new to Typescript, so I'm sure I'm missing some things 😇. Please comment on any concerns and I can address them. I can also update documentation once there's some finalization and take a stab at adding tests.

Some questions I had while implementing this:
- Should we probably warn if `ref.current` is available but is not `overflow: scroll`?
- Another solution could be to just grab the scroll closest scroll container, I have a small function from another library I can add to implement this
- We should probably handle only adding one listener per element/window similar to what you had originally
- Can we utilize `useResize` hook and implement `ResizeObserver` inside that?

I understand if you don't want to add `ResizeObserver` support. Although, I think it could be ok since most modern browsers will support this in their [next release](https://caniuse.com/#feat=resizeobserver). We could also add docs and point to the polyfill since older browser support is probably needed. 
 
Closes #182

### DIFF
--- a/src/value/use-viewport-scroll.ts
+++ b/src/value/use-viewport-scroll.ts
@@ -1,6 +1,6 @@
 import { invariant } from "hey-listen"
 import { useLayoutEffect, RefObject } from "react"
-import { MotionValue } from "../value"
+import { motionValue, MotionValue } from "../value"
 import { useMotionValue } from "./use-motion-value"
 
 export interface ScrollMotionValues {
@@ -8,6 +8,29 @@ export interface ScrollMotionValues {
     scrollY: MotionValue<number>
     scrollXProgress: MotionValue<number>
     scrollYProgress: MotionValue<number>
+}
+
+const windowScrollMotionValues = {
+    scrollX: motionValue(0),
+    scrollY: motionValue(0),
+    scrollXProgress: motionValue(0),
+    scrollYProgress: motionValue(0),
+}
+
+function useScrollMotionValues(isElement: boolean) {
+    /* ok for hook to be conditional since we don't support changing between element and window */
+    /* eslint-disable react-hooks/rules-of-hooks */
+    if (isElement) {
+        return {
+            scrollX: useMotionValue(0),
+            scrollY: useMotionValue(0),
+            scrollXProgress: useMotionValue(0),
+            scrollYProgress: useMotionValue(0),
+        }
+    } else {
+        return windowScrollMotionValues
+    }
+    /* eslint-enable react-hooks/rules-of-hooks */
 }
 
 function getWindowOffsets() {
@@ -28,7 +51,7 @@ function getElementOffsets(element: HTMLElement) {
     }
 }
 
-function createWindowListeners(updateCallback) {
+function createWindowListeners(updateCallback: () => void) {
     window.addEventListener("resize", updateCallback)
     window.addEventListener("scroll", updateCallback, { passive: true })
     return () => {
@@ -37,7 +60,10 @@ function createWindowListeners(updateCallback) {
     }
 }
 
-function createElementListeners(element: HTMLElement, updateCallback) {
+function createElementListeners(
+    element: HTMLElement,
+    updateCallback: () => void
+) {
     let resizeObserver
     if (window.ResizeObserver) {
         resizeObserver = new window.ResizeObserver(updateCallback)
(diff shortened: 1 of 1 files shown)


InventingWithMonster:
Hey man, thanks for taking a look at this! To answer your questions:

> Should we probably warn if ref.current is available but is not overflow: scroll?

I don't think this is necessary tbh. If the element doesn't scroll there probably won't be a broken expectation when scroll values don't update.

> Another solution could be to just grab the scroll closest scroll container, I have a small function from another library I can add to implement this

I don't think we should be too permissive in this instance.

> We should probably handle only adding one listener per element/window similar to what you had originally

Yeah so I definitely would like `window` to only have one listener and one set of `MotionValues`. If you want to keep the same logic for setting up both you could do something like saving the generated `MotionValue`s to a `WeakMap` that uses `window`/`ref.current` as the lookup.

> Can we utilize useResize hook and implement ResizeObserver inside that?

I'd prefer not to involve `ResizeObserver` beyond how you've set it up now. I think that strikes a good balance.

souporserious:
Thanks for taking a look! Sounds good! I'll set up the WeakMap today :)

src/value/use-viewport-scroll.ts
```diff
@@ -96,10 +132,46 @@ const viewportMotionValues: ScrollMotionValues = {
  *
  * @public
  */
-export function useViewportScroll() {
-    if (!hasEventListener) {
-        addScrollListener()
+export function useViewportScroll(ref: undefined | RefObject<Element>) {
```
paszkowskiDamian: In this case typescript will require passing either undefined or refObject. If you want optional value please use `?`

``` typescript 
export function useViewportScroll(ref?: RefObject<Element>)
```

otherwise this would throw an error:
``` typescript
const scroll = useViewportScroll() 
```

src/value/use-viewport-scroll.ts
```diff
@@ -7,55 +10,88 @@ export interface ScrollMotionValues {
...
+        window.removeEventListener("resize", updateCallback)
+        window.removeEventListener("scroll", updateCallback, { passive: true })
+    }
+}
 
-    window.addEventListener("resize", updateScrollValues)
-    window.addEventListener("scroll", updateScrollValues, { passive: true })
+function createElementListeners(element: HTMLElement, updateCallback) {
+    let resizeObserver
+    if (window.ResizeObserver) {
```
paszkowskiDamian: `ResizeObserver` has very poor support (around 60%). I don't know if it is library containers or consumers who should supply polyfill for that.

souporserious:
I looked into adding `WeakMap` support and I'm a little stuck 🤔. Since `ref` comes through as `null` on the first read there's no key to set for the map which blows other things up that are relying on motion values being created on the first pass. Any ideas here?

This is my simple implementation for the `WeakMap` portion so far:

```js
const motionValues = new WeakMap()

function getViewportMotionValues(element) {
  if (!motionValues.has(element)) {
    motionValues.set(element, {
      scrollX: motionValue(0),
      scrollY: motionValue(0),
      scrollXProgress: motionValue(0),
      scrollYProgress: motionValue(0),
    })
  }
  return motionValues.get(element)
}
```

InventingWithMonster:
Ah yeah I hadn't considered that. Maybe they are treated slightly differently then, with an event handler created for each element but window only getting a single set/motion values

InventingWithMonster:
Closed this by accident -_-

souporserious:
@InventingWithMonster been a little busy, apologies for the delay! Added support for reusing window scroll values. Not sure how to go about typing the callbacks for listener callbacks, went with `() => void` for now. Also, TypeScript is complaining about `ref.current` possibly being `undefined`, but the invariant should catch that so not sure what to do there either 😅.

souporserious:
👋 just checking in on this PR. Any interest in this still?

joebentaylor1995:
How can you achieve this without this PR?

lmartins:
Would it be possible to get an idea if this is something coming to Framer. PR seems stalled, just looking to understand if worth waiting for it.

breadadams:
I could also do with this feature on a project, can I be of any assistance in getting things moving again? 😇

InventingWithMonster:
Sorry for the delay on this!

thebuilder:
This change introduces `useLayoutEffect` to the the `useViewportScroll` hook, which causes a React warning when rendering serverside.

```
Warning: useLayoutEffect does nothing on the server, because its effect cannot be encoded into the server renderer's output format. This will lead to a mismatch between the initial, non-hydrated UI and the intended UI. To avoid this, useLayoutEffect should only be used in components that render exclusively on the client. See https://fb.me/react-uselayouteffect-ssr for common fixes.


A common fix is to a `useIsomorphicLayoutEffect` that selects the right effect for the env.

```js
const useIsomorphicLayoutEffect = isBrowser ? useLayoutEffect : useEffect

function Comp() {
    // ...
  })
}
```

PR #199, opened 2019-07-15
Exporting list of valid MotionProps

InventingWithMonster:
So we can select these in Library.

Also moving from a destructuring approach to object iteration + Set lookup for filtering out MotionProps, which is 10x faster and allocates less memory.

image

(The Map lookup is marginally faster still, but the definition is messier)

### DIFF
--- a/api/framer-motion.api.md
+++ b/api/framer-motion.api.md
@@ -687,10 +687,8 @@ export function useTransform<T>(parent: MotionValue<number>, from: number[], to:
 // @public
 export function useViewportScroll(): ScrollMotionValues;
 
-// Warning: (ae-internal-missing-underscore) The name "validMotionProps" should be prefixed with an underscore because the declaration is marked as @internal
-// 
 // @internal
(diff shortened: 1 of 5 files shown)


nvh:
How did you create that graph?

api/framer-motion.api.md
```diff
@@ -687,6 +687,11 @@ export function useTransform<T>(parent: MotionValue<number>, from: number[], to:
 // @public
 export function useViewportScroll(): ScrollMotionValues;
 
+// Warning: (ae-internal-missing-underscore) The name "validMotionProps" should be prefixed with an underscore because the declaration is marked as @internal
```
nvh: I think this is a valid warning

InventingWithMonster:
> How did you create that graph?

https://jsperf.com/prop-filtering/1

InventingWithMonster:
> Can we create a test that makes sure that fails if there is a MotionProp added that is not listed in the Set?

Preferably compilation would fail in this case but I can't get the `Set` to throw a type error if a prop is missing.

We can do a test that checks if some of the props aren't filtered out but that isn't the same thing (and these things would flag in the console anyway)

PR #200, opened 2019-07-15
Adding isValidMotionProp

InventingWithMonster:


src/dom/valid-motion-props.ts
```diff
@@ -0,0 +1,61 @@
...
+    "onHoverStart",
+])
+
+/**
+ * Check whether a prop name is a valid `MotionProp` key.
+ *
+ * @param key - Name of the property to check
+ * @returns `true` is key is a valid `MotionProp`.
+ *
+ * @public
```
nvh: Should this be part of the public api? I'm not sure. When would a consumer of the API use this?

src/motion/utils/valid-prop.ts
```diff
@@ -0,0 +1,61 @@
+import { MotionProps } from "../types"
```
nvh: One of these files needs to be removed, I think

src/motion/utils/valid-props.ts
```diff
@@ -0,0 +1,61 @@
+import { MotionProps } from "../types"
```
nvh: One of these files needs to be removed, I think

src/dom/valid-motion-props.ts
```diff
@@ -0,0 +1,61 @@
+import { MotionProps } from "../motion/types"
+
+/**
+ * A list of all valid MotionProps
+ *
+ * @internalremarks
+ * This doesn't throw if a `MotionProp` name is missing - it should.
+ */
+const validDomMotionProps = new Set<keyof MotionProps>([
```
nvh: What's the difference between this and validMotionProps?

src/dom/valid-motion-props.ts
InventingWithMonster: Same thing we're doing, or `@emotion/is-prop-valid`

Issue #196, opened 2019-07-13
[BUG] Newly-rendered sub-components stuck in initial variant


samdenty:
**Describe the bug**

**To Reproduce**
https://codesandbox.io/s/epic-hertz-u7neg

Add new items, and they're stuck in the initial variant.

**Expected behavior**

Newly-rendered components should use the `animate` variant.

**Actual behaviour**

Newly-rendered components are always in the `initial` variant

PR #201, opened 2019-07-16
Animating existing variants on newly-added variant children

InventingWithMonster:
Fixes https://github.com/framer/motion/issues/196

Part of the fix was only wrapping the rendered child in the created Context.Provider, had a bit of a cleanup around that.

### DIFF
--- a/src/motion/functionality/dom.tsx
+++ b/src/motion/functionality/dom.tsx
@@ -7,7 +7,6 @@ import { svgElements } from "../utils/supported-elements"
 import { Gestures } from "./gestures"
 import { MotionComponentConfig } from "../component"
 import { Drag } from "./drag"
-import { FunctionalProps } from "./types"
 import styler, { buildSVGAttrs } from "stylefire"
 import { parseDomVariant } from "../../dom/parse-dom-variant"
 import { MotionValuesMap } from "../../motion/utils/use-motion-values"
@@ -15,12 +14,6 @@ import { resolveCurrent } from "../../value/utils/resolve-values"
 import { Position } from "./position"
 import { isValidMotionProp } from "../utils/valid-prop"
 import { getAnimationComponent } from "./animation"
-
-type RenderProps = FunctionalProps & {
-    componentProps: MotionProps
-    style: CSSProperties
-    isStatic: boolean | undefined
-}
 
 function stripMotionProps(props: MotionProps) {
     const domProps = {}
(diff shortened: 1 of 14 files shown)


ThrowsException:
Any update on this? This is an incredibly frustrating bug when you come across it since it feels like one of those things that should "just work". Spent about 2 hours messing with my animations before seeing this Issue/PR open.

src/motion/functionality/dom.tsx
```diff
@@ -97,68 +81,52 @@ export function createDomMotionConfig<P = MotionProps>(
...
-            isStatic
+            inherit
         ) => {
             const activeComponents: JSX.Element[] = []
 
-            // TODO: Refactor the following loading strategy into something more dynamic
-            // This is also a good target for filesize reduction by making these present externally.
-            // It might be possible to code-split these out and useState to re-render children when the
-            // functionality within becomes available, or Suspense.
+            // TODO: Consolidate Animation functionality loading strategy with other functionality components
```
nvh: Is there a issue for this?

src/motion/functionality/types.ts
```diff
@@ -10,15 +10,23 @@ export interface FunctionalProps extends MotionProps {
...
+) => ReactElement<FunctionalProps>[]
+
+export type RenderComponent<P = {}> = (
     ref: RefObject<Element>,
     style: CSSProperties,
+    values: MotionValuesMap,
+    props: P,
     isStatic?: boolean
-) => ReactElement<P>[]
+) => ReactElement
```
nvh: why was the generic type removed here?

src/motion/functionality/types.ts
```diff
@@ -10,15 +10,23 @@ export interface FunctionalProps extends MotionProps {
 }
 
 export interface FunctionalComponentDefinition {
+    key: string
     shouldRender: (props: MotionProps) => boolean
-    component: ComponentType<FunctionalProps>
+    Component: ComponentType<FunctionalProps>
```
nvh: Is `Functional` the right terminology here? It is a bout the functionality, right? Maybe we can come up with better naming.

InventingWithMonster: Yeah I feel like it could do with a better name - at some point I want to open this API up so that would be a good time to come back to this.

src/motion/functionality/types.ts
InventingWithMonster: Basically it wasn't correct and started flagging.

PR #208, opened 2019-07-17
Updating library  docs for new motion features

InventingWithMonster:
Shaping-up the comments for `positionTransition` in regards to its usage in Library

### DIFF
--- a/src/motion/types.ts
+++ b/src/motion/types.ts
@@ -255,7 +255,7 @@ export interface AnimationProps {
     transition?: Transition
 
     /**
-     * @layout
+     * @library
      *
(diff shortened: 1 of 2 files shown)


src/motion/types.ts
```diff
@@ -255,29 +255,44 @@ export interface AnimationProps {
     transition?: Transition
 
     /**
-     * If `positionTransition` is defined, the component will automatically animate any changes to its layout
-     * relative to its nearest positioned parent.
+     * @layout
```
nvh: What does `@layout` do? Shouldn't this be `@library`?

InventingWithMonster: Yes - doh

Issue #209, opened 2019-07-17
[BUG] Animate height “0” to “auto” when switching to “display: block” from “none”


conor-cafferkey-sociomantic:
Animating height from `0` to `auto` not is working properly when switching to `display: block` from `display: none`:

![accordion-bug](https://user-images.githubusercontent.com/13482427/61403744-43794700-a8d6-11e9-877e-81420b0212cb.gif)

According to the [Migrate from Pose](https://www.framer.com/api/motion/migrate-from-pose/#applyatstart%2Fapplyatend) documentation, _non-animatable properties are automatically applied at the start of an animation. Anything that you want to be applied at the end can be set to `transitionEnd`_.

The bug occurs with the following combination of properties:
```
{
  open: { display: "block", opacity: 1, height: "auto" },
  collapsed: { opacity: 0, height: 0, transitionEnd: { display: "none" } }
}
```

**To Reproduce**
1. Go to https://codesandbox.io/s/framer-motion-accordion-o0to4
2. Click on the accordion headers
4. Animation doesn’t work as expected on open (works fine on close)

**Expected behavior**
Smooth animation on opening and closing the accordion:

![accordion-expected](https://user-images.githubusercontent.com/13482427/61403761-48d69180-a8d6-11e9-89f3-728e72d28f80.gif)

This works fine **in Pose** using the following combination of properties:
```
{
  open: {  opacity: 1, height: "auto" , applyAtStart: { display: "block" } },
  collapsed: { opacity: 0, height: 0, applyAtEnd: { display: "none" }
}
```

InventingWithMonster:
Yeah good catch. I'll dig into this.

conor-cafferkey-sociomantic:
Thanks!

PR #210, opened 2019-07-18
Fixing value type conversion for hidden elements

InventingWithMonster:
Fixes https://github.com/framer/motion/issues/209

To do value type conversion, we have to measure previous and target bounding boxes. If they're hidden layers, this bounding box returns `0` for every measurement, making delta calculation impossible.

### DIFF
--- a/src/dom/unit-type-conversion.ts
+++ b/src/dom/unit-type-conversion.ts
@@ -93,8 +93,21 @@ const convertChangedValueTypes = (
     const elementStyler = styler(element)
     const originBbox = element.getBoundingClientRect()
     const elementComputedStyle = getComputedStyle(element)
-    const { top, left, bottom, right, transform } = elementComputedStyle
+    const {
+        display,
+        top,
+        left,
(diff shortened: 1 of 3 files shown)


src/dom/unit-type-conversion.ts
```diff
@@ -93,9 +93,22 @@ const convertChangedValueTypes = (
...
+        bottom,
+        right,
+        transform,
+    } = elementComputedStyle
     const originComputedStyle = { top, left, bottom, right, transform }
 
+    // If the element is currently set to display: "none", make it visible before
+    // measuring the target bounding box
+    if (display === "none") {
+        elementStyler.set("display", target.display || "block")
```
nvh: I assume this behaviour is not really testable?

InventingWithMonster: Yeah sadly not. This doesn't result in a flash because we hit multiple renders before the next frame is painted, so it's left in the correct state by then.

Issue #207, opened 2019-07-17
[BUG] DOM unit type conversion doesn't account for scale transforms


InventingWithMonster:
Add `scale: 0.5` to `collapsed` and  `scale: 1` to `open` in https://codesandbox.io/s/framer-motion-viewport-scroll-and-svg-path-animation-9r68o - you can see the open/collapse animations jump at the end.

This is because we're using `getBoundingClientRect` to measure the element before and after we apply `height: auto`. We need to do this because we're also measuring unit type conversion for `x` and `y` too - otherwise we could probably just use `offsetHeight` etc.

It might be that we can first set `scale` and `rotate` to their defaults before measuring the before/after bounding boxes. And then restore them to their previous values.

Issue #204, opened 2019-07-16
[BUG] Simple render testing give element.getTotalLength is not a function


rhogeranacleto:
**Describe the bug**

I created this component and put it on my app page:

```tsx
function Punch({ punchIt, initialIsExit }: { initialIsExit: boolean, punchIt(): Promise<void> }) {

  const [exit, setExit] = useState(initialIsExit);
  const [canDrag, setCanDrag] = useState<'x' | false>('x');
  const controls = useAnimation();
  const [ok, setOk] = useState(false);
  const x = useMotionValue(0);
  const background = useTransform(x, v => backgroundColorInterpolation(Math.abs(v)));
  const scale = useTransform(x, v => backgroundScaleInterpolation(Math.abs(v)));
  const color = useTransform(x, v => indicatorColorInterpolation(Math.abs(v)));
  const tickPath = useTransform(x, v => indicatorProgressInterpolation(Math.abs(v)));
  const [svgDisplay, setSvgDisplay] = useState('none');
  const okpath = useMotionValue(0);

  async function onDragEnd(event: MouseEvent, info: PanInfo) {

    if (Math.abs(info.point.x) > moveSize) {

      setCanDrag(false);

      await controls.start({
        x: exit ? -moveSize : moveSize,
        transition: {
          ease: 'easeOut'
        }
      });

      setExit(!exit);

      await punchIt();

      setOk(true);

      await controls.start({
        x: 0,
        left: exit ? 8 : moveSize + 8,
        transition: {
          delay: 0.5,
          duration: 0.6,

        }
      });

      setOk(false);
      setCanDrag('x');

      return;
    }

    await controls.start({
      x: 0,
      transition: {
        duration: 0.5
      }
    });

    setSvgDisplay('none');
  }

  function okAnimationComplete() {

    if (okpath.get() === 0) {

      setSvgDisplay('none');
    }
  }

  return (
    <div className="punch-container">
      <motion.span>
        {exit ? 'Sair' : 'Entrar'}
      </motion.span>
      <motion.div
        drag={canDrag}
        animate={controls}
        dragMomentum={false}
        dragElastic={1}
        className="punch-button"
        onDragStart={() => setSvgDisplay('block')}
        onDragEnd={onDragEnd}
        style={{ x, left: initialIsExit ? moveSize + 8 : 8 }}>

        <motion.div className="back" style={{ scale, background }} />

        <motion.div animate={{ rotate: exit ? 180 : 0 }} className="icon-container">
[...]

PR #211, opened 2019-07-18
Fixing unit conversion when non-positional transforms are applied

InventingWithMonster:
Fixes https://github.com/framer/motion/issues/207

Stylefire updated also fixes https://github.com/framer/motion/issues/204

Stuff like rotate/scale/skew can ruin bounding box calculations, so temporarily unapplying these when measuring.


InventingWithMonster:
I'm going to put the fix in for now and if we can remove a render by interpreting a matrix then that's cool.

Issue #181, opened 2019-07-10
[BUG] whileHover doesn't animate back to int value from percent value


hennessyevan:
**Describe the bug**
Width gets stuck at 100% when hovering in a basic example

**To Reproduce**
1. Visit codesandbox link
2. hover over the black div for more than a few seconds
3. When you leave hover it stays at 100%

**Screencap**
![Screen Recording 2019-07-10 at 3 13 16 PM](https://user-images.githubusercontent.com/16711653/60997785-a5114280-a325-11e9-8503-69a7a63d5c01.gif)

**Expected behavior**
It should animate back to the initial width on mouse out.

**Codesandbox**
https://codesandbox.io/s/variants-k9vfb

**Desktop (please complete the following information):**
 - OS: macOS mojave
 - Browser chrome
 - Version 75

Andarist:
It's because the width is set to "100%" [here](https://github.com/framer/motion/blob/6eef683eed810f2bc9800c4c746130829ae75f87/src/animation/ValueAnimationControls.ts#L216) when animation ends (origin of this set is [here](https://github.com/framer/motion/blob/6eef683eed810f2bc9800c4c746130829ae75f87/src/animation/ValueAnimationControls.ts#L514)). And later values are being "restored" onHoverEnd from that set value [here](https://github.com/framer/motion/blob/6eef683eed810f2bc9800c4c746130829ae75f87/src/animation/ValueAnimationControls.ts#L377).

hennessyevan:
Thanks for that explanation @Andarist, that makes sense. Just to be clear this **is** a bug right? This is the only thing keeping me moving from react-pose at the moment, interested on what the path forward would be.

Andarist:
Yes, this is a bug. I’ve just debugged this in hope to preparing a fix but it ended up to be more difficult than i was prepared for. It was my first dive into this codebase and i dont know yet how all connects together. Nevertheless i believe my findings were good so ive wrote it down here so somebody could pick it up from here.

PR #213, opened 2019-07-18
Fixing whileX unsetting if they contain transitionEnd

InventingWithMonster:
Fixes https://github.com/framer/motion/issues/181

The transitionEnd value was being set as the base value in every instance, but we only want it set when it **isn't** part of a whileX gesture prop.

### DIFF
--- a/src/animation/ValueAnimationControls.ts
+++ b/src/animation/ValueAnimationControls.ts
@@ -20,10 +20,16 @@ export type AnimationDefinition =
     | VariantLabels
     | TargetAndTransition
     | TargetResolver
+
 type AnimationOptions = {
     delay?: number
     priority?: number
     transitionOverride?: Transition
+}
+
+type SetOptions = {
+    isActive?: Set<string>
+    priority?: number
 }
 
 /**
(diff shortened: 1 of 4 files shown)


src/animation/ValueAnimationControls.ts
```diff
@@ -20,12 +20,18 @@ export type AnimationDefinition =
...
     | TargetAndTransition
     | TargetResolver
+
 type AnimationOptions = {
     delay?: number
     priority?: number
     transitionOverride?: Transition
 }
 
+type SetOptions = {
```
nvh: Maybe rename this to `SetterOptions` to not let it bet confused by a JS `Set`

src/animation/ValueAnimationControls.ts
```diff
@@ -213,7 +219,7 @@ export class ValueAnimationControls<P extends {} = {}, V extends {} = {}> {
                 this.values.set(key, motionValue(targetValue))
             }
 
-            this.baseTarget[key] = targetValue
+            if (!priority) this.baseTarget[key] = targetValue
```
nvh: What about priority 0?

src/gestures/__tests__/hover.test.tsx
```diff
@@ -112,7 +112,7 @@ describe("hover", () => {
     test("whileHover is unapplied when hover ends", () => {
         const promise = new Promise(resolve => {
             const variant = {
-                hidden: { opacity: 0.5 },
+                hidden: { opacity: 0.5, transitionEnd: { opacity: 0.75 } },
```
Andarist: wouldn't this be better split into 2 separate tests?

PR #217, opened 2019-07-19
Fix MotionProp.style @library example doc

aron:
It was using the individual `x`, `opacity` and `scale` props instead of `style`.


jordandobson:
https://github.com/framer/company/issues/13813

Still seems like this issue exists.

aron:
@jordandobson it'll be an issue until we publish a new version of the library/framer-motion which happens roughly once a week.

Issue #216, opened 2019-07-19
[BUG] SVG animation fails when no initial value set


mrmckeb:
**Describe the bug**
When I don't define default/initial values _as attributes_, the component fails with the error:
```
Uncaught TypeError: Cannot read property 'red' of null
```

**To Reproduce**
As an example, this component/animation fails. This should be a looping spinner.
```tsx
const animation = {
  strokeDasharray: ['1px, 200px', '100px, 200px', '100px, 200px'],
  strokeDashoffset: [0, -15, -125],
  transition: { duration: 1.4, ease: 'linear' },
};

const Spinner: FC = (): ReactElement => {
  const controls = useAnimation();

  const handleAnimationComplete = useCallback(() => {
    controls.start(animation);
  }, [controls]);

  useEffect(() => {
    controls.start(animation);
  }, [controls]);

  return (
    <svg viewBox='22 22 44 44'>
      <Circle
        animate={controls}
        onAnimationComplete={handleAnimationComplete}
        cx='44'
        cy='44'
        r='20.2'
        fill='none'
        strokeWidth='3.6'
      />
    </svg>
  );
};
```

**Expected behavior**
Modifing the circle to the below works, but this is the only way to resolve the issue. Setting the styles in CSS (with styled-components as an example) does not help.
```tsx
      <Circle
        animate={controls}
        onAnimationComplete={handleAnimationComplete}
        strokeDasharray='80px, 200px'
        strokeDashoffset='0'
        cx='44'
        cy='44'
        r='20.2'
        fill='none'
        strokeWidth='3.6'
      />
```

**Desktop (please complete the following information):**
 - OS: MacOS 10.14.5 (18F132)
 - Browser Version: Chrome 75.0.3770.100 (Official Build) (64-bit)

**Additional context**
I had the same issue when trying to add a `rotate` animation to an SVG element. I could not find any workaround for this one, so any help would be greatly appreciated.
```tsx
const Spinner: FC= (): ReactElement => {
  return (
    <motion.svg viewBox='22 22 44 44' animate={{ rotate: 100 }}>
      <Circle />
    </motion.svg>
  );
};
```

Other than this, I'm loving Framer Motion - amazing work!

InventingWithMonster:
Thanks for the detailed write-up! I'll take a look at this next week.

Glad you're enjoying it!

mrmckeb:
Thanks @InventingWithMonster! Let me know if I can help at all.

artokun:
[...]

PR #221, opened 2019-07-22
Improving unencountered value initialisation

InventingWithMonster:
Fixes https://github.com/framer/motion/issues/216

When trying to animate an otherwise-undefined SVG attribute, Framer Motion would read it via Stylefire. Unlike CSS and computed style, SVG returns `null` for undefined attributes. To help with that the first thing we now do is check if the target value is a keyframes array, if so take the first value of that instead of reading from Stylefire. It also didn't correctly return SVG `transform` default values which has been fixed in an update.

### DIFF
--- a/src/motion/__tests__/component-svg.test.tsx
+++ b/src/motion/__tests__/component-svg.test.tsx
@@ -62,4 +62,22 @@ describe("SVG", () => {
         }
         render(<Component />)
     })
+
+    // https://github.com/framer/motion/issues/216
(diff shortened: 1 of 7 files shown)


src/motion/__tests__/component-svg.test.tsx
```diff
@@ -62,4 +62,22 @@ describe("SVG", () => {
         }
         render(<Component />)
     })
+
+    // https://github.com/framer/motion/issues/216
```
nvh: I like referencing issues in tests!

PR #224, opened 2019-07-23
Fix issue with resolvedConstraints being reset on update

### LINKED ISSUE #223: [BUG] dragConstraints ref is ignored when onDragStart updates a state
**Describe the bug**

When the `onDragStart` callback updates a state (and schedules a render) the dragConstraints are initially ignored.

https://codesandbox.io/s/determined-meitner-zufuz

**To Reproduce**

1. open the sandbox example
2. move the small box outside the big box

**Expected behavior**
The dragConstraints should be always respected 

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser Chrome
 - Version latest

Andarist:
fixes #223

This is a quick, probably a proper one would be to output `[0,0,0,0]` for ref constraint as well. Gonna explore doing this and adding tests later. Do you see anything that could get broken by my proposed "proper" solution?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 Framer Motion adheres to [Semantic Versioning](http://semver.org/).
 
+## Unreleased
+
+### Fixed
+
+-   Fixing issue with drag constraints (ref-based) being reset, while dragging, on unrelated parent component updates.
+
 ## [1.3.0] 2019-07-24
 
 -   Added `onAnimationStart`.
--- a/src/behaviours/__tests__/index.test.tsx
+++ b/src/behaviours/__tests__/index.test.tsx
@@ -788,7 +788,7 @@ describe("dragging", () => {
         })
 
         return expect(promise).resolves.toHaveStyle(
-            "transform: translateX(105px) translateZ(0)"
+            "transform: translateX(105px) translateY(0px) translateZ(0)"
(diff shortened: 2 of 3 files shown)


InventingWithMonster:
Thanks for taking a look at this!

I'm not sure what the "proper" solution would achieve over this one though? Could you add a `dev/example` (probably based the linked Codesandbox) that could show this working in action? It's not an idea test suite but it's better than nothing for layout-heavy stuff

Andarist:
> I'm not sure what the "proper" solution would achieve over this one though?

1. It wouldn't reapply effect in this case at all when owner component rerenders
2. it would allow for switching between constraint types - object & ref, currently, it's not always possible because React yells at us about changing effect inputs' length

I'll try to add tests for both (and apply a proposed "proper" fix) later - looking at existing `useDraggable` test suite it shouldn't be too hard to do that.

Andarist:
@InventingWithMonster I've adjusted the PR to what I believe is a more correct fix, please take a look. The fact that the issue has been fixed can be verified on this sandbox - https://codesandbox.io/s/angry-lederberg-5sokr 

In the meantime, I've also started to also work on setting up Cypress so we should be able to write tests for issues like this in the near future (having a working draft already).

InventingWithMonster:
Excellent fix, thanks

PR #229, opened 2019-07-24
Adding MotionValue.onChange example

InventingWithMonster:


src/value/index.ts
```diff
@@ -204,14 +204,23 @@ export class MotionValue<V = any> {
...
      *   const x = useMotionValue(0)
+     *   const y = useMotionValue(0)
+     *   const opacity = useMotionValue(1)
      *
      *   useEffect(() => {
-     *     const unsubscribe = x.onChange((latestX) => {
-     *       // Do stuff with latest x value
-     *     })
+     *     function updateOpacity() {
+     *       const newOpacity = transform(Math.max(x.get(), y.get()), [0, 100], [1, 0])
```
nvh: I think it would be better for the example to pull the `Math.max` out into a variable

src/value/index.ts
```diff
@@ -204,14 +204,23 @@ export class MotionValue<V = any> {
...
+     *     }
      *
-     *     return unsubscribe
-     *   })
+     *     const unsubscribeX = x.onChange(updateOpacity)
+     *     const unsubscribeY = y.onChange(updateOpacity)
+     *
+     *     return () => {
+     *       unsubscribeX()
+     *       unsubscribeY()
```
nvh: Do we have a explanation somewhere of why we need to unsubscribe?

InventingWithMonster: Yeah it says so we don't have multiple subscriptions

PR #230, opened 2019-07-24
AnimatePresence single prop

InventingWithMonster:
Enables `AnimatePresence` to block an incoming child until the old one has animated out.

For instance with paginated form flows/tutorial flows etc.

### DIFF
--- a/api/framer-motion.api.md
+++ b/api/framer-motion.api.md
@@ -21,16 +21,18 @@ import { RefObject } from 'react';
 import { SpringProps } from 'popmotion';
 import { SVGAttributes } from 'react';
 
+// Warning: (ae-forgotten-export) The symbol "AnimatePresenceProps" needs to be exported by the entry point index.d.ts
+// 
 // @public
-export const AnimatePresence: FunctionComponent<AnimatePresenceProps>;
+export const AnimatePresence: FunctionComponent<AnimatePresenceProps_2>;
 
 // @public (undocumented)
 export interface AnimatePresenceProps {
     custom?: any;
+    // @beta
+    enterBeforeExit?: boolean;
     initial?: boolean;
     onExitComplete?: () => void;
-    // @beta
-    single?: boolean;
     // @internal
     _syncLayout?: () => void;
 }
--- a/api/framer-motion.api.md
+++ b/api/framer-motion.api.md
@@ -21,8 +21,10 @@ import { RefObject } from 'react';
 import { SpringProps } from 'popmotion';
 import { SVGAttributes } from 'react';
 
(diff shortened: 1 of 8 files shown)


api/framer-motion.api.md
```diff
@@ -21,15 +21,19 @@ import { RefObject } from 'react';
 import { SpringProps } from 'popmotion';
 import { SVGAttributes } from 'react';
 
+// Warning: (ae-forgotten-export) The symbol "AnimatePresenceProps" needs to be exported by the entry point index.d.ts
+// 
 // @public
-export const AnimatePresence: FunctionComponent<AnimatePresenceProps>;
+export const AnimatePresence: FunctionComponent<AnimatePresenceProps_2>;
```
nvh: This is weird, why are there two things that are named that?

api/framer-motion.api.md
```diff
@@ -21,15 +21,19 @@ import { RefObject } from 'react';
 import { SpringProps } from 'popmotion';
 import { SVGAttributes } from 'react';
 
+// Warning: (ae-forgotten-export) The symbol "AnimatePresenceProps" needs to be exported by the entry point index.d.ts
```
nvh: This warning should also be avoided

InventingWithMonster:
> I'm not sure `single` is a great name for this API, because I think it's not really descriptive. Although there is only one child animating at the same time, it's more about the order the items, right? Don't have a great alternative, but what about sequential?

It is mostly about only allowing a single child to be rendered at any one time. Sequential invokes the idea of moving through an array but this isn't necessarily the case (ie the tabbed navigation). But I'm also struggling for a better name.

- `singleOnly`
- `singleChildOnly`
- `exitBeforeEnter`
- `exclusiveChild`

api/framer-motion.api.md
InventingWithMonster: These are both quite weird. `AnimatePresenceProps_2` isn't even defined and I do export `AnimatePresenceProps` from the root file.

InventingWithMonster: Deleted `types` and re-ran, is all good now

Andarist:
For what it's worth - from proposed named I like `exitBeforeEnter` the best.

InventingWithMonster:
> For what it's worth - from proposed named I like `exitBeforeEnter` the best.

It's not my favourite aesthetically but it's clear, uses simple words and leaves open the possibility of extending the behaviour to handle multiple children (ie 2 in 2 out)

nvh:
I like that `exitBeforeEnter` contains `exit`, which is the prop that is used on the children for animating out. It also has my vote.

PR #236, opened 2019-07-24
Fix what is used as external in rollup config

Andarist:
The hardcoded list was missing `tslib` dependency. By using package.json data it will always stay up-to-date.

This has reduced esm & cjs bundles a little bit, esm went down from 40111 to 39423 (unminified, but gzipped)


InventingWithMonster:
Excellent, thanks!

Issue #214, opened 2019-07-18
[FEATURE] Improve AnimatePresence API with custom components


InventingWithMonster:
Currently, because of a usually-hidden implementation detail with `AnimatePresence`, using custom components works a little like this:

```jsx
const Custom = ({ animate, onAnimationComplete }) => (
  <motion.div
    animate={animate}
    onAnimationComplete={onAnimationComplete})
  />
)

const MyComponent = ({ isVisible }) => (
  <AnimatePresence>
    {isVisible && (
      <Custom exit={{ opacity: 0 }} />
    )}
  </AnimatePresence>
)
```

Which is a bit of a mess and doesn't work well with types. 

It might be possible with context to achieve an intuitive API like this:

```jsx
const Custom = () => (
  <motion.div exit={{ opacity: 0 }} />
)

const MyComponent = ({ isVisible }) => (
  <AnimatePresence>
    {isVisible &&  <Custom  />}
  </AnimatePresence>
)
```

PR #238, opened 2019-07-25
Better support for custom components in `AnimatePresence`

InventingWithMonster:
Closes https://github.com/framer/motion/issues/214

Now allows custom components as direct children of `AnimatePresence` without passing arbitrary props around.

```jsx
const Image = () => (
  <motion.img animate={{ opacity: 1 }} exit={{ opacity: 0 }} />
)

const MyComponent = ({ isVisible }) => (
  <AnimatePresence>
    {isVisible && <Image />}
  </AnimatePresence>
)
```

It replaces the previous approach of cloning every immediate child and essentially copy/pasting the `exit` prop into `animate` with wrapping every child in a `MotionContext` provider that passes down `exitProps`. We're then responsible for consuming those in the `motion` component itself which I think is a much more flexible approach.


src/motion/utils/apply-exit-props.ts
```diff
@@ -0,0 +1,26 @@
...
+    { initial, isExiting, custom, onExitComplete }: ExitProps
+): P & MotionProps {
+    if (isExiting) {
+        invariant(!!props.exit, "No exit animation defined.")
+
+        return {
+            ...props,
+            custom: custom !== undefined ? custom : props.custom,
+            animate: props.exit,
+            onAnimationComplete: () => {
```
jonastreub: This could potentially override user props right?

InventingWithMonster: Yeah it does so by design - I'll add a comment

oallouch:
Hi,
I'm a new fan of this lib.
Small question : Is the comment [https://github.com/framer/motion/blob/4ccfc45be896b6b937c0b89eebd12ee8c86b2948/src/components/AnimatePresence/index.tsx#L113](url) still relevant ?

InventingWithMonster:
No: I've updated the source to read

 ```
 * You can use any component(s) within `AnimatePresence`, but the first `motion` component in each should
 * have an `exit` property defined.
```

PR #243, opened 2019-07-25
SharedLayout component

InventingWithMonster:
When layout changes happen asynchronously to their instigating render (ie when exiting children of `AnimatePresence` are removed), `SharedLayout` can wrap parent and sibling components that need to animate as a result of this layout change.

```jsx
const MyComponent = ({ isVisible }) => {
  return (
    <SharedLayout>
      <AnimatePresence>
        {isVisible && (
          <motion.div exit={{ opacity: 0 }} />
        )}
      </AnimatePresence>
      <motion.div positionTransition />
    </SharedLayout>
  )
}
```

Naming up for discussion. `SyncLayout`? Bearing in mind a potential upcoming `layoutTransition` prop. It might also play a role in a hypothetical magic move feature, but probably premature to imagine this.


src/motion/functionality/position.ts
```diff
@@ -103,6 +104,7 @@ export const Position: FunctionalComponentDefinition = {
             values,
             positionTransition,
         }: FunctionalProps) => {
+            useContext(SharedLayoutContext)
```
nvh: Why is the result from this context not used?

src/components/AnimatePresence.tsx
```diff
@@ -211,8 +175,17 @@ export const AnimatePresence: FunctionComponent<AnimatePresenceProps> = ({
...
     onExitComplete,
-    _syncLayout,
 }) => {
+    // Use a running state counter to allow us to force a re-render once all
+    // exiting animations have settled. For performance reasons it might also be
+    // necessary to track entering animations too, but this should be good enough
+    // for the majority use-cases ie slideshows, modals etc.
+    const [forcedRenderCount, setForcedRenderCount] = useState(0)
+    const incrementForcedRenderCount = () =>
+        setForcedRenderCount(forcedRenderCount + 1)
```
nvh: Shouldn't this be wrapped in a useCalback?
This also seems like it is duplicated from `SharedLayout` so it can be extracted in a `useForceUpdate` hook

src/components/AnimatePresence.tsx
```diff
@@ -211,8 +175,17 @@ export const AnimatePresence: FunctionComponent<AnimatePresenceProps> = ({
...
+    const syncLayout =
```
nvh: Why not call this `forceUpdate`, like class components have?

src/components/SharedLayout.tsx
```diff
@@ -0,0 +1,56 @@
...
+ * @internalremarks
+ *
+ * The way this component works is by memoising a function and passing it down via context.
+ * The function, when called, updates the local state, which is used to invalidate the
+ * memoisation cache. A new function is called, performing a synced re-render of components
+ * that are using the SharedLayoutContext.
+ *
+ * @beta
+ */
+export const SharedLayout = ({ children }: SharedLayoutProps) => {
```
nvh: I'm not sure about the naming for this. I like to incorporate in the naming that it is about the Animations that happen inside this component.

src/motion/functionality/position.ts
InventingWithMonster: When you use a context you can also be saying "update me when this context updates" - the SharedLayoutContext is just updating so this renders as a result

src/components/SharedLayout.tsx
InventingWithMonster: `SyncLayoutAnimations` ?

InventingWithMonster: Would work if with magic move too

InventingWithMonster: @benjamindenboer shared a preference for a shorter name ie `SyncLayout`

InventingWithMonster:
I'm going to release as `UnstableSyncLayout` as `@internal` for now and we can land on a name in tandem with the `layoutTransition` prop API

PR #245, opened 2019-07-25
useInvertedScale

InventingWithMonster:
A hook for returning scale values that update with the inverse of their parent scale.

Used to remove visual distortion induced from scaling parent components.

```jsx
const MyComponent = () => {
  const { scaleX, scaleY } = useInvertedScale()
  return <motion.div style={{ scaleX, scaleY }} />
}
```

### DIFF
--- a/src/value/use-inverted-scale.ts
+++ b/src/value/use-inverted-scale.ts
@@ -1,6 +1,13 @@
 import { useContext } from "react"
 import { MotionContext } from "../motion/context/MotionContext"
 import { useTransform } from "../value/use-transform"
+import { MotionValue } from "./"
+import { invariant } from "hey-listen"
+
+interface ScaleMotionValues {
+    scaleX: MotionValue<number>
+    scaleY: MotionValue<number>
+}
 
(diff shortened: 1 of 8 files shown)


src/value/use-inverted-scale.ts
```diff
@@ -0,0 +1,33 @@
...
+ *   return <motion.div style={{ scaleX, scaleY }} />
+ * }
+ * ```
+ *
+ * @beta
+ */
+export const useInvertedScale = () => {
+    const { values } = useContext(MotionContext)
+    const scaleX = useTransform(values!.get("scaleX", 1), invertScale)
+    const scaleY = useTransform(values!.get("scaleY", 1), invertScale)
```
nvh: Why do we need to force unwrap here? Can't we early `return {scaleX: 1, scaleY: 1}` when values does not exist, maybe logging a warning?

InventingWithMonster: Because they might exist in the future, and it doesn't harm anything to initialise them as 1 in the meantime

PR #247, opened 2019-07-25
Add e2e tests using Cypress

Andarist:
In some regards this is not ideal, i.e. I couldn't make watcher on src/dev behave correctly. Fixing this can be explored in the future.

Also waits in the test are not really that good, but for the test to be correct we have to wait for React flushing changes to the DOM before proceeding to next steps and I was not sure how to achieve this effect correctly without rerendering component and waiting for some rerendered content. Also had a problem waiting for a selector (`.get(selector)`) in the middle of the dragging sequence - it kinda looked like cypress has abandoned pointermove on .get call 🤷‍♂


Andarist:
@InventingWithMonster friendly 🏓

InventingWithMonster:
Thanks for this we really need it!

There's an error on CircleCI, do you know what it could be?

```
 start-server-and-test start-dev-server http://localhost:9990 'cypress run' --ci --reporters=jest-junit
/home/circleci/repo/node_modules/lazy-ass/index.js:117
      throw err;
      ^

Error: expected <NPM script name that starts server> <url or port> <NPM script name that runs tests>
example: start-test start 8080 test
```

Andarist:
I suspect that it’s about your CI script, extra CLI arguments are being passed to each subcommand - including part of the cypress script which seems not to like it. The script itself runs OK on my local machine (havent run full CI script though). Would prefer you fixing this as you might have some preferences on how those scripts being ran - if you don’t have time right now and you can accept “whatever works” then i can look into this some time later

InventingWithMonster:
Cool ok - I'll take a look at merging this tomorrow and see if I can get them integrated. Thanks for implementing this, could be the start of something big with our testing.

Issue #227, opened 2019-07-24
Nested dragging not working as expected


tjahn:
**Describe the bug**

Hello, I tried copied example /dev/examples/dragPropagation.tsx which renders a red box nested within a blue box, both draggable.

What I experience is that after dragging the blue box once, I can not drag the inner red box without effecting the blue box.
When dragging the red box before touching the blue box everything works as I expected.
Somehow stopPropagation stops working after the first drag on the outer box.

**To Reproduce**
* Copy the example https://github.com/framer/motion/blob/master/dev/examples/dragPropagation.tsx
* bundle it into some react application
* drag red first -> it is working
* drag blue, then drag red -> blue moves together with red

**Expected behavior**
I expected the blue box to be uneffected by the drag of the inner, red box. Regardless of the order in which I touch them.

**Desktop**
 - MacOS
 - Google Chrome 75 and Safari 12.1.1.
- "framer-motion": "^1.2.6"
- "react": "^16.8.6" on

InventingWithMonster:
Ah this sounds related to an emergent bug in Framer X that we're going to look at on Friday. Thanks for the report - the minimal reproduction I needed was sat right here all along!

PR #249, opened 2019-07-26
Fixing drag propagation

InventingWithMonster:
Fixes https://github.com/framer/motion/issues/227

Accidentally commited test for this to master: https://github.com/framer/motion/blob/master/src/behaviours/__tests__/index.test.tsx#L406

I've moved the reference to the `openGlobalLock` to a ref as this was being lost in hot-reloading environments like CodeSandbox. The actual fix here is deleting the reference to the unlock function as this was being called by every component that had gained reference to it. So once you'd started dragging a draggable parent, it gains access to unlocking the global lock, and now when the child is dragged they both can gain the lock by simply unlocking it first.

### DIFF
--- a/src/behaviours/use-draggable.ts
+++ b/src/behaviours/use-draggable.ts
@@ -602,6 +602,10 @@ export function useDraggable(
         y: dragOriginY || defaultOriginY,
     }
 
+    // This is a reference to the global drag gesture lock, ensuring only one component
+    // can "capture" the drag of one or both axes.
+    const openGlobalLock = useRef<Lock | null>(null)
+
     const { transformPagePoint } = useContext(MotionPluginContext)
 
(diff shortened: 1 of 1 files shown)


src/behaviours/use-draggable.ts
```diff
@@ -939,12 +938,12 @@ export function useDraggable(
...
                 info: PanInfo
             ) => {
-                cancelDrag()
-
-                if (!openGlobalLock || !hasDragged) {
+                if (!openGlobalLock.current || !hasDragged) {
                     return
                 }
 
+                cancelDrag()
```
nvh: Should the cancelDrag move to after the return?

InventingWithMonster:
I think the gist is the drag shouldn’t be active at this point anyway - but
yeah I will double check

PR #250, opened 2019-07-26
Upgrade testing-library libraries

Andarist:


Andarist:
should be ready to merge now - previously was failing because I've forked from a failing master

InventingWithMonster:
Excellent thanks for this!

PR #253, opened 2019-07-30
Fixing pan and drag gestures with PointerEvents

InventingWithMonster:
`PointerEvent` inherits from `MouseEvent` so we need to check `event.pointerType` to determine whether it actually is a mouse event or not.


InventingWithMonster:
I don't think so as the reason I had to add `typeof PointerEvents !== 'undefined'` was because it wasn't recognised in Node

Issue #262, opened 2019-08-02
[FEATURE] Improved support for managing both variant-style animations as well as arbitrary-value style animations at the same time


jamesplease:
**Is your feature request related to a problem? Please describe.**

This isn't really a problem, per se, but it is a way to make the declarative component props API work in more situations.

Some background:

framer-motion's component props API supports two distinct kinds of animations. The first kind is "variant"-based animations, where the different variants capture discrete versions of a particular element. For instance, a component might have the following variants:

- open, closed, big, small

Consumers can animate between these variations using the `animate` prop:

```jsx
// Starting with these variants...
<motion.div variants={variants} animate={['open', 'big']} />

// ...you can then animate over to these
<motion.div variants={variants} animate={['closed', 'small']}/>
```

It also supports arbitrary animations using the object form of the `animate` prop. For instance, the position of an element could be animated based off of a number that's randomly generated every few seconds:

```jsx
const [x, setX] = useState(50);

useEffect(() => {
  setInterval(() => {
    setX(Math.random() * 100);
  }, 1000);
}, []);

// As `x` changes arbitrarily, this div will move around the page.
return (<motion.div animate={{ x }) />);
```

Tthe component API of framer-motion does not support doing both kinds of these animations at the same time, but a slight change to that API could make it so that it does.

**Describe the solution you'd like**

Updating the `custom` prop triggers animations when it changes would allow components to handle both kinds of animations. That way, even when the variants specified in `animate` remain constant, they can dynamically update from a value passed in through `custom`.

Right now, changes to `custom` do not trigger rerenders. [**View a CodeSandbox demonstrating this here.**](https://codesandbox.io/s/framer-motion-custom-doesnt-animate-8st85)

However, consider the following:

```jsx
// Given the following variant, which depends on `custom`:
const someVariants = {
  position(xValue) {
    return {
      x: xValue * 50
    };
  }  
};

// as `xValue ` changes, the `position` variant would animated, even if the
// `animate` prop doesn't change.
<motion.div variants={someVariants} animate="position" custom={xValue} />
```

[**View a CodeSandbox demonstrating this change using the code from #264**](https://codesandbox.io/s/framer-motion-custom-now-animates-4cqim)

This particular example is simple enough that you can solve it using existing framer-motion APIs, but the motivation for this change comes from more complex examples; namely, when you also want to use variants. See the CodeSandbox link below for an example of that kind of situation.

**Describe alternatives you've considered**

[...]

PR #264, opened 2019-08-03
Support dynamic custom props

jamesplease:
This PR is an example implementation that would add the feature described in #262 . 

**Pros**

By supporting dynamic `custom` props, users can leverage the declarative component API to perform animations based on "states" (like open/closed) **as well as** arbitrary values (using `custom`) at the same time.

**Cons**

Consumers passing non-primitives as `custom` would be required to `useMemo` them to prevent unnecessary calls to `controls.start()`, which could cause performance issues.

**Improvements that could be made to this PR**

Ideally, `custom` should only be considered as a dependency if one of the variants is defined as a function. Otherwise, it should be ignored. This would mitigate, but not completely resolve, the con described above.

---

Here are some CodeSandboxes showing why this might be useful.

1. [Using AnimationControls using the existing API](https://codesandbox.io/s/framer-motion-imperative-2-bdxo5)
Question: is it bad that you have to call `controls.start` twice in this example, or is that fine?

2. [Using two components to do this declaratively using the existing API](https://codesandbox.io/s/framer-motion-two-components-9exvt)
obviously not great; just including it here as another example

3. [The solution using the code from this PR](https://codesandbox.io/s/framer-motion-pr-solution-rkj7p)
With the code in this PR you can accomplish this using one component.

### DIFF
--- a/src/animation/use-variants.ts
+++ b/src/animation/use-variants.ts
@@ -25,7 +25,8 @@ export function useVariants(
     initial: VariantLabels,
     animate: VariantLabels,
     inherit: boolean,
-    controls: ValueAnimationControls
+    controls: ValueAnimationControls,
+    custom: any
 ) {
     let targetVariants = resolveVariantLabels(animate)
     const context = useContext(MotionContext)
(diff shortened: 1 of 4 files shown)


jamesplease:
I'm closing this based on the discussion over in #262

PR #266, opened 2019-08-05
Events refactor

InventingWithMonster:
We originally wrote the `useEvent` while scoping out the limitations of hooks (specifically its unconditional nature) which led us to a specific architecture. Recently I was investigating adding external drag controls and I gave up as the architecture of deeply composed hooks obfuscated made it difficult to change the default behaviour of the drag component.

This refactor shallows event listener application and this has brought a number of changes.

`useEffect` is now `useDomEffect` as there's an upcoming React Flare hook called `useEffect` and if we were to expose this function I'd like it to be explicit about how it's applying events differently than React.

`usePointerEvents` is now `usePointerEvent`. It is solely responsive for two things:
1) It wraps native events with `EventInfo`
2) Maps pointer-specifc names to device-specific ones ie `pointerdown` -> `touchstart`.

By allowing `usePointerEvent` to only accept a single pointer event and mapping that to a single event name we remove all the merging logic, and rather than running a `useEffect` for every hypothetical event (previously 20 `useEffect` calls for every `usePointerEvents` call) we can just run a single `useEffect`.

Both `useDomEvent` and `usePointerEvent` have been reduced to single signature and instead also expose an imperative function called `addDomEvent` and `addPointerEvent`. These can be used for asynchronous application of events (for instance adding `pointermove` on a `pointerdown`), rather than overloading the hooks to conditionally return controls.

I've removed `safeWindow` as it's no longer needed. When I upgrade Library I'll have to reintroduce this there.

I've also removed many `useCallback`/`useMemo` etc. Newly created functions never trigger a rerender here, given they don't instigate much extra work I'm not convinced that we're saving any thing by memoising them except incurring overhead for dependency comparisons. These functions still get created with `useMemo` etc anyway, they just aren't retained.

The refactor has led to a 1kb bundle size reduction 25 -> 24kb


src/gestures/use-pan-gesture.ts
```diff
@@ -397,150 +394,118 @@ export function usePanGesture(
...
+            removeOnPointerUp && removeOnPointerUp()
         }
-    }, [])
-    let handlers: Partial<{ onPointerDown: EventHandler }> = { onPointerDown }
-    if (!(onPan || onPanStart || onPanEnd || onPanSessionStart)) {
-        handlers = {}
     }
 
-    return useConditionalPointerEvents(handlers, ref)
+    usePointerEvent(ref, "pointerdown", hasPanEvents && onPointerDown)
```
nvh: The third argument looks like a boolean to me, but it is `undefined | function` this tripped me up, is there a clearer notation for this?

nvh: Maybe a ternary operator to make it explicit?

`hasPanEvents ? onPointerDown : undefined`

src/gestures/use-pan-gesture.ts
```diff
@@ -397,150 +394,118 @@ export function usePanGesture(
...
+            warning(false, "onPointerUp fired without pointer session")
+            return
+        }
+
+        onPanEnd && onPanEnd(event, getPanInfo(transformPoint(info)))
+
+        session.current = null
+    }
+
+    function onPointerDown(
```
nvh: I no this isn't much different from what we did before, but isn't creating all these inline functions wasteful?

src/gestures/use-tap-gesture.ts
```diff
@@ -183,152 +179,98 @@ export interface TapHandlers {
...
-const usePropsRef = <T>(props: T) => {
-    const propsRef = useRef(props)
-
-    for (const key in props) {
-        propsRef.current[key] = props[key]
-    }
-
-    return propsRef.current
-}
+type RemoveEvent = () => void
```
nvh: Can this type be shared with the pan-gesture?

src/behaviours/use-drag.ts
```diff
@@ -0,0 +1,584 @@
...
+    const constraintsNeedResolution = isRefObject(dragConstraints)
+    const state = useRef<DragState>({
+        isDragging: false,
+        hasDragged: false,
+        currentDirection: null,
+        constraints: false,
+        handlers: {},
+    }).current
+
+    // TODO: Explain why we keep these in a ref
```
nvh: Let's turn this TODO into a comment explaining exactly that 😄

nvh: It seems like if this is not react state created by React.useState, it should not be called state

src/gestures/use-tap-gesture.ts
```diff
@@ -183,152 +179,98 @@ export interface TapHandlers {
...
-    props: TapHandlers & ControlsProp,
-    ref?: RefObject<Element>
-): undefined | { onPointerDown: EventHandler } {
+) {
+    const hasTapListeners = onTap || onTapStart || onTapCancel || whileTap
     const isTapping = useRef(false)
-    const { onTap, onTapStart, onTapCancel, whileTap, controls } = props
-    const propsRef = usePropsRef(props)
+
+    const pointerEventSubscription = useRef<RemoveEvent | undefined | null>(
```
nvh: This is not really a 'subscription' right? More like a `cancelPointerEventHandler`?

src/behaviours/utils/block-viewport-scroll.ts
```diff
@@ -1,16 +1,17 @@
...
-    (event: TouchEvent) => {
-        if (isViewportScrollBlocked) {
-            event.preventDefault()
-        }
-    },
-    { passive: false }
-)
+if (isBrowser) {
+    window.addEventListener(
+        "touchmove",
```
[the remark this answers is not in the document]
InventingWithMonster: Yeah I believe there was a reason for this - something like if you add the event listener on `touchstart`, you aren't then able to block subsequent `touchmove` events. But if you always had the listener on and toggled this boolean it would work.

I'd like to revisit this with a PR for conditionally blocking the viewport scroll based on scroll direction but that is a project in itself.

src/gestures/use-hover-gesture.ts
```diff
@@ -86,42 +90,33 @@ export function useHoverGesture(
...
 
-    const handlers = useMemo(
-        () => {
-            const onPointerEnter = (
-                event: MouseEvent | TouchEvent | PointerEvent
-            ) => {
-                if (isTouchEvent(event)) return
+    useDomEvent(
+        ref,
+        "mouseenter",
```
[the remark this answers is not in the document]
InventingWithMonster: Good point what I can do is move this to `usePointerEvent` and also pass through `EventInfo`

src/events/__tests__/utils/event-helpers.ts
```diff
@@ -1,31 +1,21 @@
 export const enableTouchEvents = () => {
     const originalEvents = {
         ontouchstart: window.ontouchstart,
-        ontouchmove: window.ontouchmove,
-        ontouchend: window.ontouchend,
```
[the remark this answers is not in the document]
InventingWithMonster: Previously we were checking for all three to check whether the device supported those events, whereas I've reduced them to `pointerdown` (and thus mocking them here)

src/gestures/use-pan-gesture.ts
```diff
@@ -360,35 +354,38 @@ export interface PanHandlers {
...
-): undefined
-export function usePanGesture(
-    handlers: PanHandlers
-): { onPointerDown: EventHandler }
 export function usePanGesture(
     { onPan, onPanStart, onPanEnd, onPanSessionStart }: PanHandlers,
-    ref?: RefObject<Element>
+    ref: RefObject<Element>
 ) {
+    const hasPanEvents = onPan || onPanStart || onPanEnd || onPanSessionStart
```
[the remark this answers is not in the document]
InventingWithMonster: We can bail out sooner, after all the hooks have resolved

src/gestures/use-pan-gesture.ts
InventingWithMonster: We were already creating them:

```jsx
useCallback(() => {}, [...deps])
```

Is still creating the function. The only performance benefit is not letting that lead to a re-render. But you also have to compare the deps, so it could actually lead to worse performance.

Now there's an early return this shouldn't be a problem but if it is we can look into other ways of loading things like this

src/behaviours/utils/block-viewport-scroll.ts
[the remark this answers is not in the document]
nvh: I think we should figure out what the exact reason was and add it as a comment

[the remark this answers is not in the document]
InventingWithMonster: I agree but that investigation doesn’t look like part of this PR to me. Nothing’s changed here apart from dropping the window mock server side.

src/gestures/use-pan-gesture.ts
```diff
@@ -360,35 +355,55 @@ export interface PanHandlers {
...
+    function useAttachPanListeners(
+        cancelPanHandler: RemoveEvent,
+        onPointerDownHandler?: EventHandler
+    ) {
+        usePointerEvent(ref, "pointerdown", onPointerDownHandler)
+        useUnmountEffect(cancelPanHandler)
+    }
+
+    // Early return if we have no pan events
+    // Note: No more hooks other than useAttachPanListeners permitted past here.
```
nvh: This seems to be a bit risky. how do we ensure that this is enforced? Is there a test covering both paths?

src/gestures/__tests__/tap.test.tsx
```diff
@@ -68,6 +69,7 @@ describe("tap", () => {
     })
 
     test("tap event listeners are cleaned up when unmounted", () => {
+        console.log("=================== unmounted")
```
nvh: These don't need to log

src/gestures/use-pan-gesture.ts
InventingWithMonster: True but I think it's the only way to early return. I'll revert it because you're right and it's tricky to test the alternative path.

InventingWithMonster: Actually thinking about it we do test the early return path because it uncovered a bug where I didn't actually `return` - but I still think we should revert as it's an odd pattern.

PR #267, opened 2019-08-05
Introduce invalidateAnimate prop

### LINKED ISSUE #262: [FEATURE] Improved support for managing both variant-style animations as well as arbitrary-value style animations at the same time

jamesplease:
This PR is an effort to resolve #262 . It supersedes #264 

#### To do

- [x] Ensure `animateKey` doesn't pass through to the underlying DOM node
- [x] Add code comment for `animateKey` within `types.ts`, and possibly move it within the interface (I just put it at the top for now)
- [x] Update `api/framer-motion.api.md` (it looks like this may be automatically generated?)
- [x] Rename to `invalidateAnimate`
- [x] Update docs to use `@library` and `@motion`
- [x] Update `useAnimateProp`

#### Separate PRs

- Update the website (https://github.com/framer/api-docs)
  - Document the existence of `animateKey` and how it can be used
  - Explicitly state that changes to `custom` do not trigger rerenders, and mention how `animateKey` can be used to accomplish that
- Create some examples (this repo)


jamesplease:
@InventingWithMonster , I updated this PR with some changes.

- `animateKey` was added to `valid-props.ts`, ensuring that it's stripped out before being passed into the underlying DOM node
- `api/framer-motion.api.md` was updated (automatically). I'm not sure if that's all that needs to be done here?
- I added some placeholder code comments for `animateKey` within `types.ts`. Some further refinement is probably needed. I'm not sure where these comments show up to consumers, or what the significance of the `@library` / `@motion` sections is.

Also, I'm thinking of creating some examples in a separate PR.

Lmk what ya think!

InventingWithMonster:
Hey man, looks great! Thanks for this.

There's one more place we need to invalidate which is the `useAnimateProp` hook: https://github.com/framer/motion/blob/master/src/animation/use-animate-prop.ts

I'm not sure this one will work just to pass through the dependency array. We might need to keep a ref to the previous one and then if they're different, consider this in the check here: https://github.com/framer/motion/blob/master/src/animation/use-animate-prop.ts#L81

I'm happy to go over the docs once it's merged. `@motion` is examples for the Framer Motion docs and we re-export everything through the Framer Library so `@library` tends to have examples written with `Frame` instead of `motion` components. Also with simplified syntax for the designer crowd.

We're talking about another name for this prop as the thing about `animateKey` is that React's `key` in general is to keep things the same - it's quite rare we abuse it to force a component to re-render. Something like `invalidateAnimate` might be clearer about intent. But this won't hold up the PR, if you have any ideas I'm happy to hear them!

jamesplease:
Thanks for the thorough review @InventingWithMonster !

> There's one more place we need to invalidate which is the useAnimateProp hook: 

:+1: Sounds good – I'll add that!

> I'm happy to go over the docs once it's merged. @motion is examples for the Framer Motion docs and we re-export everything through the Framer Library so @library tends to have examples written with Frame instead of motion components. Also with simplified syntax for the designer crowd.

Ah, alright, that makes sense. Thanks for the explanation.

> Something like invalidateAnimate might be clearer about intent.

I think this is a much better name.

src/animation/use-animate-prop.ts
```diff
@@ -59,6 +60,14 @@ export function useAnimateProp(
         prevValues.current = targetWithoutTransition(targetAndTransition, true)
     }
 
+    const currentInvalidateAnimate = useRef()
+    const prevInvalidateAnimate = useRef()
+
+    currentInvalidateAnimate.current = invalidateAnimate
+    useEffect(() => {
+        prevInvalidateAnimate.current = invalidateAnimate
+    })
```
jamesplease: There are a few different ways to write this. Let me know if you'd prefer a different approach!

src/animation/use-animate-prop.ts
```diff
@@ -87,7 +96,17 @@ export function useAnimateProp(
                     finalTarget[key]
                 )
 
-                if (isValidValue && (valueHasUpdated || shouldAnimateOnMount)) {
+                const isAnimateInvalid =
+                    !isInitialRender.current &&
+                    currentInvalidateAnimate.current !==
+                        prevInvalidateAnimate.current
```
jamesplease: If the user supplies an `invalidateAnimate`, then the animation is always invalid on the first render, since `prevInvalidateAnimate` starts off as `undefined`. So we ignore the first render.

jamesplease:
@InventingWithMonster , some updates:

- this branch was rebased off of master
- renamed `animateKey` to `invalidateAnimate`
- updated `useAnimationProp`

Let me know what you think!

Also, should I take a stab at writing some unit tests?

src/animation/use-animate-prop.ts
InventingWithMonster: I think it's possible just to initialise this 

```jsx
prevInvalidateAnimate = useRef(invalidateAnimate)
```

Check it on line 99

```jsx
const isAnimateInvalid = prevInvalidate.current !== invalidateAnimate
```

And then update it on line 110

```jsx
prevInvalidateAnimate.current = invalidateAnimate
```

InventingWithMonster:
> @InventingWithMonster , some updates:
> 
> * this branch was rebased off of master
> * renamed `animateKey` to `invalidateAnimate`
> * updated `useAnimationProp`
> 
> Let me know what you think!
> 
> Also, should I take a stab at writing some unit tests?

Looks great! Just one comment. And yeah tests would be great - probably needs one in `animate-prop.test.tsx` and another in the variant file

src/animation/use-animate-prop.ts
jamesplease: That's a better approach. I'll update it!

src/animation/use-animate-prop.ts
jamesplease: That's a better approach. I'll update it!

yuhr:
I do need this feature definitely. Is there anything I can help?

mattgperry:
Sorry I've gone back and forth on this and every time I think about merging there's some internal refactor for deciding when to animate that changes my mental model.

I'm going to close this without merging though. There's two reasons for this:

1. In `3.x`, when a `while` gesture is active and the `while` prop changes, we animate to this new state. Do we add a dependency/key prop for all of those too? I feel like it's weird either way.
2. The API already has a mechanism to achieve this, with the `useAnimation` hook.

@yuhr You can decide when to start an animation without this feature:

```javascript
const controls = useAnimation()

useEffect(() => {
  controls.start({ x: 100 })
}, [invalidateAnimate])

return <motion.div animate={controls} />
```

PR #268, opened 2019-08-06
Layout transition

InventingWithMonster:
Adds a `layoutTransition` prop.

This supercedes the `positionTransition` prop by also accommodating changes to size. Thus we can declaratively animate layouts.


jonastreub:
Would be nice to have some high fidelity examples at some point.

src/motion/functionality/layout.ts
```diff
@@ -0,0 +1,274 @@
...
+        return { left, top, width, height }
+    },
+}
+
+function readPositionStyle(element: HTMLElement) {
+    return window.getComputedStyle(element).position
+}
+
+function getLayoutType(prev: string | null, next: string | null): LayoutType {
+    return prev === next ? offset : boundingBox
```
nvh: I don't understand why this difference exists? Why do we use offset when they are the same?

InventingWithMonster: They aren't the same.`offset` is calmer because it doesn't account for transforms, so for instance if we always used bounding box it takes more work to figure out if an element has changed size. But it's also not a reliable measure when switching between `position` types - but in these instances we can also presume that size has changed so it's ok to use bounding box.

rijk:
Awesome! I’ll try this out right away.

rijk:
When will 1.6.0 be released to NPM?

rijk:
When will this be released to NPM? When I try to install from Github (never done this before, apologies if it's my error) with `npm install framer/motion`, I get an almost empty directory:

image

Do you know if there's anything else I need to do to install framer-motion from Github?

rijk:
Never mind, figured it out: `npm install framer-motion@beta` :)

rijk:
All right, I have run into a problem with `useInvertedScale()`. As I understand it, this feature was implemented using a scale transform (as opposed to `height` + `overflow: hidden`) for performance reasons. However, this makes it difficult to build a simple Collapse component that will grow when stuff is added (without distorting the content, just gradually adding size at the bottom). When using a child component with `useInvertedScale()`, you get the following effect:

ezgif-3-c7d63300f704

So scale is indeed inverted, but the content also moves in the Y direction. Is there a way to cancel this out as well, so that the content visually remains in the same position on screen, and just the container?

InventingWithMonster:
> So scale is indeed inverted, but the content also moves in the Y direction. Is there a way to cancel this out as well, so that the content visually remains in the same position on screen, and just the container?

Hi Rijk

Can you open an issue with a CodeSandbox example? Thanks

rijk:
Hey @InventingWithMonster — first off really great work. I already found the answer to my previous question: applying `transform-origin: top;` to the inverter div does the trick (although there is a teeny tiny subpixel size shift visible when the transition finishes and goes back to `none` — probably because the element goes from GPU->CPU mode in the browser?). 

If I run into any other problems I'll set up a CodeSandbox, thanks!

rijk:
All right, here’s a Sandbox (if you want me to open separate issues for this, let me know): https://codesandbox.io/s/framer-motion-sandbox-c7urd

Here you can see that due to the transform approach I cannot achieve the desired effect of having an animated auto height; the next comment instantly shoots into place, while the replies container is still animating.

I have also found that with longer content, the inversion isn't always exactly in sync with the parent (haven't managed to reproduce this in the Sandbox, but here's a screen recording):

ezgif-1-3ccc9ad12e73

rijk:
I think, at least for this use case, it would be better to just animate `height` instead of the transform + invert approach. Or is this not what this feature was meant for? What do you think?

Issue #293, opened 2019-08-25
[BUG] `exit` doesn't work when triggered from `onDragEnd`


lintonye:
**Describe the bug**
`exit` animation doesn’t play if triggered from `onDragEnd`

**To Reproduce**
1. Go to 'https://codesandbox.io/s/task-list-3684r'
2. Drag any item to the right

**Expected behavior**
The item should fade out since there's an `exit` prop set on the div that animates its opacity.

Clicking the "remove" button, you'll see the expected animation.

**Screenshots**
video: https://framer-team.slack.com/archives/G5VQ146US/p1563386007188000

**Desktop (please complete the following information):**
 - OS: macOS 10.14.4
 - Browser: Chrome Version 76.0.3809.100 (Official Build) (64-bit)

InventingWithMonster:
Thanks for the bug report Linton, very clear example. I'll take a look at this one next.

InventingWithMonster:
What's happening here is `AnimatePresence` adds an `onAnimationComplete` to know when the exit animation has completed.

However, the way `exit` works is by internally overwriting the `animate` prop, and the way `onAnimationComplete` works is by firing when *any* animation completes.

In the included example, if you throw hard enough the exit animation behaves as expected. But if dragging ends without momentum, the drag end animation completes immediately, triggering `onAnimationComplete` before the exit animation has completed.

On solution may be to change `applyExitProps` to `useExitProps` or similar, and have that accept the component's `ValueAnimationControls`. Instead of overwriting `exit` with `animate`, we'd use a `useEffect` to fire the `exit` animation using the `ValueAnimationControls` directly rather than "tricking" the associated animation handler to run its normal comparison algo. This would also fix https://github.com/framer/motion/issues/276

InventingWithMonster:
This will be fixed in `1.6.8` next week

Issue #276, opened 2019-08-09
[BUG] AnimatePresence does not work under certain conditions


jamesplease:
**Describe the bug**

I expect `AnimatePresence` to use the `exit` prop to animate a motion component when it unmounts. Most of the time it does this, but you can configure a series of props that breaks that functionality.

What happens instead is that you try to unmount the components but they do not animate to `exit` and instead stay mounted indefinitely.

**To Reproduce**

1. You must be using animation controls on the component
2. You must have `variants` defined on the component
3. You must have an `initial` prop set on the component and it must be a variant string

This combination of factors will break `AnimationControls`.

- If you remove `initial`, then the problem does not occur
- If you define `initial` using an animate object, then the problem also does not occur

[**View an isolated CodeSandbox here**](https://codesandbox.io/s/broken-animate-presence-z1j78)

To give myself some confidence that I actually understood AnimationControls, I made another CodeSandbox that has some functioning versions of this Box component, too. You can see that [here](https://codesandbox.io/s/gifted-gould-fic4o).

**Expected behavior**

I would expect `AnimatePresence` to work even when the above 3 conditions are met.

**Desktop (please complete the following information):**
 - OS: macOs 
 - Browser: Chrome
 - Version: 75

**Additional context**
I can try and dig into the code more to learn what's happening here, but it might take me some time to get an understanding of how AnimationControls works under-the-hood.

InventingWithMonster:
We just ran into this too! Animate controls aren’t properly supported with
exit yet, I’ve been holding off on tackling it. There’s either the scenario
where it accepts props/variant labels and we just pass that to any existing
controls. Or it could be a function that’s provided the controls and
returns a promise that resolves once the animation is complete. This might
be better for
More intricate series of animations but it might also not work as expected
as exiting components don’t receive the latest props from React (as they’ve
been deleted from the tree already)

jamesplease:
I might have some time to look more into this over the next few days. In the meantime, I plan to write a resolver that reduces a variants list to an animate object as a workaround. That way I can still organize my animations using variants without running into this issue.

i.e.;

```js
const variants = {
  open: { ... },
  closed: { ... },
  big: { ... },
  small: { ... }
};

const initial = resolveVariants(variants, ['open', 'big']);
// => {
//   x: 200,
//   y: 100,
//   opacity: 1
// }

return (
  <motion.div animate={controls} initial={initial} />
);
```

[...]

PR #303, opened 2019-08-30
Exit animation fixes

InventingWithMonster:
Fixes https://github.com/framer/motion/issues/276
Fixes https://github.com/framer/motion/issues/293

The `exit` prop is used to define an animation state to enter once a component has been removed as a child of an `AnimatePresence` component.

```jsx
<AnimatePresence>
  {isVisible && <motion.div animate={{ opacity: 1 }} exit={{ opacity: 0 }} />}
</AnimatePresence>
```

Currently it works by, if an `exit` prop is found and the component is exiting, we internally overwrite the component's `animate` prop with it. The component's current animation reconciler (either variants or object) sees a change in `animate` and triggers an animation.

This worked well when consistently using variants or objects, but when `animate` was a `useAnimation()` controls there could be weird behaviour.

```jsx
// Good
<motion.div animate={{ opacity: 1 }} exit={{ opacity: 0 }} />

// Good
<motion.div animate="visible" exit="hidden" variants={variants} />

// Probably bad?
<motion.div animate="visible" exit={{ opacity: 0 }} variants={variants} />

// Definietly bad
<motion.div initial="hidden" animate={controls} exit="hidden" variants={variants} />
```

In this PR, we instead handle `exit` as a special case. By not hijacking the current `animate` handler, we can make `exit` work with any value currently in `animate`, even animation controls.

This technique also fixes a second bug. Before, `AnimatePresence` passed an `onExitComplete` callback via context. This was then monkey-patched into the component's `onAnimationComplete` function. `onAnimationComplete` is fired when *any* animation completes. So if another animation completed before the exit animation (for instance drag end momentum animation), the component would fire `onExitComplete` too early.

By explicitly handling exit animations we can also avoid monkey-patching `onAnimationCompelete` and instead firing *only* `onExitComplete`. This way the component can handle any other animations separately and can know for a fact that it's the *exit* animation that has completed.


amannn:
I also just encountered the bug with `animate={useAnimation()}`.

I wanted to test `framer-motion@1.6.8` in the project where I had the issue and found that it fixes the exit issue, but introduces a different bug: The children of `AnimatePresence` don't unmount anymore.

See this CodeSandbox: https://codesandbox.io/s/framer-motion168-animate-presence-bug-viqo1.

Note that I published the 1.6.8 version of `framer-motion` under the name `framer-motion-amannn` since the version isn't released yet.

noisyneuron:
I'm also having an issue with the exit animation.. I can't seem to understand when it will or will not work. 

This [CodeSandbox](https://codesandbox.io/s/serene-currying-usgof) is a bit of a contrived example -- it randomly adds or removes an item, just to test the animation. 

I've tried whatever suggestions I could find (don't include `initial`, five the `motion.div` a key, etc), but no luck. Any help would be much appreciated! Is there something silly I'm missing or have misunderstood?

Issue #290, opened 2019-08-22
[BUG] motion.custom should accept type parameters


iamkevinwolf:
The `motion.custom` component should accept type parameters for passing the custom component props.

Having the following base component:

```tsx
import styled from '@emotion/styled'
import * as system from 'styled-system'

export type BoxProps = system.BackgroundColorProps &
  system.BorderProps &
  system.FlexboxProps &
  system.LayoutProps &
  system.OpacityProps &
  system.SpaceProps

export default styled.div<BoxProps>`
  ${system.compose(
    system.border,
    system.color,
    system.flexbox,
    system.layout,
    system.space
  )};

  box-sizing: border-box;
`
```

Wrapping the custom component with `motion.custom`:
```tsx
import Box from 'components/box'

const Container = motion.custom(Box)
```

And using the motion wrapped component passing my custom component original props:

```tsx
<Container
  minHeight={viewportHeight}
  display="flex"
  flexDirection="column"
  alignItems="center"
  justifyContent="center"
  opacity={viewportHeight === 0 ? 0 : 1}
  p={2}
>
  Test
</Container>
```

Gives the following TypeScript error:
```
Type '{ children: (string | number | boolean | {} | ReactElement<any, string | ((props: any) => ReactElement<any, string | ... | (new (props: any) => Component<any, any, any>)> | null) | (new (props: any) => Component<...>)> | ... 4 more ... | undefined)[]; ... 6 more ...; p: number; }' is not assignable to type 'IntrinsicAttributes & MotionProps & RefAttributes<Element>'.
  Property 'children' does not exist on type 'IntrinsicAttributes & MotionProps & RefAttributes<Element>'.ts(2322)
```

So my solution is to just accept a generic type parameter in `motion.custom` to avoid this errors:

```tsx
import Box, { BoxProps } from 'components/box'

const Container = motion.custom<BoxProps>(Box)
```

petecorreia-appital:
Ran into this issue too and I believe a good fix could be:

```ts
// framer-motion.d.ts — line: 1953

custom: <Props = {}>(Component: ComponentType<Props>) => ForwardRefExoticComponent<MotionProps & RefAttributes<Element> & Omit<ComponentProps<typeof Component>, keyof MotionProps> & { children: ReactNode }>;
```

**Highlights**:
* Infers the props of the passed component
* Makes sure the motion prop definitions take precedence over the inferred props
* Adds children type

**This would need no further changes to the `motion.custom` API and would _just work_.**

---

[...]

PR #326, opened 2019-09-18
Adding type support for motion.custom

InventingWithMonster:
Fixes https://github.com/framer/motion/issues/290

Allows the ability to extend `MotionProps` with custom component props when using `motion.custom`.

```
motion.custom<Props>(Component)
```


api/framer-motion.api.md
```diff
@@ -731,6 +732,10 @@ export type Variants = {
 };
 
 
+// Warnings were encountered during analysis:
+// 
+// types/motion/index.d.ts:217:5 - (ae-forgotten-export) The symbol "custom" needs to be exported by the entry point index.d.ts
+
```
nvh: I think this warning is valid. It will work in the test, but when used imported from motion, it will just have type any

PR #327, opened 2019-09-18
Support clicks within draggable elements on iOS Safari

### LINKED ISSUE #322: [BUG] Click handlers don't fire inside draggable elements

**2. Describe the bug**

When you have a draggable motion element and you have clickable elements inside, the click handlers aren't called on iOS.

Note that they are called on Chrome on Android though.

**3. Provide a CodeSandbox reproduction of the bug**

https://codesandbox.io/s/framer-motion-derived-state-bug-p6q20

**4. Steps to reproduce**

1. Go to https://codesandbox.io/s/framer-motion-derived-state-bug-p6q20 on your iOS device.
2. Try to click on the button.

**5. Expected behavior**

The click handler fires.

**6. Video or screenshots**

N/A

**7. Environment details**

iOS Simulator on OSX running iPhone X on iOS 12.4.

amannn:
Fixes #322 .

I found out that when you don't `preventDefault` the `touchstart` event, Safari will try to scroll the window. I found your `blockViewportScroll` utility, but it didn't seem to work. Apparently you need to register the event handler on `document` instead of `window`. Now it works properly in my simulator. Not sure if there's a case for also handling events from `window`?

For the check if the device supports touch events, I used the existing function in `use-pointer-event`. Is that ok?

### DIFF
--- a/src/behaviours/use-drag.ts
+++ b/src/behaviours/use-drag.ts
@@ -20,7 +20,7 @@ import {
 import { useMotionValue } from "../value/use-motion-value"
 import { DraggableProps, DragHandlers } from "./types"
 import { useUnmountEffect } from "../utils/use-unmount-effect"
-import { supportsTouchEvents } from "../events/use-pointer-event"
+import { supportsTouchEvents } from "../events/utils"
 
 type DragDirection = "x" | "y"
 
@@ -388,6 +388,8 @@ export function useDrag(
             event.target &&
             !allowDefaultPointerDown.has((event.target as Element).tagName)
         ) {
+            // On iOS it's important to not `preventDefault` the `touchstart`
+            // event, as otherwise clicks won't fire inside the draggable element.
             if (!supportsTouchEvents()) {
                 // Prevent browser-specific behaviours like text selection or Chrome's image dragging.
                 event.preventDefault()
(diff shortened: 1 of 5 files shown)


src/behaviours/use-drag.ts
```diff
@@ -383,15 +384,18 @@ export function useDrag(
...
-        // Prevent browser-specific behaviours like text selection or Chrome's image dragging.
         if (
             event.target &&
             !allowDefaultPointerDown.has((event.target as Element).tagName)
         ) {
-            event.preventDefault()
-            // Make sure input elements loose focus when we prevent the default.
-            if (document.activeElement instanceof HTMLElement) {
-                document.activeElement.blur()
+            if (!supportsTouchEvents()) {
```
InventingWithMonster: Could you add a note as to why we're adding this text (describe the bug we're avoiding)

src/events/use-pointer-event.ts
```diff
@@ -5,7 +5,8 @@ import { wrapHandler, EventListenerWithPointInfo } from "./event-info"
 // We check for event support via functions in case they've been mocked by a testing suite.
 const isBrowser = typeof window !== "undefined"
 const supportsPointerEvents = () => isBrowser && window.onpointerdown === null
-const supportsTouchEvents = () => isBrowser && window.ontouchstart === null
+export const supportsTouchEvents = () =>
```
InventingWithMonster: Could you move these `supports` functions to a `events/utils` file?

amannn:
Thanks for the feedback! I've added a commit which addresses both points.

InventingWithMonster:
Excellent, thanks!

amannn:
Thanks for the review and merge! Looking forward to the release 🙂

akramabdulrahman:
having the same problem in 2.9.5

Issue #319, opened 2019-09-08
Sidebar example doesnt work in safari [BUG]


umxr:
open https://codesandbox.io/s/framer-motion-side-menu-mx2rw?fontsize=14&module=%2Fsrc%2FExample.tsx in safari. The reason is due to clip-path not having the -webkit prefix. is this possible to do as the style gets applied inline

goldmont:
Same here.
The issue seems to be `motion.div` related.

PR #328, opened 2019-09-18
Fixing clipPath in webkit

InventingWithMonster:
Fixes https://github.com/framer/motion/issues/319

Webkit incorrectly detects `element.style.clipPath` as supported, when it isn't. So an update to `stylefire` ensures the un-prefixed version doesn't take precedence.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 Framer Motion adheres to [Semantic Versioning](http://semver.org/).
 
+## [1.6.9] Unreleased
+
+## Fix
+
+-   Upgrade `stylefire@6.0.11` to fix `clipPath` in Webkit.
+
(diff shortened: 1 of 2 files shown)


DynamicArray:
Should this be true for the very latest version of motion? This demo https://codesandbox.io/s/framer-motion-side-menu-mx2rw works on iOS Safari 12, but taking the code into my project fails, because clip-path is not prefixed (at least anywhere I can see) in the build code.

DynamicArray:
Oddly, my code works if I code the variants like (for example) this:
`
clipPath: "circle(29px at 40px 40px)",
-webkit-clip-path": "circle(29px at 40px 40px)",
`

with the specific webitkit prefix. And yet your example works without it.... would love to know why! :)

PR #335, opened 2019-09-24
Add failing test case for nested components inside StrictMode

amannn:
I encountered an error in an app that I'm working on and isolated it to the test case in this PR.

When you run the test, you get this error:

```
Uncaught Error: No valid node provided. Node must be HTMLElement, SVGElement or window.
    at Object.exports.invariant (index.js:15)
    at createDOMStyler (index.js:536)
    at getStyler (index.js:541)
    at index (index.js:547)
    at checkAndConvertChangedValueTypes (framer-motion.cjs.js:3198)
    at unitConversion (framer-motion.cjs.js:3295)
    at ValueAnimationControls.eval [as makeTargetAnimatable] (framer-motion.cjs.js:3304)
    at ValueAnimationControls.animate (framer-motion.cjs.js:1269)
    at getAnimations (framer-motion.cjs.js:1313)
    at ValueAnimationControls.animateVariant (framer-motion.cjs.js:1337)
    at eval (framer-motion.cjs.js:1354)
    at Array.forEach (<anonymous>)
    at ValueAnimationControls.animateChildren (framer-motion.cjs.js:1353)
    at getChildrenAnimations (framer-motion.cjs.js:1317)
    at ValueAnimationControls.animateVariant (framer-motion.cjs.js:1337)
    at eval (framer-motion.cjs.js:1301)
    at Array.map (<anonymous>)
    at ValueAnimationControls.animateVariantLabels (framer-motion.cjs.js:1301)
    at ValueAnimationControls.start (framer-motion.cjs.js:1243)
    at eval (framer-motion.cjs.js:3785)
```

You can work around the error by:
1. Removing `StrictMode`.
2. Changing `animate={useAnimation()}` to `animate="visible"`.
3. Animating `opacity` instead of `y` in the inner component.

The stack trace goes into `stylefire`. I thought maybe you have an idea where the bug could originate from?

Maybe there's an [unexpected side effect](https://reactjs.org/docs/strict-mode.html#detecting-unexpected-side-effects)? If so, it could be helpful to fix this in order to work with the upcoming concurrent mode of React.

### DIFF
--- a/src/motion/__tests__/component.test.tsx
+++ b/src/motion/__tests__/component.test.tsx
@@ -312,16 +312,14 @@ describe("motion component rendering and styles", () => {
                 <React.StrictMode>
                     <motion.div
                         animate="visible"
-                        exit="hidden"
                         initial="hidden"
                         variants={{
                             visible: { y: 0 },
                             hidden: { y: 5 },
                         }}
                     >
-                        <motion.div
+                        <motion.span
                             animate={useAnimation()}
-                            exit="hidden"
                             initial="hidden"
                             variants={{
                                 visible: { y: 0 },
--- a/src/motion/__tests__/component.test.tsx
+++ b/src/motion/__tests__/component.test.tsx
@@ -1,6 +1,7 @@
 import "../../../jest.setup"
 import { render, fireEvent } from "@testing-library/react"
 import { motion } from "../"
+import { useAnimation } from "../../animation/use-animation"
 import * as React from "react"
 import styled from "styled-components"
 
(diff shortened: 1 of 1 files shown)


InventingWithMonster:
Great PR. It looks to me like an animation can be invoked in strict mode before the component has mounted.

amannn:
I had another look.

So the main issue is that here:

https://github.com/framer/motion/blob/df6041d41ceff4f28b28f70e8e098aba46f8f76f/src/dom/unit-type-conversion.ts#L166

… the element is `null`. Note that in the test we render an outer and an inner motion component. In the outer component `element` is defined, but it's the inner one where it's `null`.

It seems like in my test the variant changes from `[]` to `['visible']` in the nested component.

In this code:

https://github.com/framer/motion/blob/df6041d41ceff4f28b28f70e8e098aba46f8f76f/src/animation/use-variants.ts#L45-L49

`shouldAnimate` is `true`, because the variant has changed.

Is it intentional that it's `hasMounted.current ||`? If I replace `||` with `&&`, my test passes. However it breaks a few tests of `useVariants`.

I'm struggling a bit to understand what `hasMounted` expresses in the `useVariant` hook. It's set to `true` once the `useEffect` hook has run for the first time. Is this because `useEffect` fires in children first and later in the parent?

I feel like the inheritance of variants get's a bit in the way in this case, so I also opened https://github.com/framer/motion/issues/343 which is something I've encountered in a different scenario as well.

amannn:
I just did a quick check: Passing `inherit={false}` to the nested component also avoids the issue.

Issue #344, opened 2019-10-03
[BUG] X, Y values not being registered on SVG elements


Mellis84:
Hey Guys,

I can't seem to translate SVG elements, I can scale and rotate but translating from X or Y doesn't seem to work.

The following code I have below, I use a pressed object for my whileTap method on one of my variants to translate:
```
const container = {
  initial: {
    scale: 1
  },
  pressed: {
    scale: 0.8
  },
  hover: {
    transition: {
      delay: 0.5,
      staggerChildren: 0.3
    }
  }
};

const moonVariants = {
  pressed: {
    x: 40,
    y: 50,
    rotate: 50
  }
};

const ThemeTrigger = () => {
  const [hovered, setHovered] = useState(false);
  return (
    <button onClick={() => console.log("clicked")}>
      <motion.svg
        variants={container}
        animate={hovered ? "hover" : "initial"}
        onHoverStart={() => setHovered(true)}
        onHoverEnd={() => setHovered(false)}
        whileTap="pressed"
        viewBox="0 0 78.19 69.82"
        height="100%"
        width="100%"
      >
        <motion.path
          d="M61.92,37.65A21.37,21.37,0,0,1,48.07,0,26.89,26.89,0,1,0,78.19,30.12,21.31,21.31,0,0,1,61.92,37.65Z"
          fill="#ffffff"
          variants={moonVariants}
        />
      </motion.svg>
    </button>
  );
};
```

InventingWithMonster:
Hey there

This is because there's the attr `x` and the transform `translateX`. For HTML we use the former as a shortcut for the latter. But in SVG you have to use the latter for now. I'd like to come up with a better way of handling it as I think most ppl want the transform.

InventingWithMonster:
We're considering this a bug and in `1.6.10` `x` will be standardised to `translateX` across all elements.

PR #350, opened 2019-10-08
Make x/y consistent across HTML and SVG attributes

InventingWithMonster:
Fixes https://github.com/framer/motion/issues/344

This PR upgrades `stylefire` to `7.0.0` which changed its behaviour to treat SVG `x` and `y` elements as `translateX` and `translateY` respectively: https://github.com/Popmotion/popmotion/pull/839/files#diff-1734860cf394ce48efe4992634e89d13L51

This means that it won't be possible to animate `x` and `y` SVG attributes but these only exist on a handful of elements which has been causing confusion and perceived bugs. We're going to keep an ear to the ground if this is functionality that people want but in the meantime we've decided to prefer animating transforms over layout. Given the number of bug reports and behaviour I've run into myself this is considered a bug rather than a breaking change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 Framer Motion adheres to [Semantic Versioning](http://semver.org/).
 
+## [1.6.11] 2019-10-09
+
+### Fix
+
+-   Inconsistency in handling `x`/`y` between SVG and HTML. Now always a shorthand for `translateX` and `translateY`.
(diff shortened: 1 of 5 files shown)


InventingWithMonster:
Yeah it was there that we decided how to handle x/y. I've added `attrX`/`attrY` there for the attributes and we can expose this however we like, or not at all, in the future.

Issue #341, opened 2019-09-30
[BUG] AnimatePresence with motion.div containing a ref blows up


mattcolman:
**2. Describe the bug**
When a child of AnimatePresence contains a ref there is a bug that occurs.
"Uncaught Error: No valid node provided. Node must be HTMLElement, SVGElement or window."

**3. Provide a CodeSandbox reproduction of the bug**
https://codesandbox.io/s/framer-motion-animate-presence-6ksyh

**4. Steps to reproduce**
1. Click 'add new' TWICE
2. Click clear
3. Notice the error
4. *Note* if you remove the ref from the codesandbox example it works fine.

**5. Expected behavior**
There should be no error

**7. Environment details**
MacOS latest Chrome

Thanks!!

amannn:
I'm wondering if this is related to https://github.com/framer/motion/pull/335. The error seems to be the same, not sure if the cause is related.

InventingWithMonster:
This is an interesting one - I'm going to look into it further but I suspect it has something to do with the fact that although we're rendering "one" component at a time we're actually rendering multiple components until the exiting ones animate out. By providing `ref`, rather than allowing them to all hold an internal ref to their own element we're forcing them to share one. Which could be leading to a host of unexpected behaviours. I hope to resolve this but it *might* be impossible.

InventingWithMonster:
Fixed in `1.6.10`. @amannn could you try this version for your other issue, if it's not fixed I'll take a deeper look.

amannn:
@InventingWithMonster I've synced my fork where [the failing PR is open](https://github.com/framer/motion/pull/335) with the latest master, but it seems like there's still the same error.

PR #352, opened 2019-10-09
Fixing external ref with single-child AnimatePresence

InventingWithMonster:
Fixes https://github.com/framer/motion/issues/341

For a `motion` component to work, we need to maintain a `ref` to the component's rendered HTML/SVG element. This is because we animate outside the React render cycle, directly on the element itself.

Previously, we used a hook called `useExternalRef` to use an externally-provided `ref` if it existed, or one created internally in the `motion` component if it wasn't. This way users could provide their own `ref`s.

This led to a problem with `AnimatePresence` where it might be used to render one component at a time:

```jsx
<AnimatePresence>
  <motion.div ref={ref} key={id} />
</AnimatePresence>
```

In the above code, by changing `key`, we are actually creating new components. This is useful, for instance, when creating a slideshow. We can animate through the slides while only rendering one "canonical" slide (the current slide), rather than keeping track of incoming/outgoing slides.

However this led to a situation, the specifics around are unclear but the gist is when a user provided this component a ref, it would technically be possible for two or more components to all share the same ref as new components were added as old ones animated out. This created race conditions where outgoing components might set this ref to `null` while another component was still using it.

To fix this, this PR changes things so every component has its own `ref` and on mount/unmount defensively updates an external ref if it's provided.

### DIFF
--- a/src/motion/utils/use-external-ref.ts
+++ b/src/motion/utils/use-external-ref.ts
@@ -1,4 +1,5 @@
-import { useEffect, useRef, Ref, MutableRefObject, RefObject } from "react"
+import { useEffect, useRef, Ref, RefObject, MutableRefObject } from "react"
+import { isRefObject } from "../../utils/is-ref-object"
 
 /**
  * Uses the ref that is passed in, or creates a new one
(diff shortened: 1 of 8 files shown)


src/motion/utils/use-external-ref.ts
```diff
@@ -1,25 +1,39 @@
...
-            return () => external(null)
+            // If we've been provided a RefObject, we need to assign its current with our
+            // current on mount
+            mutableExternal.current = internalRef.current
+
+            return () => {
+                // We only set our external ref value to null on unmount if it still contains the
+                // same element as our internal ref. This is because the component might be a child
+                // of `AnimatePresence` where we might be in a situation where a user is providing
+                // the same ref to multiple components.
```
melissarofman: Really nice comment!

Issue #356, opened 2019-10-12
[FEATURE] Be able to send custom data to controls.set function


hassenc:
**Is your feature request related to a problem? Please describe.**
When using animation controls (`useAnimation`) we can pass custom props to `.start() `function using the `custom` prop. We can't do that for the `.set()` function.

**Describe the solution you'd like**
On button click, would like to reset an animation to a state that depends on some props
````
const onClick() => {
  controls.set( i => {{x: i*10}})
  controls.animate( i => {{x: i*20}})
}
````
**Describe alternatives you've considered**
We can use variantLabels instead.

Let me know if this seems a legit feature request, I can open a PR for this.

InventingWithMonster:
Yeah this sounds like a great feature request, totally understand the need for it. We'd take a PR for it.

PR #361, opened 2019-10-14
Support custom attributes in controls .set()

hassenc:
**Feature description**
This is the PR for #356 
It makes `controls.set` function accept a method that 's based on custom props.

**Separate PRs**
- No need for a separate PR. 
.set() method docs will be updated thanks to the typescript class changes.

### DIFF
--- a/src/animation/ValueAnimationControls.ts
+++ b/src/animation/ValueAnimationControls.ts
@@ -198,29 +198,29 @@ export class ValueAnimationControls<P extends {} = {}, V extends {} = {}> {
     /**
      * Set motion values without animation.
      *
-     * @param target -
+     * @param definition
      * @param isActive -
      */
     private setValues(
-        { transition, transitionEnd, ...target }: TargetAndTransition,
+        definition: Variant,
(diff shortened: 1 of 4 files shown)


src/animation/ValueAnimationControls.ts
```diff
@@ -198,29 +198,29 @@ export class ValueAnimationControls<P extends {} = {}, V extends {} = {}> {
...
     ) {
+        let { target, transitionEnd } = this.resolveVariant(definition)
         target = this.transformValues({ ...target, ...transitionEnd } as any)
-
         return Object.keys(target).forEach(key => {
             if (isActive.has(key)) return
-
             isActive.add(key)
-            const targetValue = resolveFinalValueInKeyframes(target[key])
+            if (target) {
```
InventingWithMonster: Would this not always be `true` given we're iterating over `Object.keys(target)`?

hassenc: Typescript needs this otherwise it throws a `TS2532: Object is possibly 'undefined'`

Issue #366, opened 2019-10-17
[BUG] Motion component refs not available during `componentDidMount`


xanido:
**1. Read the [FAQs](#faqs) 👇**

**2. Describe the bug**

Since v1.6.10, motion component refs are not available during `componentDidMount`.

**3. IMPORTANT: Provide a CodeSandbox reproduction of the bug**

v1.6.9 (working as expected): https://codesandbox.io/s/cool-microservice-borwk
v1.6.10 (not working as expected): https://codesandbox.io/s/amazing-dijkstra-ptnvs

**4. Steps to reproduce**

Load up v1.6.10 sandbox. Ref is undefined during `componentDidMount`

**5. Expected behavior**

Motion component ref should be available during `componentDidMount`

**6. Video or screenshots**

N/A

**7. Environment details**

Google Chrome 77.0.3865.120 on macOS 10.14.4

Issue #360, opened 2019-10-14
[QUESTION] Collect motion element dimensions


souporserious:
Hello 👋

I'm working on a button to input animation ([codesandbox](https://codesandbox.io/s/framer-motion-button-input-yj0si)) and I need to measure motion elements when the component mounts using `useLayoutEffect` to prevent flashing and animating on load. For some reason, `ref.current` is not available when the hook fires. I'm assuming because `motion` needs to know about it? Is there any way to grab the dimensions before the first paint? I appreciate any help 🙏.

InventingWithMonster:
Hey man

Sorry for the delay in response - we've been heads-down on a different project.

This is because of the unique way the `motion` component dynamically loads in functionality components. I'm going to look into this a little bit now and see if there's a way we can re-order some stuff to make this available earlier.

InventingWithMonster:
Ah it's because we load external refs manually in a useEffect: https://github.com/framer/motion/blob/master/src/motion/utils/use-external-ref.ts#L13

I'll have to look into why this change was made because there was a bug this fixed. Obviously we can't move this to useLayoutEffect because React will kick off during SSR.

InventingWithMonster:
I think the "fix" here might have to be to revert the change made in `1.16.10` and add a warning in components that are children of `AnimatePresence` that a unique ref must be given to each unique child. This can be solved by making the child a separate component that has its own call to `useRef`

There's no real other solution here as far as I can tell.

CC @mattcolman

souporserious:
Sounds good, Matt! I appreciate you looking into this, thank you! 🙏

souporserious:
Not to get off-topic, but maybe a first-class way to access dimensions could help? I know you need to do measurements in certain use-cases like FLIP, it might be nice to expose a hook to read dimensions at the same time framer does to save from having to read from the DOM twice.

InventingWithMonster:
That is a fantastic idea

PR #383, opened 2019-10-31
Fixing hydration of ref during mount useLayoutEffect

InventingWithMonster:
Fixes https://github.com/framer/motion/issues/360
Fixes https://github.com/framer/motion/issues/366

Un-fixes https://github.com/framer/motion/issues/341

## Background

In issue [#341](https://github.com/framer/motion/issues/341) there was a bug where Motion would blow up if we provided a single `ref` to the single child of `AnimatePresence`. This is because each component would use that externally-created `ref` to store a reference to the DOM node that they're animating.

To fix this, in PR [#352](https://github.com/framer/motion/pull/352) I made a change whereby every component always created its own `ref` and manually hydrated any externally-provided `ref`s via a `useEffect`. This meant we could guarantee that each component had a reference to its own underlying DOM node.

## The unforeseen consequence

The issues that this PR fixes is this:

```jsx
const MyComponent = () => {
  const ref = useRef(null)
  useLayoutEffect(() => {
    ref.current === null // This should be a DOM node
  })
  return <motion.div ref={ref} />
}
```

Because we're hydrating external refs with a `useEffect`, if a user wants to access the underlying node during a `useLayoutEffect`, the ref is still `null`.

## The fix

One approach would be to hydrate the ref using `useLayoutEffect`. However:
1. `useLayoutEffect` is render-blocking. This would have performance implications in Framer and on user's sites.
2. It'd have implications with SSR sites as React (IMO incorrectly) throws warnings for every component it encounters that uses `useLayoutEffect` in a server-rendered setting.

Instead, this PR reverts to the previous behaviour where external refs are, if present, used as the component's ref. On one hand this reintroduces the bug we were trying to fix. However, there's a counter-point in that this bug is valid and expected behaviour. When we provide a single `ref` to multiple components that exist simultaneously, this is a bug waiting to happen. Consider a `ref` function:

```jsx
<motion.div ref={() => {}} />
```

It's obvious when this should fire with an element, but when should it fire with `null`? IMO when the node is removed from the DOM, when the component *really* unmounts. So when this same `ref` is provided to a component that can switch out for another (for instance via `key` change), without some mind-bending logic and edge-case handling this `ref` will fire with the exiting component's `null` *after* the incoming component fires it with its DOM node. This logic holds for a normal ref:

```jsx
const MyComponent = ({ isOpen }) => {
  const ref = useRef(null)

  return (
    <AnimatePresence>
      <motion.div ref={ref} key={isOpen ? "a" : "b"} exit={{ opacity: 0 }} />
    </AnimatePresence>
  )
}
```

When should this `ref` represent "a", and when should it represent "b"? These time periods overlap. We shouldn't be trying to fix this and the previous PR was a mistake.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 Framer Motion adheres to [Semantic Versioning](http://semver.org/).
 
+## [1.6.16] 2019-10-31
+
+### Fixes
+
+-   Fixing `ref` hydration in `useLayoutEffect`. (Note: This release effectively reverts `1.6.10`. Each child of `AnimatePresence` with a unique `key` should be given a unique `ref`).
+
 ## [1.6.15] 2019-10-24
 
 ### Added
--- a/dev/examples/positionTransition.tsx
+++ b/dev/examples/positionTransition.tsx
@@ -1,0 +1,88 @@
+import * as React from "react"
+import { useState, useEffect } from "react"
+import { motion } from "@framer"
+import { clamp } from "@popmotion/popcorn"
+
+export const App = () => {
+    const [colors, setColors] = useState(initialColors)
+    const [widths, setWidths] = useState(initialWidths)
+
+    useEffect(() => {
+        setTimeout(() => {
+            setWidths({
+                "#FF008C": 200,
+                "#D309E1": 300,
+                "#9C1AFF": 100,
+                "#7700FF": 150,
+            })
+        }, 2000)
+    }, [])
+
+    return (
+        <ul className="items">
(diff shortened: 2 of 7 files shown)


souporserious:
Thanks for getting this fix out! ❤️

Regarding the issue you were originally trying to fix, would exposing a special version of `useRef` help here? This way Framer would know if it's a special use case and could throw a warning when trying to access a ref in `AnimatePresence` and point to the proper usage. Not sure the implications, but maybe it could also do something similar to what we discussed in #360 with providing dimensions collected from Framer to mitigate multiple reads.

PR #395, opened 2019-11-19
Fix mutating handler refs in render

Andarist:
Followup to the discussion I've had with @InventingWithMonster . 

Mutating refs in a render is still a side effect - even in disguise. I think this version is async-react safe and also easier to reason about (in terms of async react), because which copy of those handlers gets actually executed is known right now (it's the one that was there during last commit phase of React), whereas previously it could be any copy from any React render (in between 2 commits of course) and there might be many of them.


jest.config.json
```diff
@@ -9,5 +9,9 @@
     "coverageDirectory": "<rootDir>/../coverage",
     "setupFilesAfterEnv": ["<rootDir>/../jest.setup.ts"],
     "testMatch": ["**/__tests__/**/*.test.(js|ts)?(x)"],
-    "testPathIgnorePatterns": ["ssr.test.tsx"]
+    "testPathIgnorePatterns": ["ssr.test.tsx"],
+    "watchPlugins": [
```
Andarist: added bonus of this PR - it's a super useful jest plugin, which once was in the core, no idea why they have removed it, see a gif-demo here: https://github.com/jest-community/jest-watch-typeahead

If you want I can remove this or move to another PR

package.json
```diff
@@ -62,13 +62,14 @@
         "jest": "^24.8.0",
         "jest-dom": "^3.5.0",
         "jest-junit": "^6.4.0",
+        "jest-watch-typeahead": "^0.4.2",
         "lint-staged": "^8.0.4",
         "lodash": "^4.17.13",
         "prettier": "^1.15.2",
         "progress-bar-webpack-plugin": "^1.11.0",
-        "react": "16.8.6",
+        "react": "^16.12.0",
```
Andarist: I've needed async act to fix tests and this has been introduced in 16.9, so had to update those.

src/behaviours/__tests__/utils.tsx
```diff
@@ -33,6 +35,31 @@ export const drag = (element: any) => {
     return controls
 }
 
+export const asyncDrag = (element: any) => {
```
Andarist: this should become just a `drag` but that will require updating all the tests using `drag` at the moment, so I didn't want to include such changes as part of this PR - if you want I can work on migrating other tests to this one in a followup PR

src/behaviours/__tests__/utils.tsx
```diff
@@ -33,6 +35,31 @@ export const drag = (element: any) => {
...
+    fireEvent.mouseDown(element)
+
+    const controls = {
+        to: async (x: number, y: number) => {
+            pos.x = x
+            pos.y = y
+
+            await act(async () => {
+                fireEvent.mouseMove(document.body, { buttons: 1 })
+                await new Promise(resolve => sync.update(resolve))
```
Andarist: kinda leverages internal knowledge about `usePanGesture`, but this is the only way to fix act warnings

src/gestures/__tests__/pan.test.tsx
```diff
@@ -30,16 +30,16 @@ describe("pan", () => {
...
+            sync.postRender(async () => {
+                await pointer.to(50, 50)
                 sync.postRender(() => {
                     pointer.end()
                 })
             })
         })
 
-        return expect(promise).resolves.toBe(7)
+        return expect(promise).resolves.toBe(10)
```
Andarist: As you might notice the expected value has changed, I've analyzed the situation - and this is now correct. This test is really time-dependent and an expected result, such as this, is really sensitive to changes. Ideally this test should be rewritten somehow, so it wouldn't care about the timing that much (it's hard to tell if it tests the correct thing right now), but I couldn't figure out quickly how this thing should be properly tested.

InventingWithMonster:
Excellent PR thanks for this. I'm going to take a pass at strict mode and concurrent mode next week too so this is a great start.

Andarist:
@InventingWithMonster let me know if you want any help regarding this, will you still want to merge this PR before next week?

src/gestures/__tests__/pan.test.tsx
Andarist: after some thinking - we could also change the expected result here to an array of strings which would contain a mutable "render" counter in it (it would indicate that handlers are indeed coming from a particular render, reach to its closure). Right now with a plain counter, it's hard to determine if they really have been incremented by "correct" handlers - especially given the fact that `onDrag` is being called each frame

InventingWithMonster:
I'm probably going to merge it after lunch yeah

InventingWithMonster:
If it's ready?

Andarist:
Tests are passing 😉 

One thing that could be addressed is this - https://github.com/framer/motion/pull/395#discussion_r348435406 but it can also be addressed as followup PR because it's not a newly introduced test.

PR #397, opened 2019-11-22
Migrate tests to act-wrapped drag helper

Andarist:
It's a followup PR to https://github.com/framer/motion/pull/395 . In that PR I've introduced `asyncDrag` helper to limit its scope, but as it got merged I've decided to migrate all the other tests to use it, as there should only be one way to test drag interactions.

This allowed me to remove the old implementation of this helpers as `asyncDrag` became just `drag`. In addition to that those tests have now more flat structure, leveraging async/await - most of the time there is no need to wrap stuff in `postRender`, nor there is any need to create Promise-wrappers for those tests.


src/behaviours/__tests__/index.test.tsx
```diff
@@ -754,168 +657,155 @@ describe("dragging", () => {
...
+        await pointer.to(-500, 0)
+        pointer.end()
+        await sleep(50)
 
-        return expect(promise).resolves.toBeCloseTo(-500)
+        // if we set really strong spring and animation has ended - this has worked
+        expect(x.get()).toBeCloseTo(-500)
     })
 
     test("pointer down kills momentum", async () => {
```
Andarist: Please review change to this test extra carefully - check if this still tests the correct thing

src/gestures/__tests__/pan.test.tsx
```diff
@@ -1,45 +1,46 @@
...
+        }
 
-            const { container, rerender } = render(<Component />)
-            rerender(<Component />)
+        const { container, rerender } = render(<Component />)
+        rerender(<Component />)
 
-            const pointer = await asyncDrag(container.firstChild).to(100, 100)
+        const pointer = await drag(container.firstChild).to(100, 100)
+        await frame.postRender()
```
Andarist: Ideally this shouldn't be needed. I've only preserved explicit calls to `sync.postRender` in those tests concerned about calling latest version of handlers, because I had to match expected counter. I believe those 2 tests could be rewritten somehow in such a way that we wouldn't have to explicitly call this API.

InventingWithMonster:
The await postRender is really nice

Andarist:
@InventingWithMonster fixed tests, should be ready to merge now

Andarist:
@InventingWithMonster friendly 🏓

InventingWithMonster:
Sorry for the delay, was off for most of December. Thanks for this PR!

Andarist:
No problem Matt :) Just wanted to avoid potential conflicts, dont have much time for all projects I'd like to contribute lately 😢 so trying to optimize the time i have 😉

Issue #401, opened 2019-12-02
[FEATURE] Add support for `prefers-reduced-motion`


green-arrow:
**Is your feature request related to a problem? Please describe.**

This is an issue from the `pose` library that I don't think ever made it in, but I would love to see it in framer motion: https://github.com/Popmotion/popmotion/pull/452

**Describe the solution you'd like**

Ideally, there would be an option to globally disable all animations, which would solve `prefers-reduce-motion` as well as disabling animations during tests, which is currently causing issues for me as well.

green-arrow:
Additionally, I'd be happy to help implement / make a PR if someone can guide me in the right direction 😄

soiluong:
The solution I had involved the “animate” attribute and to aggregate the animation objects to a global place or a single file. That way I could introduce a “simpler” or reduced animation option and allow the user to toggle between them.

InventingWithMonster:
The logic does exist https://github.com/framer/motion/blob/master/src/dom/accessibility.ts

Before it's wired up there's some questions that need answering:

- What do we do in development mode for people who are using prefersReducedMotion?
- Is there an API to force motion on for all users?
- Do we only disable certain animations such as x/y/scale/rotate while converting into opacity animations like iOS?

PR #407, opened 2019-12-10
Add support for `prefers-reduced-motion`

InventingWithMonster:
This PR adds a new `useReducedMotion` hook to allow developers to create accessible animations based on the device's Reduced Motion setting.

`useReducedMotion` returns a boolean that indicates the state of this setting.

```jsx
function BackgroundVideo() {
  const isReducedMotion = useReducedMotion()
  return <video autoplay={!isReducedMotion} />
}
```

Consuming components will re-render as this setting is enabled and disabled.

I attempted an auto-magic solution but there were too many caveats, I'd first rather a solution that can be used for every use-case.

### Disabling parallax

```jsx
function ParallaxItem({ parentScrollY }) {
  const isReducedMotion = useReducedMotion()
  const y = useTransform(parentScrollY, [0, 1], [0, 0.5])

  return <motion.div style={!isReducedMotion && { y }} />
}
```

### Replacing position animation with opacity

```jsx
function Component({ isOpen }) {
  const isReducedMotion = useReducedMotion()
  let animate

  if (isOpen) {
    animate = isReducedMotion ? { opacity: 1 } : { x: 0 }
  } else {
    animate = isReducedMotion
      ? { opacity: 0 }
      : { x: "-100%" }
  }

  return <motion.div animate={animate} />
}
```
Closes  https://github.com/framer/motion/issues/401
Fixes https://github.com/Popmotion/popmotion/issues/677
Closes https://github.com/Popmotion/popmotion/pull/452


nvh:
Nice! One thing that is unclear to me is if `useReducedMotion` will listen to the options specified by the `ReducedMotion` component. And is there a way to override it at the hook call site? Maybe the force and detect options can be passed into it?

InventingWithMonster:
Yes sorry the two are connected. `ReducedMotion` sets options for a tree and `useReducedMotion` can be used to read whether we should use reduced motion based on

1. If we're being forced to
or
2. If we're a) detecting `prefers-reduced-motion` and b) this is `true`

By default we're not being forced into reduced motion and we are detecting the device setting.

And yeah I can see the value of overriding these in the hook itself for sure.

dev/examples/reduceMotion.tsx
```diff
@@ -0,0 +1,29 @@
...
+const style = {
+    width: 100,
+    height: 100,
+    background: "red",
+    opacity: 1,
+}
+
+export const App = () => {
+    const [isVisible, setIsVisible] = React.useState(false)
+    const isReducedMotion = useReducedMotion()
```
aron: Suggestion: use `shouldReduceMotion` as the result here as it reads better throughout the code an implies it's on the caller to do something with the value. `isReducedMotion` sounds like you're checking if motion has been suppressed somehow by the library.

src/dom/accessibility.ts
```diff
@@ -1,23 +1,40 @@
...
+     * Control whether to detect the device's Reduce Motion setting.
+     * @internal
+     */
+    detect: boolean
+
+    /**
+     * Whether to force reduce motion mode.
+     * @internal
+     */
+    force: boolean
```
aron: Is there a benefit to having these two decoupled? What happens if `force` is unset and `detect` is false? Do we fallback to a default value? 

It feels like passing force should just override the detection? What's the case where you'd want to disable detection but not force a setting?

src/utils/use-reduced-motion.ts
```diff
@@ -0,0 +1,44 @@
...
+ * `x`/`y` animations with `opacity, disabling the autoplay of background videos, or turning off parallax motion.
+ *
+ * It will actively respond to changes and re-render your components with the latest setting.
+ *
+ * ```jsx
+ * export function MyComponent() {
+ *   const isReducedMotion = useReducedMotion()
+ *
+ *   return <video autoplay={!isReducedMotion} />
+ * }
```
aron: I'd suggest going with an example that uses motion here such as replacing the x/y animation with opacity.

dev/examples/reduceMotion.tsx
```diff
@@ -0,0 +1,29 @@
...
+    const transition = isReducedMotion ? { type: false } : { duration: 1 }
+    const variants = {
+        visible: { opacity: 1, transition },
+        hidden: { opacity: 0, transition },
+    }
+
+    React.useEffect(() => {
+        setTimeout(() => setIsVisible(!isVisible), 1500)
+    })
```
aron: Want to pass an empty array here? Otherwise this timeout will be set on each render, unclear if that's intended or not.

InventingWithMonster: yeah it's intended

InventingWithMonster: could pass isVisible

src/dom/accessibility.ts
InventingWithMonster: It's a good point. It is currently:

```
force={true} // Force reduce motion
force={false} // (default) Do not force reduce motion (detect)
```

It could be:

```
force={true} // Force reduce motion
force={false} // Force reduce motion off
force={null} // (default) Do not force reduce motion (detect)
```

aron: Yeah, I think that makes sense. I take it we need to specify that `null` or `undefined` will revert back to detection for use in a nested context?

InventingWithMonster: Do you think rather than detect and force we have a prop more explicit ie:

```
<ReduceMotion isReducedMotion={}>
```

InventingWithMonster: Or `enabled`

aron: I like `enabled`.

src/components/ReducedMotion.tsx
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Define accessibility options for a tree. Can be used to force the tree into Reduced Motion mode,
+ * or disable device detection.
+ *
+ * @internal
+ */
+export function ReducedMotion({ children, enabled }: Props) {
+    let context = useContext(MotionContext)
+
+    context = {
```
Andarist: maybe worth memoizing this? at the moment each ReducedMotion rerender will scan the whole subtree to update context value, most likely this ReducedMotion should always be put on the top level of the app (so it actually wouldnt ever - or really rarely - be rerendered), but it's not a given

src/motion/context/MotionContext.ts
```diff
@@ -121,6 +122,7 @@ export const useMotionContext = (
             animate: targetAnimate,
             values,
             hasMounted,
+            isReducedMotion: parentContext.isReducedMotion,
```
Andarist: will this propagate `isReducedMotion` change correctly if it changes? 🤔 it's not used as a dependency to this memo

src/utils/use-reduced-motion.ts
```diff
@@ -0,0 +1,53 @@
...
+ */
+export function useReducedMotion() {
+    const { isReducedMotion } = useContext(MotionContext)
+    const [shouldReduceMotion, setShouldReduceMotion] = useState(
+        determineShouldReduceMotion(prefersReducedMotion.get(), isReducedMotion)
+    )
+
+    useEffect(
+        () => {
+            return prefersReducedMotion.onChange(() => {
```
Andarist: I'm not too familiar with motionValue API and I'm removing this from GitHub so I can't quite check right now type signature, but seems a little bit weird to call `prefersReducedMotion.get()` here, isnt current value provided to the onChange callback?

src/components/ReducedMotion.tsx
InventingWithMonster: Yeah great point thanks

src/utils/use-reduced-motion.ts
InventingWithMonster: yeah i was just keeping it the same as above but this is a good point

src/dom/accessibility.ts
```diff
@@ -1,23 +1,28 @@
...
- * @internal
- */
-export const isReducedMotion = () => prefersReducedMotion
+export function determineShouldReduceMotion(
+    prefersReduced: boolean | null,
+    isReducedMotion: boolean | undefined
+): boolean {
+    return isReducedMotion !== undefined
+        ? isReducedMotion
+        : Boolean(prefersReduced)
```
aron: A nit here, but it feels like this would be clearer written as `typeof isReducedMotion === "boolean" ?`.

PR #409, opened 2019-12-11
Remove excessive direct usage of act in tests

Andarist:
`render` from `@testing-library/react` is already act-wrapped, so no need to wrap it again. Additionally, I've created act-wrapped version of some fireEvents in jest.setup.


InventingWithMonster:
Ah awesome to know - thanks for this!

PR #420, opened 2019-12-19
Re-add 'rollup-plugin-uglify' to patch a security vulnerability in 'seralize-javascript'

koenrh:
Re-add 'rollup-plugin-uglify' to trigger an update of transitive dependency 'serialize-javascript' in order to patch a security vulnerability.

https://snyk.io/vuln/SNYK-JS-SERIALIZEJAVASCRIPT-536840


TrySound:
I recommend to migrate to terser plugin instead. It's better supported and has more trust.

PR #426, opened 2020-01-04
Add TargetResolver to type of exit prop

### LINKED ISSUE #423: [BUG] Type definition for `exit` prop doesn't support functions
**Describe the bug**

Passing a function into the `exit` prop doesn't appear to be supported. The reproduction below is using `AnimatePresence` with a `custom` prop, and providing a function to the `exit` prop. This works, but there is a type error.

**IMPORTANT: Provide a CodeSandbox reproduction of the bug**

https://codesandbox.io/s/wispy-bush-y9ims?fontsize=14&hidenavigation=1&theme=dark

**Expected behavior**

I believe I read this in the documentation (and it works), so the expected behavior would be that there are no type errors.

airjp73:
This fixes #423.

I mentioned in the issue that this behavior was in the documentation, but now I can't find it. Is it intended to work this way?

If it is, should I add this type to the `initial` and `animate` props too?

### DIFF
--- a/src/motion/types.ts
+++ b/src/motion/types.ts
@@ -6,6 +6,7 @@ import {
     Target,
     Transition,
     TargetAndTransition,
+    TargetResolver,
     Omit,
     MakeCustomValueType,
 } from "../types"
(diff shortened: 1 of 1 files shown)


InventingWithMonster:
Thanks for the PR!

It wasn't intended to work this way because these props as objects would only be resolved with data from the current render anyway. So even though it works I'd rather leave this out of `initial` and `animate` to keep the API slimmer. However because of the unique way `exit` works, on components that aren't actually there any more, I totally see a use-case for this kind of pattern.

PR #482, opened 2020-03-16
Added a null check in getTranslateFromMatrix() method

JoyalJoyMadeckal:
We are doing integration testing with react and we were facing issues exception from framer motion module. Added Null check which is working fine for us.


JoyalJoyMadeckal:
Updated code and fixed test failures

JoyalJoyMadeckal:
Can I expect an update on this?
We are actually facing a problem in our tests as matrix is coming as null in our tests.

JoyalJoyMadeckal:
Thanks

PR #483, opened 2020-03-17
Feature/magic motion border radius

hemlok:
Allows magic-motion animations for individual border-radius properties.

### DIFF
--- a/src/motion/magic/Magic.tsx
+++ b/src/motion/magic/Magic.tsx
@@ -189,7 +189,17 @@ export class Magic extends React.Component<FunctionalProps> {
         const prevStyle = this.prev.style
         const nextStyle = this.next.style
         const isAnimatingRotate = prevStyle.rotate !== nextStyle.rotate
-        const hasBorderRadius = prevStyle.borderRadius || nextStyle.borderRadius
+
+        const hasBorderTopLeftRadius =
+            prevStyle.borderTopLeftRadius || nextStyle.borderTopLeftRadius
+        const hasBorderTopRightRadius =
+            prevStyle.borderTopRightRadius || nextStyle.borderTopRightRadius
+        const hasBorderBottomLeftRadius =
+            prevStyle.borderBottomLeftRadius || nextStyle.borderBottomLeftRadius
+        const hasBorderBottomRightRadius =
+            prevStyle.borderBottomRightRadius ||
+            nextStyle.borderBottomRightRadius
+
         const hasBoxShadow =
             prevStyle.boxShadow !== "none" || nextStyle.boxShadow !== "none"
         const updateBoxShadow =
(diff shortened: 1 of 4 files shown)


src/motion/magic/utils.ts
```diff
@@ -36,7 +46,10 @@ export function snapshot(element: NativeElement): Snapshot {
         style: {
             backgroundColor,
             border,
-            borderRadius: borderRadius ? parseFloat(borderRadius) : 0,
+            borderTopLeftRadius: borderRadiusMatrix[0],
+            borderTopRightRadius: borderRadiusMatrix[1],
+            borderBottomLeftRadius: borderRadiusMatrix[2],
+            borderBottomRightRadius: borderRadiusMatrix[3],
```
hemlok: We could also combine this into a single array of border radii `[tl, tr, bl, br]` 🤔

src/motion/magic/Magic.tsx
```diff
@@ -207,7 +217,13 @@ export class Magic extends React.Component<FunctionalProps> {
...
         const scaleY = values.get("scaleY", 1)
         const rotate = values.get("rotate", 0)
-        const borderRadius = values.get("borderRadius", "")
+        const borderTopLeftRadius = values.get("borderTopLeftRadius", "")
+        const borderTopRightRadius = values.get("borderTopRightRadius", "")
+        const borderBottomLeftRadius = values.get("borderBottomLeftRadius", "")
+        const borderBottomRightRadius = values.get(
+            "borderBottomRightRadius",
+            ""
+        )
```
hemlok: Similarly to how we could combine the `style` values for individual border radii into a single prop, we could take a similar approach here, and get and set a single value for border-radius with all 4 values to simplify this down to a single get/set 🤔 

I'm not sure it would necessarily simplify the code over all, but it would make it more compact by each property type being handled in a singular way.

InventingWithMonster: I was under the understanding that `border` can only set each corner to a single radius if its setting four values, whereas to correct for each axis we need to set each corner to two values. I think its fine like this for now though.

src/motion/magic/Magic.tsx
```diff
@@ -337,18 +362,29 @@ export class Magic extends React.Component<FunctionalProps> {
         rotate.set(target)
     }
 
-    updateBorderRadius(p: number, borderRadius: MotionValue<string>) {
+    updateBorderRadius(
+        p: number,
+        borderRadius: MotionValue<string>,
+        index: number
+    ) {
+        const indexMap = [
```
InventingWithMonster: As `updateBorderRadius` runs four times per frame per component (actually currently potentially eight times as it might run again to account for parent distortion) I try and avoid creating data structures and functions in any "per frame" code. This is why the `boxShadow` updater is the way it is (in `createUpdateBoxShadow`) - its creating all the expensive stuff up front.

Also I don't think the array approach leads to any space savings so you could probably do without the array entirely, you could just make it more explicit and save the lookups with `updateBorderRadius(p, borderTopLeftRadius, "borderTopLeftRadius")`

src/motion/magic/Magic.tsx
hemlok: Ahh you might be right. Great 👍

src/motion/magic/Magic.tsx
hemlok: Good to know, great 👍

PR #495, opened 2020-03-30
Replacing the drag pointerdown `e.preventDefault()` with CSS and HTML attributes

InventingWithMonster:
Calling `e.preventDefault()` from `pointerdown` disables mouse tracking outside of iframes.

### DIFF
--- a/src/motion/functionality/dom.tsx
+++ b/src/motion/functionality/dom.tsx
@@ -58,6 +58,28 @@ function filterValidProps(props: MotionProps) {
     return domProps
 }
 
+const buildHTMLProps = (
+    values: MotionValuesMap,
+    style: CSSProperties,
+    isStatic?: boolean,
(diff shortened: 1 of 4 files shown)


src/motion/functionality/dom.tsx
```diff
@@ -58,6 +58,28 @@ function filterValidProps(props: MotionProps) {
...
+    const props: any = {
+        style: buildStyleAttr(values, style, isStatic),
+    }
+
+    if (isDrag) {
+        // Disable text selection
+        props.style.userSelect = "none"
+
+        // Disable the ghost element when a user drags
+        props.draggable = false
```
jonastreub: Why do we override draggable?

InventingWithMonster: We probably should have been doing this anyway but if you don't do this you get a weird ghost element

Issue #254, opened 2019-07-30
[FEATURE] Reducing bundle size


mattgperry:
Currently Framer Motion is around 24kb gzipped and minified as a whole. With tree-shaking the `motion` component alone is about 22kb.

This isn't unprecedented, with Pose being around 25kb and GSAP being in the low 30s. But a low bundle size is quite (either the American or British "quite") important to me, and I've had this thought in my head for a while that it's daft that the animation part of an animation library has to load synchronously.

I've done some experiments and I think it'd be possible to crank the core library down to 10kb. The rest could either be loaded synchronously with the main bundle, but as optional bits (for instance someone who only wants to use animations). Or using `import()` the rest could be loaded in asynchronously.

For ease of use I'd like the main entry point to remain the same, with something like `motion/lazy` as a safe second entry point for people who want to crank down the bundle size.

### Functionality: ~5kb

Functionality is loaded in via renderless components so we don't run, for instance, draggable code on components that aren't draggable. 

These are already abstracted and loaded into a [single function](https://github.com/framer/motion/blob/master/src/motion/functionality/dom.tsx#L84). It could be that we allow this to be configured and loaded in via the currently private `MotionPluginContext` API. So only the functionality used in the site is loaded at all.

A further step could be that users could dynamically import this configuration, re-rendering `MotionPluginContext` and its subscribers when functionality has downloaded. This could be used to defer the loading of dynamic functionality until after the initial render has finished.

### Popmotion animations: ~4.5kb

At this point all animations have been abstracted behind a [single function that returns a promise](https://github.com/framer/motion/blob/master/src/animation/ValueAnimationControls.ts#L527). As all composition also works via Promises it might work quite neatly that calls to `startAnimation` occuring before Popmotion has loaded in simply wait for it to load before running.

### Element names: ~0.7kb

Currently `motion` components are created by iterating over [two lists of tag names](https://github.com/framer/motion/blob/master/src/motion/utils/supported-elements.ts). Pose [used to use a `Proxy` to do this](https://github.com/Popmotion/popmotion/blob/2399722bb40eb2671be8f99c21a85213a439fe8b/packages/react-pose/src/posed.tsx#L58) without maintaining lists but I had to replace it because Googlebot would crash on Pose-powered pages. `Proxy` is now supported by Googlebot, which means we can support it too.

This would also theoretically reduce memory usage and start times, and allow us to support arbitrary web components out of the box.

### Value conversion: ~2kb

[...]

PR #539, opened 2020-05-11
WIP: Size reduction

InventingWithMonster:
Fixes https://github.com/framer/motion/issues/254

This PR attempts to enable the reduction of the Framer Motion bundle size while not compromising on the simple API.

The outcome will be an unchanged API and bundle size for normal usage, with the option to import a critical-path only version of the `motion` component using something like:

```jsx
import { m as motion } from 'framer-motion'
```

By default, when this component is used, only `initial`, and `style` will be resolved:

```jsx
<motion.div initial={{ opacity: 0 }} style={{ x }} />
```

Supporting this for HTML and SVG will reduce the `m` bundle size down to about 4kb. It might be that we also make SVG support optional and reduce this by another 0.5kb.

From there, a user will be able to add only the features they use via a `MotionConfig` component.

```jsx
import { drag, variants } from 'framer-motion'

// later
export const App = () => {
  return (
    <MotionConfig features={{ drag, variants }}>
      {children}
    </MotionConfig>
  )
}
```

Each added feature will increase the bundle size accordingly.

Depending on a user's needs, some features will be identified as less crucial than others. These can be split out into a separate module and imported dynamically:

```jsx
// features.js
import { drag } from 'framer-motion'
export default { drag }

// index.js
import { useFeatures, variants } from 'framer-motion'

export const App = () => {
  const features = useFeatures(() => import('./features'))
  
  return (
    <MotionConfig features={{ ...features, variants }}>
      {children}
    </MotionConfig>
  )
}
```

This will reduce initial bundle size accordingly.

`MotionConfig` will also allow portions of the tree to be reconfigured with different renderers, for instance a Framer Motion that works with `react-three-fibre`.


otorrillas:
I have a few of questions about the usage and changes involved:

## Migration strategy

If I understand this correctly:
> The outcome will be an unchanged API and bundle size for normal usage, with the option to import a critical-path only version of the `motion` component using something like:
> 
> ```js-jsx
> import { m as motion } from 'framer-motion'
> ```

For getting the benefits of bundle reduction, we would need to change the default export for a named `m` export. So, from:

```js-jsx
import motion from 'framer-motion'
```
To:
```js-jsx
import { m as motion } from 'framer-motion'
```

Since the change is "straightforward", do you think that providing a [`codemod`](https://github.com/facebook/jscodeshift) could be useful for users with migrating to the new API? I'm more than happy to help with that.

## MotionConfig API

### useFeatures

What's the difference between:

> ```js-jsx
> import { drag, variants } from 'framer-motion'
> 
> // later
> export const App = () => {
>   return (
>     <MotionConfig features={{ drag, variants }}>
>       {children}
>     </MotionConfig>
>   )
> }
> ```

And:

> ```js-jsx
> // features.js
> import { drag } from 'framer-motion'
> export default { drag }
> 
> // index.js
> import { useFeatures, variants } from 'framer-motion'
> 
> export const App = () => {
>   const features = useFeatures(() => import('./features'))
>   
>   return (
>     <MotionConfig features={{ ...features, variants }}>
>       {children}
>     </MotionConfig>
>   )
> }
> ```

Is it that imports in `feature.js` file will be lazy loaded? So they do not influence the initial bundle size? What happens if `motion` library tries to use a feature that is not loaded yet?

### Another approach for lazy loading

Assuming `MotionConfig` is a context provider... what is the expected behavior if we nest `MotionConfig`s? So, we start with "minimal" configuration, but some features are only if their components are loaded? 

Example:
I have an upload section in my app, let's say `UploadPage`, that I lazy load only if the user chooses to upload. 

Something like:
```js-jsx
// app.js
import { variants } from 'framer-motion'

export const App = () => (
  <MotionConfig features={{ variants }}>
        {children}
  </MotionConfig>
)

// upload.js - deeper in the tree
import { drag } from 'framer-motion'

export const Upload = () => (
  <MotionConfig features={{ drag }}>
        {children} // here 
  </MotionConfig>
)
```
Thus, in `Upload` we "extend" the parent `MotionConfig` so we can add more "features" only in trees we need them. But I'm not sure if this over-complicates things.



@InventingWithMonster this is my first time reviewing something in `motion` repository, but I hope you find the feedback useful. Thanks for all the work, `motion` is awesome! 😄

InventingWithMonster:
> Since the change is "straightforward", do you think that providing a [`codemod`](https://github.com/facebook/jscodeshift) could be useful for users with migrating to the new API? I'm more than happy to help with that.

Yeah, I do think it would be that straightforward so a codemod would be helpful here.

> Is it that imports in feature.js file will be lazy loaded? So they do not influence the initial bundle size?

Correct

> What happens if motion library tries to use a feature that is not loaded yet?

Nothing will break. The functionality is already loaded in via renderless components so these will just mount when the functionality loads, and all lifecycle methods will fire as you'd expect (ie mount animations will run). In development, we could generate a one-time warning to remind people to ensure these components are loaded.

> Thus, in Upload we "extend" the parent MotionConfig so we can add more "features" only in trees we need them. But I'm not sure if this over-complicates things.

Yeah I expect `MotionConfig` to be nestable. This will also be how we load in alternative, scoped renderers.

ShanonJackson:
I'd absolutely love to be of help here i'm experienced in making bundles smaller/faster; Iv'e reached out to you on discord Matt for an initial pull request (as you know) however after skimming through why the bundle is so big i think it wouldn't be too much work to slice a large portion of the bundle (however iv'e only skimmed over why the size is what it is).

Some examples of this is in the bundle the entire popmotion library is being included even though you are only using a few exports; This library is required into "motion".

I'd love to help, i can basically rinse/repeat what i'm trying to do in framer-motion in that fork in popmotion and will have same effect, unused exports will be wiped clean.

InventingWithMonster:
With ES6 everything that can be code split already is. Popmotion is
included with everything because it’s used by every animating module
(animation, variants, auto animate and drag). It’ll take this
rearchitecture to allow people to drop the modules they don’t import and in
turn Popmotion will only be included when one of them needs it. Only the
bits and pieces of Popmotion that are used by the respective modules are
imported - it’s full bundle size is 13kb but again that isn’t represented
here.

InventingWithMonster:
Closing this as most of the architectural moves have already been landed in `main`.

Issue #572, opened 2020-05-24
[FEATURE] capture dependency tree


sioanis:
Add `package-lock.json` as a mechanism to capture the exact dependency tree installed at any point in time.

This helps with collaboration across different environments in which you want everyone fetching dependencies for a specific version of your project to fetch the same tree.

Enables the ability to use [npm ci](https://blog.npmjs.org/post/171556855892/introducing-npm-ci-for-faster-more-reliable) for faster, more reliable builds.

sioanis:
Opened a PR here: https://github.com/framer/motion/pull/573

sioanis:
`yarn.lock` was added :)

PR #573, opened 2020-05-24
Capture dependency tree

sioanis:
Add `package-lock.json` as a mechanism to capture the exact dependency tree installed at any point in time.

This helps with collaboration across different environments in which you want everyone fetching dependencies for a specific version of your project to fetch the same tree.

Enables the ability to use [npm ci](https://blog.npmjs.org/post/171556855892/introducing-npm-ci-for-faster-more-reliable) for faster, more reliable builds.

Issue: #572

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,13880 @@
+{
+    "name": "framer-motion",
+    "version": "1.11.0",
+    "lockfileVersion": 1,
+    "requires": true,
+    "dependencies": {
+        "@babel/code-frame": {
+            "version": "7.8.3",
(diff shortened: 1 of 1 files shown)


InventingWithMonster:
Thanks for the PR! How was this file generated and how does it get updated? I'd expect a config change or similar.

adibfirman:
Actually that file generated if we first init a package either with yarn or npm it will be generated automatically a `*.lock` file and then that file will be updated if we trying to upgrade a package in our 
`packages.json` file

> Thanks for the PR! How was this file generated and how does it get updated? I'd expect a config change or similar.

sioanis:
@InventingWithMonster Ah, it looks like you added a `yarn.lock`. Awesome, I'll close this. Thanks!

PR #591, opened 2020-06-08
VisualElement refactor

InventingWithMonster:
This PR removes the `stylefire` dependency and introduces the `VisualElement` class. A `VisualElement` is a generic adapter between Framer Motion and different renderers. Currently we only have HTML and SVG renderers but this will be expanded to others. By dealing with VisualElement the hope is that we can extend functionality to other renders simply by writing new VisualElements. 

This PR isn't fully generic but provides enough of a refactor that we'll better identify where further progress can be made when we make another renderer.

By removing `stylefire` as a dependency and moving the direct renderer to within Framer Motion, we're also making it possible to refactor the layout animation API to be additive. This means we can first animate and/or correct the layout, then additionally animate the transform if this has been defined by the user.

It also reduces generated bundle size by ~1.5kb


dev/examples/visualElement.tsx
```diff
@@ -0,0 +1,26 @@
...
+        ? { x, y: 10, backgroundColor: "red" }
+        : { x: -100, y: 50, backgroundColor: "green" }
+
+    return (
+        <motion.div
+            onClick={() => setIs(!is)}
+            style={{ ...style, ...motionStyle }}
+        />
+    )
+}
```
hemlok, line 26: Should this example animate?

src/animation/use-animated-state.ts
```diff
@@ -1,67 +1,56 @@
...
+    }
+
+    readNativeValue(key: string) {
+        return this.latestState[key] || 0
+    }
+
+    render() {
+        console.log("render", this.latest.foo)
+    }
+}
```
hemlok, line 25: Is this WIP code? Can it be committed safely?

src/components/AnimatePresence/use-presence.ts
```diff
@@ -34,11 +34,20 @@ type NotPresent = [false, SafeToRemove]
...
     return !isPresent && onExitComplete ? [false, onExitComplete] : [true]
 }
+
+/**
+ * @public
+ */
+export function useIsPresent() {
+    const context = useContext(PresenceContext)
+    return context === null ? true : context.isPresent
+}
```
hemlok, line 53: 👍

src/motion/__tests__/variant.test.tsx
```diff
@@ -39,474 +39,474 @@ describe("animate prop as variant", () => {
...
-        }
-
-        const { rerender } = render(<Component length={1} />)
-        rerender(<Component length={1} />)
-        rerender(<Component length={2} />)
-        rerender(<Component length={2} />)
-
-        expect(x.get()).toBe(100)
-    })
+    // test("fires onAnimationStart when animation begins", async () => {
```
hemlok, line 42: 😂 What's the deal with these tests? Not valuable anymore? Or just not compatible with Visual Element yet and need to be updated down the road?

dev/examples/visualElement.tsx
InventingWithMonster: No, actually I'll delete it as I was just ensuring styles get applied/removed

src/animation/use-animated-state.ts
InventingWithMonster: Ah yeah good catch on the render. This is basically just some hack to reimplement useAnimatedState using the new internal APIs. This isn't a public API, but some Framer users were told about it and now use it in their projects.

src/motion/__tests__/variant.test.tsx
InventingWithMonster: Ha no I forgot to uncomment them. I was testing just the one test at the time.

src/animation/use-animated-state.ts
hemlok: Cool 👍

src/render/dom/HTMLVisualElement.ts
```diff
@@ -0,0 +1,193 @@
...
+
+    /**
+     * Render the Element by rebuilding and applying the latest styles and vars.
+     */
+    render() {
+        // Rebuild the latest animated values into style and vars caches.
+        this.build()
+
+        // Directly assign style into the Element's style prop. In tests Object.assign is the
+        // fastest way to assign styles.
```
hemlok, line 180: Good to know 👀

hemlok:
Code looks good 🧙  Going to run through the examples now.

hemlok:
Just ran through all the examples. Most everything is unchanged from before 🎉 

There are 3 examples that seem to be not functioning the same as on `feature/magic-move`: `animateSVGPath`, `heightAutoFromDisplayNone`, `viewportScroll`.

PR #592, opened 2020-06-09
Replace useLayoutEffect with useIsomorphicEffect

### LINKED ISSUE #578: [BUG] useLayoutEffect warning when rendering serverside
The change in #195 introduces `useLayoutEffect` to the the `useViewportScroll` (and the new `useElementScroll`) hook, that causes a React warning when rendering serverside.

See [CodeSandbox](https://codesandbox.io/s/small-http-u4fnj?file=/pages/index.js)

```
Warning: useLayoutEffect does nothing on the server, because its effect cannot be encoded into the server renderer's output format. This will lead to a mismatch between the initial, non-hydrated UI and the intended UI. To avoid this, useLayoutEffect should only be used in components that render exclusively on the client. See https://fb.me/react-uselayouteffect-ssr for common fixes.
```

A common fix is to create a `useIsomorphicLayoutEffect` that selects the right effect for the env.

```js
const useIsomorphicLayoutEffect = isBrowser ? useLayoutEffect : useEffect

function Comp() {
    // ...
  })
}
```

thebuilder:
This fixes #578 by adding a `useIsomorphicEffect` to use instead of `useLayoutEffect`. 
Stops React throwing a warning when rendering serverside.

### DIFF
--- a/src/utils/use-isomorphic-effect.ts
+++ b/src/utils/use-isomorphic-effect.ts
@@ -1,0 +1,4 @@
+import { useEffect, useLayoutEffect } from "react"
+const isBrowser = typeof window !== "undefined"
+
+export const useIsomorphicLayoutEffect = isBrowser ? useLayoutEffect : useEffect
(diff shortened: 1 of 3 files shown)


InventingWithMonster:
Great PR, thanks for this. I'll have it published sometime this morning

InventingWithMonster:
Published in 1.11.1

PR #593, opened 2020-06-09
Change prop "isOpen" to "isVisible" in the example of "positionTransition"

SilencerWeb:
I guess somebody forgot to change "isOpen" to "isVisible" in these 2 places

image


InventingWithMonster:
Thanks!

PR #596, opened 2020-06-10
chore: improve builds

stokesman:
Improvements: 
- Build time decreases more than 2x (average 19.30s down to 7.19s on my modest machine)
- Bundle size reduces from 26750 to 25205 (as reported by measure script)
- Less config, fewer dev dependencies

Builds are faster because instead of rollup using a typescript plugin which runs tsc once for every bundle, it simply uses what tsc outputs as the input for all bundles. Bundles are lighter because there was duplication of dependencies in umd bundles (which could be eliminated with yarn upgrade, no build changes required). Also somehow terser is outdoing uglify in minification by 1338 bytes (299 after gzip).
 
This started as a test to see if bundle size could reduce enough that even with a switch to Popmotion@8.7.3 the bundle would be smaller. That proved possible with less changes than I've made here but I got curious to see what could be improved.


InventingWithMonster:
That's really clever, nice simple solution! Do you know if this has any effect on code splittability?

stokesman:
I cannot see how it would have any effect on code-splitting. Rollup will have the same modules available for entry points only as js instead of ts. Would the question be in the interest of [`feature/size-reduction`](https://github.com/framer/motion/tree/feature/size-reduction)? I don't see the config for splits there but I'm guessing that's coming.

PR #597, opened 2020-06-11
Feature/navigation dev example

hemlok:
reproduces: https://github.com/framer/company/issues/18969

Trying to magic motion from a screen that was mounted without the AnimatedSharedLayout dependency changing causes subsequent magic motion transitions when dependency changes to fail 

Bug seems to be partially caused by https://github.com/framer/motion/blob/731e9685055389538f51c18884256901314f93ab/src/components/AnimateSharedLayout/index.tsx#L263-L264 where we check if `stack.lead.isPresent()`. Removing that specific check prevents the exception from being thrown as we measure the target, but prevents future animations from working 🤔


src/components/AnimateSharedLayout/index.tsx
```diff
@@ -248,23 +257,7 @@ export class AnimateSharedLayout extends React.Component<
...
-                    // If this component has an animate prop
-                    isAutoAnimate(child) ||
-                    // If this component is either entering or present
-                    child.presence !== Presence.Exiting ||
-                    // If the lead component in the stack is present and should animate, snapshot
-                    // TODO: Figure out what this breaks if removed
-                    (stack?.lead?.isPresent() && stack?.shouldStackAnimate())
-                ) {
-                    child.snapshotTarget()
-                }
```
hemlok, line 266: I know this is an optimisation, is it worth confirming if we need to delete it?

PR #633, opened 2020-07-16
[WIP] Adding layoutX and layoutY props

InventingWithMonster:
# The Problem

In Motion 2, layout animations and drag gestures are performed using **layout projection**. Once every frame, we calculate the transform needed to project an element from its actual layout, as transformed by parent transforms, into a desired viewport-relative bounding box.

As such, this transform is often seemingly arbitrary, and as such can't really be leveraged by users for any useful effects.

In Motion 1, it was often the case that people would hook into x/y transforms for effects like this: https://codesandbox.io/s/framer-motion-path-drawing-drag-and-usetransform-jnqk2. This is no longer possible with drag gestures in Motion 2.

# The Solution

This PR adds a new viewport-relative `delta` model that is saner for users. It is calculated only if the user indicates they want to "listen" to it by setting `onViewportBoxUpdate`, or the new `layoutX`/`layoutY` props.

`layoutX` and `layoutY` accept readonly motion values that can be hooked up the same way as the linked sandbox (see added example). They're "readonly" in the sense that they're not the source of truth and, like `scrollX`/`scrollY` from `useViewportScroll`, if set by a user they're not expected to have an effect.

# Question

Currently the `deltaViewport` calculated for `onViewportBoxUpdate` and the new props is based on the delta between the layout bounding box and the target bounding box **before user-set transforms are applied to it**. Is it potentially more useful for us to calculate it in relation to the target bounding box **after** transforms are applied? My intuition is to keep them separate, but I thought I'd get opinions.

# Bug fixes

This PR also fixes a bug where the user-defined `originX` and `originY` weren't correctly factored into the measurement of bounding boxes which lead to glitchy behaviours when mixing transforms with drag.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 Framer Motion adheres to [Semantic Versioning](http://semver.org/).
 
+## [2.2.0] 2020-07-20
+
+### Added
+
+-   Added `layoutX` and `layoutY` readonly MotionValue props for hooking other MotionValues into the layout calculations.
+
+### Fixed
+
+-   Fixed `originX` and `originY` being used incorrectly in bounding box measurements for layout children.
+
 ## [2.1.1] 2020-07-20
 
 ### Fixed
--- a/api/framer-motion.api.md
+++ b/api/framer-motion.api.md
@@ -322,6 +322,8 @@ export type KeyframesTarget = ResolvedKeyframesTarget | [null, ...CustomValueTyp
 export interface LayoutProps {
     layout?: boolean | "position";
     layoutId?: string;
+    layoutX?: MotionValue<number>;
+    layoutY?: MotionValue<number>;
     onLayoutAnimationComplete?(): void;
     onViewportBoxUpdate?(box: AxisBox2D, delta: BoxDelta): void;
 }
--- a/dev/examples/Drag-drag-to-reorder.tsx
+++ b/dev/examples/Drag-drag-to-reorder.tsx
@@ -44,6 +44,8 @@ const Item = ({ color, setPosition, moveItem, i }) => {
                     background: "white",
                     height: heights[color],
                     borderRadius: 5,
+                    originX: 0,
+                    originY: 0,
                 }}
                 whileHover={{
                     scale: 1.03,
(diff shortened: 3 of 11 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,16 @@
...
 
+## [2.1.0] 2020-07-16
+
+### Added
+
+-   Added `layoutX` and `layoutY` readonly MotionValue props for hooking other MotionValues into the layout calculations.
+
+### Fixed
+
+-   Fixed `originX` and `originY` being used incorrectly in bounding box measurements for layout children.
```
hemlok, line 13: I can't seem to see the code that effects this? 🤔

InventingWithMonster: https://github.com/framer/motion/pull/633/files#diff-1838108ae6ed35972d15efc143a8197bR102

hemlok: Ahh 👍

InventingWithMonster:
I agree I have reservations with the names. I don't want to call out the readonly nature of the props in the name but perhaps something like `deltaX` and `deltaY` that's calculated from the `deltaFinal` has less connotations that you might be able to set them?

hemlok:
>  perhaps something like `deltaX` and `deltaY` 

I quite like that!!

o-alexandrov:
Thank you, it would be huge for us, because we want to update `x` and `y` positions programmatically, based on user interaction outside of dragging events.
- for example, you rescale an image, `dragConstraints` changed and you want to update the position of an image inside the `dragConstraints`

Previously, as you mentioned in the original message and in your sandbox example, we were hooking up `useMotionValue()` to:
- `style={{ scale, x, y }}`

But after the update from `2.0.0-beta.77` to `2.1.4`, there is currently no solution for us, as you mention here.
As I understand, instead of `x` and `y`, we would be connecting to `deltaX` (and Y) or `layoutX` (and Y).

Please correct me, if I am wrong and thank you for such an easy to use, for your happy users like me, library

InventingWithMonster:
Added WIP - I'm going to investigate a different approach here that looks at consolidating x/y transforms with layout projection.

Related issue: https://twitter.com/alexbassy/status/1288128857577422849

Issue #637, opened 2020-07-16
[BUG] Layout distortion correction doesn't work in response to child changes


dminkovsky:
Please have a look at https://codesandbox.io/s/framer-motion-2-layout-animations-pei0x?file=/src/App.js

Based on the docs and other examples, you would expect this to work. However, it does this:

![ezgif-4-2224cfaf3dda](https://user-images.githubusercontent.com/218725/87727327-c6a1d700-c78e-11ea-806f-bbfb1be406a9.gif)

Oddly, when you change `layout={true}` to `layout={false}` on the inner `motion.div` that's supposed to perform the correction, despite what the docs say (that `layout={true}` should be set to perform the correction), the functionality starts working:

![ezgif-4-39ff8ac09309](https://user-images.githubusercontent.com/218725/87727261-a07c3700-c78e-11ea-9a38-852e4e5c1ac7.gif)

Stranger still: if you save `layout={false}` and reload, it stops working. This behavior is consistent for me in Chrome and FF.

InventingWithMonster:
Hmm this is a tricky one. By providing the `layout` prop to both elements they both perform layout animations. Because the text content is changing in the child, this looks like distortion, but in essence this is the correct behaviour. You can test this by popping it out of the parent - it still distorts in the exact same way.

Perhaps what is needed is a setting that doesn't perform any layout animations and allows for the instant switching of layout but maintains the component's scale, something like:

```jsx
<motion.div layout="preserve" />
```

eschaefer:
> `<motion.div layout="preserve" />`

TypeScript gave me feedback that `"preserve"` wasn't a valid value for `layout`, but `"position"` worked! 

Was experiencing a similar distortion issue and `<motion.div layout="position" />` fixed it.

InventingWithMonster:
@eschaefer Sorry yeah I landed on position. Glad that fixed it for you!

PR #638, opened 2020-07-17
Adding layout="position" option

InventingWithMonster:
Fixes https://github.com/framer/motion/issues/637

## The Problem

Currently, the recommended method of correcting scale on child elements of a layout animation is by adding `layout` to immediate children, essentially putting those into a layout animation too.

This works well in most situations but it doesn't tend to work well on elements you wouldn't otherwise want to put into a layout animation.

For instance, the text in this example:

before

By animating this layer between its two bounding box sizes, the text is becoming distorted (correctly).

## The Solution

Adding a setting for `layout`, `"position"`.

When `layout="position"`, we disable size animations by resizing the origin bounding box based on the target box.

after2

position

### DIFF
--- a/src/motion/features/layout/Animate.tsx
+++ b/src/motion/features/layout/Animate.tsx
@@ -14,6 +14,7 @@ import {
 } from "../../../components/AnimateSharedLayout/types"
 import { mix, circOut, linear, progress } from "@popmotion/popcorn"
 import { usePresence } from "../../../components/AnimatePresence/use-presence"
+import { LayoutProps } from "./types"
 
 interface AxisLocks {
     x?: () => void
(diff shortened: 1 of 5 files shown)


src/motion/features/layout/Animate.tsx
```diff
@@ -72,6 +74,15 @@ class Animate extends React.Component<AnimateProps> {
...
 
         const animations = eachAxis(axis => {
+            /**
+             * If layout is set to "position", we can resize the origin box based on the target
+             * box and only animate its position.
+             */
+            if (layout === "position") {
+                const targetLength = target[axis].max - target[axis].min
+                origin[axis].max = origin[axis].min + targetLength
+            }
```
hemlok, line 84: Just so I'm clear, we're adding to the max so that the later calculation of the width/height difference doesn't create a flip'd scale, right?

InventingWithMonster: Essentially yeah, it'll be FLIPped in terms of position but not scale.

PR #645, opened 2020-07-20
Layout projection performance

InventingWithMonster:
This PR changes the way layout deltas are updated per frame.

When a component has its viewport box updated, we need to update its transform delta so it correctly projects to this new bounding box.

For these calculations to be correct, we need to make sure all the component's parents are updated too. Currently, each component traverses its parents first and make sure those components have run their calculations.

This is safe in that it means the calculations are always correct but it's quite wasteful as it means that given a tree

```
A
 |-B
     |-C
```

The update methods for A will run three times if all three components are updating that frame.

This can been very wasteful - in the added dev/examples/Animation-layout-update-stress.tsx example, for 84 components we have ~550 updates(!). This is quite a common size for Magic Motion docs.

With this update we instead, when any component's viewport box is updated, schedule a function to run in the root component of that tree at the end of the current update step in `framesync`. This function traverses the tree, updating the layout box for each in turn. For 84 components we only update 84 times.

### DIFF
--- a/src/render/dom/HTMLVisualElement.ts
+++ b/src/render/dom/HTMLVisualElement.ts
@@ -16,12 +16,17 @@ import {
     applyBoxTransforms,
     removeBoxTransforms,
 } from "../../utils/geometry/delta-apply"
-import { updateBoxDelta } from "../../utils/geometry/delta-calc"
+import {
+    updateBoxDelta,
+    updateTreeScale,
+} from "../../utils/geometry/delta-calc"
 import { Transition } from "../../types"
 import { eachAxis } from "../../utils/each-axis"
 import { motionValue, MotionValue } from "../../value"
 import { startAnimation } from "../../animation/utils/transitions"
 import { getBoundingBox } from "./layout/measure"
+import sync, { getFrameData } from "framesync"
+import { createDeltaTransform } from "./utils/project-layout"
 
 export type LayoutUpdateHandler = (
     layout: AxisBox2D,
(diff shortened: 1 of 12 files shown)


src/render/dom/HTMLVisualElement.ts
```diff
@@ -423,39 +412,46 @@ export class HTMLVisualElement<
...
         resetBox(this.boxCorrected, this.box)
 
+        /**
+         * If this component has a parent, update this treeScale by incorporating the parent's
+         * delta into its treeScale.
+         */
+        if (this.parent) {
+            updateTreeScale(
+                this.treeScale,
+                (this.parent as any).treeScale,
```
InventingWithMonster, line 447: Casting here as treeScale doesn't currently exist on VisualElement. I'd prefer to leave it off until we work on writing the ThreeVisualElement so we can see what abstracts nicely.

src/render/dom/HTMLVisualElement.ts
```diff
@@ -423,39 +414,46 @@ export class HTMLVisualElement<
...
+                (this.parent as any).delta
+            )
+        }
+
         /**
          * Apply all the parent deltas to this box to produce the corrected box. This
          * is the layout box, as it will appear on screen as a result of the transforms of its parents.
          */
-        applyTreeDeltas(this.boxCorrected, this.treeScale, this.treePath as any)
+        applyTreeDeltas(this.boxCorrected, this.treePath as any)
```
InventingWithMonster, line 456: Because we've already updated parent deltas we can simply use the latest treeScale value in updateTreeScale rather than traversing the tree a second time. We still do run this loop but we're doing less work per iteration as a result.

src/render/dom/HTMLVisualElement.ts
```diff
@@ -485,13 +483,20 @@ export class HTMLVisualElement<
...
         /**
          * If we have a listener for the viewport box, fire it.
-         * TODO: Instead of manually checking this, use framesync postRender
          */
-        if (!isReactRender) {
-            this.hasViewportBoxUpdated &&
-                this.config.onViewportBoxUpdate?.(this.targetBox, this.delta)
-            this.hasViewportBoxUpdated = false
-        }
+        this.hasViewportBoxUpdated &&
```
InventingWithMonster, line 487: This function is no longer being triggered from build so it won't fire synchronously with a React render

PR #661, opened 2020-07-24
Fixing Drag for SVG components

InventingWithMonster:
This PR adds layout projection to SVG components. So far this is just to reintroduce drag, but is the foundation of layout animations too.


hemlok:
I'm seeing some issues on this example `_layoutDragTransform`

Screen Recording 2020-07-24 at 12 43 30

PR #668, opened 2020-07-27
Support rotate in layout animations

InventingWithMonster:
In Magic Motion, we support user-set rotation on layers that will be part of a layout animation. This PR re-introduces that capability into Framer Motion.

## Problem

Accurate bounding boxes can't be measured when rotation or skew transforms are applied. Therefore, before measuring any kind of bounding box accurately, we must first unset these transforms.

## Solution

By providing an internal `_supportRotate` prop to `AnimateSharedLayout`, we can instruct it to reset rotation on all children before snapshotting their previous viewport box.


src/render/dom/utils/__tests__/build-html-styles.test.ts
```diff
@@ -0,0 +1,261 @@
...
+        build(latest, { style })
+
+        expect(style).toEqual({
+            transform:
+                "translateX(1px) translateY(2px) rotateX(90deg) translateZ(0)",
+        })
+    })
+
+    test("Builds transform with defined value types", () => {
+        const latest = { x: "1vw", y: "2%", rotateX: "90turn" }
```
hemlok, line 36: "turn" TIL 🤯

src/render/dom/utils/build-html-styles.ts
```diff
@@ -52,8 +57,8 @@ export function buildHTMLStyles(
     transformKeys.length = 0
 
     // Track whether we encounter any transform or transformOrigin values.
-    let hasTransform = !!isLayoutProjectionEnabled
-    let hasTransformOrigin = !!isLayoutProjectionEnabled
+    let hasTransform = false
+    let hasTransformOrigin = false
```
nvh, line 61: Why don't we base this of the layoutProjectionEnabled anymore?

src/render/dom/utils/build-html-styles.ts
```diff
@@ -133,25 +161,10 @@ export function buildHTMLStyles(
                 enableHardwareAcceleration,
                 allowTransformNone
             )
-        } else {
-            style.transform = createDeltaTransform(deltaFinal!, treeScale!)
-            if (transformTemplate)
-                style.transform = transformTemplate(transform, style.transform)
```
nvh, line 139: I don't understand where we apply the transformTemplate now?

src/render/dom/utils/build-html-styles.ts
InventingWithMonster: Before the way the later logic was written, we always wanted resolve *some* transform if layout projection was enabled.

Whereas now the logic is slightly different in that if isLayoutProjectionEnabled is true, we still resolve the projection transform, but we *also* want to know if the user has set a transform so we can apply rotate and skew if so.

src/render/dom/utils/build-html-styles.ts
InventingWithMonster: Ah yes I believe the answer is we don't - this is fixed now

PR #669, opened 2020-07-28
Adding useTemplate

InventingWithMonster:
This PR adds the `useTemplate` hook, which allows users to combine multiple motion values into a single string using template literals.

```jsx
import {
  motion,
  useSpring,
  useMotionValue,
  useTemplate
} from "framer-motion"
 
function Component() {
  const shadowX = useSpring(0)
  const shadowY = useMotionValue(0)
  const shadow = useTemplate`drop-shadow(${shadowX}px ${shadowY}px 20px rgba(0,0,0,0.3))`
 
  return <motion.div style={{ filter: shadow }} />
}
```


src/value/use-template.ts
```diff
@@ -0,0 +1,81 @@
...
+     * Synchronously update the motion value with the latest values during the render.
+     * This ensures that within a React render, the styles applied to the DOM are up-to-date.
+     */
+    updateValue()
+
+    /**
+     * Subscribe to all motion values found within the template. Whenever any of them change,
+     * schedule an update.
+     */
+    useSubscriptions(values, () => sync.update(updateValue, false, true))
```
nvh, line 71: I'm not sure about the rest of the terminology used in motion, but as the motionValue function that is called eventually is called `onChange` _and_ the subscriptions are not available outside of this hook, wouldn't it be better to name this hook similarly? Maybe `useOnChangeValue`, `useValueChangeHandlers`?

src/value/use-template.ts
```diff
@@ -0,0 +1,81 @@
...
+ *   const shadowY = useMotionValue(0)
+ *   const shadow = useTemplate`drop-shadow(${shadowX}px ${shadowY}px 20px rgba(0,0,0,0.3))`
+ *
+ *   return <motion.div style={{ filter: shadow }} />
+ * }
+ * ```
+ *
+ * @public
+ */
+export function useTemplate(
```
nvh, line 28: Is the name `useTemplate` too generic for this purpose? I'm on the fence.

hemlok: Had the same thought. Maybe `useMotionValueTemplate`? 🤔

hemlok:
Literally ran into the use case for this over the weekend, nice one 🎉

src/value/use-template.ts
InventingWithMonster: The [original name](https://gist.github.com/InventingWithMonster/38d7cdd3557f398c2ca4a2c81defed08) was `useMotionTemplate` which I feel is a nice mix of specificity and brevity? 

I shortend it because we have a few motion value factories like `useSpring` and `useTransform` without "motion" or "value", but I do agree that out of context, "spring" and to a lesser extend "transform" more clearly live in an animation-esque context than "value" and "template"

nvh: Let's go with useMotionTemplate, then!

Issue #251, opened 2019-07-29
useRelative


steveruizok:
The `useRelative` hook allows for complex transforms against _multiple_ values or motion values.

![ScreenFlow](https://user-images.githubusercontent.com/23072548/62044331-8999a900-b1fa-11e9-9263-5c398ad6a658.gif)

```tsx
const averageX = useRelative(
    (a, b, c) => (a + b + c) / 3, 
    xA,
    xB,
    xC
)
```

The hook works by taking a _computing_ function and one or more _related values_. These related values may be numbers, strings, or motion values. When any of the related values change, the hook will compute a new value for its motion value by running the current value of each of its related values through the computing function.

The computed values may be quite complex – and they may also include _other computed values_. However, the resulting code is nonetheless declarative, highly efficient, and fully juices the creative plane.

You can preview the `useRelative` hook on the [useRelative-useBetween branch](https://github.com/framer/motion/tree/useRelative-useBetween). (An earlier version of this branch also included a second hook, `useBetween`). 

## Limitations

At the moment, each related value passed to the hook must be either of the same type (e.g. all `number`s) or a mix of values and motion values of the same type (e.g. `number`s and `MotionValue<number>`s). This is not a hard constraint: the hook should be able to take any values and pass them into the callback accordingly; however, this may be beyond the limits of my TypeScript-Fu.

## Background

The `useTransform` hook is great for computing a motion value from the result of some other motion value.

```tsx
const shadowX= useTransform(xA, v => v - 4)
```

However, the library does not support computed values that are the result of several other values. For example, if I wanted a motion value that was between two other motion values, and which would update when _either_ of those values changed, the solution would be pretty tedious:

```tsx
const betweenX = useMotionValue(0)

const calc = React.useCallback((a, b) => {
   betweenX.set(a + (b - a) / 2)
}, [])

React.useEffect(() => {
  const cancelA = xA.onChange(v => calc(v, xB.get()))
  const cancelB = xB.onChange(v => calc(xA.get(), v))
  return () => {
    cancelA()
    cancelB()
  }
}, [])
```

This hook greatly simplifies this (common) part of complex interfaces/animations.

## Other options

This hook could possibly be merged into the `useTransform` hook.

```
// Simple
useTransform(xA, v => v - 4)

// Range interpolation
useTransform(xA, [0, 100], [32, 68])

// Relative
useTransform([xA, xB], (a, b) => (a + b)  / 2)
```

InventingWithMonster:
[...]

PR #696, opened 2020-08-10
Accept multiple MotionValues in useTransform

InventingWithMonster:
Closes https://github.com/framer/motion/issues/251

This PR upgrades `useTransform` to accept a list of `MotionValue`s.

```jsx
const a = useMotionValue(0)
const b = useMotionValue(0)
const c = useTransform([a, b], ([latestA, latestB]) => latestA * latestB)
```

### DIFF
--- a/src/value/__tests__/use-transform.test.tsx
+++ b/src/value/__tests__/use-transform.test.tsx
@@ -36,6 +36,26 @@ describe("as function", () => {
     })
 })
 
+describe("as function with multiple values", () => {
+    test("sets initial value", async () => {
+        const Component = () => {
+            const x = useMotionValue(4)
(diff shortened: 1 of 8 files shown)


src/value/__tests__/use-transform.test.tsx
```diff
@@ -53,12 +73,15 @@ describe("as input/output range", () => {
             const x = useMotionValue(100)
             const opacity = useTransform(x, [0, 200], [0, 1])
 
-            x.set(20)
+            React.useEffect(() => {
```
InventingWithMonster, line 76: Subscriptions are now added via useEffect, which is the recommended concurrent-safe way to establish subscriptions in React. So these tests have been updated to reflect that.

PR #701, opened 2020-08-11
Stub out SharedLayoutTree

hemlok:
Todo:
- [x] Implement _shouldLayoutAnimate callback
- [x] Bring over dev example from poc pr
- [ ] Bugs 🐛

### DIFF
--- a/src/components/AnimateSharedLayout/SharedLayoutTree.tsx
+++ b/src/components/AnimateSharedLayout/SharedLayoutTree.tsx
@@ -1,0 +1,108 @@
+import * as React from "react"
+import {
+    SharedLayoutContext,
+    SharedLayoutSyncMethods,
+} from "./SharedLayoutContext"
+import { HTMLVisualElement } from "render/dom/HTMLVisualElement"
+import { Presence } from "./types"
+
+export interface SharedLayoutTreeProps {
+    /**
+     * @public
+     */
+    children: React.ReactNode
+
+    /**
+     * Are the children of the tree valid targets when computing lead and follow.
+     */
+    isPresent: boolean | undefined
+    /**
+     * Are the children of the tree valid targets when computing lead and follow.
+     */
+    isValidTree: boolean
+
+    /**
+     * What priority should children of this tree take when determining the lead and follow.
+     */
+    layoutOrder: number
+
+    isLead: boolean
+}
+
+interface SyncProps extends SharedLayoutTreeProps {
+    syncContext: SharedLayoutSyncMethods
+}
+
+export class LayoutTree extends React.Component<SyncProps> {
+    /**
+     * A list of all the children in the tree
+     */
+    children = new Set<HTMLVisualElement>()
+
+    isPresent? = true
+    isValid = true
+    layoutOrder = 0
+
+    /**
+     * The methods provided to all children in the shared layout tree.
+     */
+    syncContext: SharedLayoutSyncMethods = {
+        ...this.props.syncContext,
+        register: child => this.addChild(child as HTMLVisualElement),
+        remove: child => this.removeChild(child as HTMLVisualElement),
+    }
+
+    componentDidMount() {
+        // register the tree object with the parent sync context
+        this.layoutOrder = this.props.layoutOrder
+        this.props.syncContext.register(this)
+    }
+
+    componentWillUnmount() {
+        this.props.syncContext.remove(this)
+    }
+
+    addChild(child: HTMLVisualElement) {
+        this.children.add(child)
+        child.presence = Presence.Entering
+    }
+
+    removeChild(child: HTMLVisualElement) {
+        this.children.delete(child)
+    }
+
+    shouldComponentUpdate(prevProps: SharedLayoutTreeProps) {
+        const { isPresent, isValidTree, layoutOrder, isLead } = this
+            .props as SharedLayoutTreeProps
+
+        this.isPresent = isPresent
+        this.isValid = isValidTree
+        if (layoutOrder === undefined) return true
+
+        const hasOrderChanged = prevProps.layoutOrder !== layoutOrder
+        // const hasPresenceChanged = prevProps.isPresent !== isPresent
+        const hasBecomeLead = isLead && !prevProps.isLead
+
+        if (hasBecomeLead || hasOrderChanged)
+            this.children.forEach(child => (child.presence = Presence.Entering))
+
+        if (hasBecomeLead) return true
+
+        return false
+    }
+
+    render() {
+        return (
+            <SharedLayoutContext.Provider value={this.syncContext}>
+                {this.props.children}
+            </SharedLayoutContext.Provider>
+        )
+    }
+}
+
+export const SharedLayoutTree = (props: SharedLayoutTreeProps) => {
+    const syncContext = React.useContext(
+        SharedLayoutContext
+    ) as SharedLayoutSyncMethods
+    return <LayoutTree {...props} syncContext={syncContext} />
+}
--- a/src/components/AnimateSharedLayout/index.tsx
+++ b/src/components/AnimateSharedLayout/index.tsx
@@ -1,7 +1,7 @@
 import * as React from "react"
 import { SharedLayoutProps, Presence } from "./types"
 import { createCrossfadeAnimation, createSwitchAnimation } from "./animations"
-import { LayoutStack } from "./stack"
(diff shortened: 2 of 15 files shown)


src/components/AnimateSharedLayout/SharedLayoutTree.tsx
```diff
@@ -0,0 +1,108 @@
...
+     * Are the children of the tree valid targets when computing lead and follow.
+     */
+    isValidTree: boolean
+
+    /**
+     * What priority should children of this tree take when determining the lead and follow.
+     */
+    layoutOrder: number
+
+    isLead: boolean
```
hemlok, line 29: This is KINDA just here to trigger an update on the tree we're animating too in the scenario where the layoutOrder doesn't update.

src/components/AnimateSharedLayout/SharedLayoutTree.tsx
```diff
@@ -0,0 +1,108 @@
...
+}
+
+interface SyncProps extends SharedLayoutTreeProps {
+    syncContext: SharedLayoutSyncMethods
+}
+
+export class LayoutTree extends React.Component<SyncProps> {
```
hemlok, line 36: I started with this extending AnimateSharedLayout but there were circular dependancies and it was getting more confusing dealign with the types than just breaking it off. Happy to reconsider if you had something in mind though!

src/components/AnimateSharedLayout/SharedLayoutTree.tsx
```diff
@@ -0,0 +1,108 @@
...
+    isValid = true
+    layoutOrder = 0
+
+    /**
+     * The methods provided to all children in the shared layout tree.
+     */
+    syncContext: SharedLayoutSyncMethods = {
+        ...this.props.syncContext,
+        register: child => this.addChild(child as HTMLVisualElement),
+        remove: child => this.removeChild(child as HTMLVisualElement),
```
hemlok, line 52: Use the parent syncContext but just overnight register and remove so that the children for this tree stay scoped to `this.children`

src/components/AnimateSharedLayout/SharedLayoutTree.tsx
```diff
@@ -0,0 +1,108 @@
...
+        this.isPresent = isPresent
+        this.isValid = isValidTree
+        if (layoutOrder === undefined) return true
+
+        const hasOrderChanged = prevProps.layoutOrder !== layoutOrder
+        // const hasPresenceChanged = prevProps.isPresent !== isPresent
+        const hasBecomeLead = isLead && !prevProps.isLead
+
+        if (hasBecomeLead || hasOrderChanged)
+            this.children.forEach(child => (child.presence = Presence.Entering))
```
hemlok, line 87: This is kinda just to work inside the existing stack logic that chooses origins based on presence...

src/components/AnimateSharedLayout/index.tsx
```diff
@@ -140,8 +145,30 @@ export class AnimateSharedLayout extends React.Component<SharedLayoutProps> {
...
+         * If AnimateSharedLayout is the parent of SharedLayout Trees,
+         * generate new stacks based on the contents of the Lead and Follow trees.
+         */
+        return this.isSharedLayoutTreeParent
+            ? this.updateTree()
+            : this.stacks.forEach(stack => stack.updateLeadAndFollow())
+    }
+
+    updateTree() {
+        this.tree.sortOrder()
```
hemlok, line 169: Instead of removing and registering whole trees, just sort the tree in place before we calc the lead and follow so that the latest layoutOrder is last

src/components/AnimateSharedLayout/stack.ts
```diff
@@ -210,17 +205,29 @@ export class LayoutStack {
...
-        //     : this.follow && this.follow?.props._shouldAnimate === true
+    }
+
+    shouldPrimaryAnimate(
+        primary: HTMLVisualElement,
+        secondary?: HTMLVisualElement
+    ) {
+        if (primary.config._shouldAnimateLayout === undefined) return true
+        secondary && secondary.config._shouldAnimateLayout?.(false)
+        return primary.config._shouldAnimateLayout(true) as boolean
```
hemlok, line 224: I'd love thoughts on a better way to do, or at very least name this. 

Essentially we need to call a function defined on the follow to ensure that the framer shouldAnimate context has the follows `props` assigned to the layoutId. That way when `_shouldAnimateLayout()` is called on the lead, we can compare props and return a boolean.

Framer side, we're creating a callback like `(boolean) => boolean ? calculate() : register()`.

So we could do something like _prepareShouldAnimate(): void and remove the need to pass a boolean, but my gut feeling is to keep the surface area of this framer-specific stuff as small as possible, hence the single prop 🤔

src/components/AnimateSharedLayout/stack.ts
```diff
@@ -210,17 +205,29 @@ export class LayoutStack {
...
     }
 
     getFollowOrigin(): AxisBox2D | undefined {
         // This shouldAnimate check is quite specifically a fix for the optimisation made in Framer
         // where components are kept in the tree ready to be re-used
-        return this.follow // && this.follow.shouldAnimate
+        return this.follow
```
hemlok, line 230: I think I removed the commented flag in a PR after `main` was forked to fix a framer bug

src/components/AnimateSharedLayout/stack.ts
```diff
@@ -241,3 +248,62 @@ export class LayoutStack {
...
+        if (this.follow) {
+            for (const child of this.follow.children) {
+                if (!child.layoutId) continue
+
+                const stack = new LayoutStack()
+
+                child.isPresent = this.follow.isPresent
+                stack.add(child)
+                stacks.set(child.layoutId, stack)
+                stack.updateLeadAndFollow()
```
hemlok, line 285: Maybe if we could instantiate LayoutStack with a child that would set the single child to the lead and avoid having to call `updateLeadAndFollow` early here? Essentially I just want to ensure that the stacks created for layoutId's that aren't present in the lead tree are setup right

Alternatively maybe this isn't even necessary, if there is no follow does the stack even animate and need to have a lead? 🤔

src/components/AnimateSharedLayout/stack.ts
```diff
@@ -241,3 +248,62 @@ export class LayoutStack {
...
+
+                let stack = stacks.get(child.layoutId)
+                const existingStack = !!stack
+
+                stack = stack || new LayoutStack()
+
+                child.isPresent = this.lead.isPresent
+
+                stack.add(child)
+                stack.updateLeadAndFollow()
```
hemlok, line 301: Calling updateLeadAndFollow seems to make sense here since it allows us to avoid another loop though

tsconfig.json
```diff
@@ -30,7 +30,8 @@
         "noUnusedParameters": true,
         "removeComments": false,
         "lib": ["es2017", "dom"],
-        "skipLibCheck": true
+        "skipLibCheck": true,
+        "downlevelIteration": true
```
hemlok, line 34: I thought I saw you add this in a different PR but maybe not?

src/components/AnimateSharedLayout/stack.ts
```diff
@@ -210,17 +205,29 @@ export class LayoutStack {
...
+        if (this.follow)
+            return this.shouldPrimaryAnimate(this.follow, this.lead)
+
         return true
-        // return this.lead && this.lead?.isPresent
-        //     ? this.lead?.props?._shouldAnimate === true
-        //     : this.follow && this.follow?.props._shouldAnimate === true
+    }
+
+    shouldPrimaryAnimate(
```
InventingWithMonster, line 218: What's primary/secondary vs lead/follow?

src/motion/features/layout/Animate.tsx
```diff
@@ -158,8 +164,12 @@ class Animate extends React.Component<AnimateProps> {
          * If this is a crossfade animation, create a function that updates both the opacity of this component
          * and the one being crossfaded out.
          */
-        const crossfade =
-            crossfadeOpacity && this.createCrossfadeAnimation(crossfadeOpacity)
+        let crossfade: (p: number) => void
+        crossfadeOpacity && this.createCrossfadeAnimation(crossfadeOpacity)
```
InventingWithMonster, line 168: I think this line needs deleting

Issue #699, opened 2020-08-11
[BUG] delayChildren only works one level in tree


rijk:
**Describe the bug**

Not sure if this is a bug or works as intended, [the docs on this](https://www.framer.com/api/motion/types/#orchestration.delaychildren) are relatively sparse.

I am trying to orchestrate the timing of fading in from a parent component. There are multiple levels of animation in the child components; for example a fade in of the container + a staggered slide in for each line of text. 

While trying this using variants, I found that **only the direct children** of the parent follow the delay set via `delayChildren`. _Their_ children just animate instantly, without delay.

**Provide a CodeSandbox reproduction of the bug**

https://codesandbox.io/s/nameless-https-ywzfh

**Steps to reproduce**

1. Go to Sandbox
2. Click on "Go to other page"
3. Note that the text scales up with a 1.5s delay — but also note that the second paragraph is already visible from the start

**Expected behavior**

The second paragraph should only start to fade in after the same 1.5s delay defined in the parent’s `delayChildren`.

InventingWithMonster:
This is a bug - delays are intended to be cumulative.

PR #704, opened 2020-08-12
Fixing delay propagation

InventingWithMonster:
Fixes https://github.com/framer/motion/issues/699

Fixed the propagation of `delay` through children and added a little cleaning up with the animateVariant method.

### DIFF
--- a/src/animation/VisualElementAnimationControls.ts
+++ b/src/animation/VisualElementAnimationControls.ts
@@ -8,7 +8,6 @@ import {
     TargetAndTransition,
     Variant,
     TargetWithKeyframes,
-    Orchestration,
 } from "../types"
 import { VariantLabels, MotionProps } from "../motion/types"
 import { resolveFinalValueInKeyframes } from "../utils/resolve-value"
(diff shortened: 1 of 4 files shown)


src/animation/VisualElementAnimationControls.ts
```diff
@@ -552,48 +551,55 @@ export class VisualElementAnimationControls<
...
-                when = transition.when || when
-                delayChildren = transition.delayChildren || delayChildren
-                staggerChildren = transition.staggerChildren || staggerChildren
-                staggerDirection =
-                    transition.staggerDirection || staggerDirection
-            }
-        }
-
+        /**
+         * If the transition explicitly defines a when option, we need to resolve either
```
nvh, line 588:
```suggestion
         * If the transition explicitly defines a "when" option, we need to resolve either
```

src/animation/VisualElementAnimationControls.ts
```diff
@@ -552,48 +551,55 @@ export class VisualElementAnimationControls<
...
             ? () => this.animate(variant, opts)
             : () => Promise.resolve()
 
-        const getChildrenAnimations: () => Promise<any> = this.children
-            ? () => {
+        /**
+         * If we have children, create a callback that runs all their animations.
+         * Otherwise, we resolve a Promise immediately for a composable no-op.
+         *
+         *
```
nvh, line 572: The lines with just stars can be removed, right?

src/animation/VisualElementAnimationControls.ts
```diff
@@ -552,48 +551,55 @@ export class VisualElementAnimationControls<
     }
 
     private animateVariant(variantLabel: string, opts?: AnimationOptions) {
-        let when: Orchestration["when"] = false
-        let delayChildren = 0
-        let staggerChildren = 0
-        let staggerDirection = 1
```
nvh, line 558: Where are these defaults now defined?

InventingWithMonster: Ah they were already double defined in animateChildren

Issue #577, opened 2020-05-27
[BUG] modifyTarget is causing a jumpy transition 


Elia-Darwish:
**2. Describe the bug**

Using modifyTarget on a Draggable's element dragTransition is resulting in a jumpy behavior, this is happening on drag end where it seems like the element is skipping a bit ahead!

**3. IMPORTANT: Provide a CodeSandbox reproduction of the bug**

https://codesandbox.io/s/polished-waterfall-u037u?file=/src/App.js

**5. Expected behavior**

a smooth page transition like animation like the one available to use in Framer

**6. Video or screenshots**

here is a video that shows how it is behaving:
https://res.cloudinary.com/elia-d/video/upload/v1590566454/Screen_Recording_2020-05-27_at_09.49.06.mov

stokesman:
This could be fixed with an update of the Popmotion dependency. I've experimented with couple ways to remove any potential blocks of that update. First, #596 was to make possible a switch to Popmotion ^8.7.3 without bundle size increases. Second, I made [a branch of Popmotion](https://github.com/stokesman/popmotion/tree/8omega) to work as a publishable beta-9. I would have made a PR with that branch to discuss but there's not a reasonable branch to make the request against (feature/pure-9 is well ahead).

PR #709, opened 2020-08-18
Upgrading to Popmotion 9 RC

InventingWithMonster:
Closes https://github.com/framer/motion/issues/577

This PR upgrades to the latest Popmotion 9 release candidate. In addition to fixing the linked bug, it will allow us (in future PRs) to:

- Support alternative animation drivers (including sync for test environments, and XR devices for our `react-three-fibre` work)
- Support the new proposed repeat syntax `repeat: 3, repeatType: "loop"`
- Support repeat options for spring animations
- Remove the prerequisite of explicitly adding `type: "spring"` if spring animation can be inferred by `stiffness`/`damping`/`mass` options


src/motion/__tests__/animate-prop.test.tsx
```diff
@@ -169,7 +168,7 @@ describe("animate prop as object", () => {
             }
             const Component = ({ rotate, onComplete }: any) => (
                 <motion.div
-                    animate={{ rotate }}
+                    animate={{ rotate: `${rotate}deg` }}
```
InventingWithMonster, line 171: We didn't explicitly support mixing number and string types before, this was basically a slip.

PR #714, opened 2020-08-19
Allowed automatic keys in AnimatePresence

reductio:
Using this change, we can cover more cases to use AnimatePresence without a key and make it feel more like "normal" react. Objects without keys get automatic ".x" keys. This way, this small sample code works, because the first Block always automatically receives ".1" as key, and the second Block always gets ".2", even if only one of them is rendered.

`<AnimatePresence>
        {condition && <Block x={0} y={0} className="block left" />}
        {!condition && <Block x={0} y={200} className="block right" />}
      </AnimatePresence>`

Still, ternary expressions like `{condition ? <Block x={0} y={0} className="block left" /> : <Block x={0} y={200} className="block right" />}` won't work, but I think it's still a nice improvement. Please let me know what you think.


mattgperry:
Thanks for the PR but I'd rather keep this rule "always add key" simple to reduce ambiguity. The example in this PR is a great example - the ternary expression *with* key would be simpler than the two separate statements. Reducing the number of ways in which AnimatePresence works correctly also makes it easier to spot bugs in user code.

PR #734, opened 2020-08-26
Fix default transitions

InventingWithMonster:


src/animation/utils/transitions.ts
```diff
@@ -186,7 +192,7 @@ export function startAnimation(
     target: ResolvedValueTarget,
     transition: Transition = {}
 ) {
-    return value.start(onComplete => {
+    return value.start((onComplete) => {
```
hemlok, line 195: Prettier update?

hemlok (approved):

src/animation/utils/transitions.ts
InventingWithMonster: yeah i believe this the default now

hemlok: Is it worth running on the whole project to get it out of the way? Or setting updating the config to not bring this change? I think it's just a bit distracting if every PR from now one introduces these formatting changes.

Issue #706, opened 2020-08-13
[BUG] Adding transitionEnd ignores transition


selbekk:
When I apply a `transitionEnd`, the `transition` doesn't seem to be respected anymore. It's a bug better described through an example, which you'll find below.

Here's a code sandbox where you can switch between the bug being present (with transitionEnd) and the bug not being there.

https://codesandbox.io/s/eloquent-leftpad-1rhmj?file=/src/App.js

**Steps to reproduce**

1. Go to the sandbox.
2. Click the big coral question mark
3. Notice that there is no delay before the feedback box pops up - this is the bug!
4. Now, click the "Turn off transitionEnd" button, and do the same
5. Notice that everything is buttery smooth now.

**Expected behavior**

I would expect the delay to still be valid, or at least get a warning in the console if this is a pattern that's not supported.

selbekk:
If I switch out `display: none|block` with `visibility: hidden|visible` it works fine though - perhaps this is a bug that hits the `display` property directly?

IvanIsakof:
Hello, this problem is similar to mine! See https://github.com/framer/motion/issues/648

InventingWithMonster:
@selbekk Re: the finding in your second comment. I think you'll find it will also work correctly if you force all the `0`s to be percentages, or the `"200%"`s to be absolute pixel values.

The reason this and the `visibility` thing works is to convert between different value types we need to measure the before/after states of the component. `visibility: hidden` leaves the component rendered as part of the layout so the conversion works ok, while forcing both values to be of the same type skips the conversion completely (better performance)

It's unlikely I can fix this because there's no good way of knowing what the `display` prop of a component would otherwise be if we find it as `none`.

However what I can look at, which would probably have avoid the bug in this instance, is checking whether we're animating between `0` and another value type and coercing `0` to that value type (as 0 anything is no transform) rather than running the whole conversion process.

PR #735, opened 2020-08-26
Coerce zero to unit type

InventingWithMonster:
Fixes https://github.com/framer/motion/issues/706

Motion implements its ability to animate between unit types by measuring the before/after state and animating via pixels. The way it does this is less sophisticated than the layout animation implementation so it can be a performance issue. It also causes problems when we involved `display: none` as we can't measure the DOM in these instances.

This PR adds the detection of a common pattern where a value is animated from `0` (considered pixels) and another value type and simply coerces `0` into the unit type.


hemlok (approved):
Yesssss 💪

Issue #752, opened 2020-09-06
[BUG] useTransform changes behaviour from old to new version


mariasimo:
**Describe the bug**

I'm trying to use useTransform to change the styles on scroll position. I'm reusing a snippet i had used successfully in a previous project (with version 1.1.1 of Framer Motion installed) but now the behavior is not the same. 

```jsx
  const [elementTop, setElementTop] = useState(null);

  useLayoutEffect(() => {
    if (!ref.current) return;
    setElementTop(ref.current.offsetTop);
  }, [ref]);
  
  const opacity = useTransform(scrollY, [elementTop, elementTop + 600], [1, 0]);
```
This works ok for previous versions, but in current version elementTop arrives to useTransform as a null value and final value of ouput range is applied inmediatly

**IMPORTANT: Provide a CodeSandbox reproduction of the bug**
https://codesandbox.io/s/solitary-meadow-6du76

**Steps to reproduce:** 
1. Scroll and observe behavior: opacity of bottom div becomes 0 before element is in viewport
2. Change version to 2.3.0. and older at package.json and notice behavior how now opacity starts changing when element is in top of viewport

**Expected behavior**
I would expect this to behave as in previous versions so I can pass elements' position and dimension info into useTransform
Maybe now I have to arrange my component differently in order to work? Please let me know

Thanks!

afroald:
I have the same issue, `useTransform` is not usable anymore since 2.3.0. There is another issue already: https://github.com/framer/motion/issues/513

PR #754, opened 2020-09-08
Ensure useTransform uses the latest handler

InventingWithMonster:
Fixes https://github.com/framer/motion/issues/752

I chose just to drop the dependency arrays as the resubscription cost is so low.

### DIFF
--- a/src/value/use-on-change.ts
+++ b/src/value/use-on-change.ts
@@ -6,15 +6,14 @@ export function useOnChange<T>(
     value: MotionValue<T> | number | string,
     callback: Subscriber<T>
 ) {
-    useEffect(
-        () => (isMotionValue(value) ? value.onChange(callback) : undefined),
-        [value]
+    useEffect(() =>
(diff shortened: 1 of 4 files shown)


src/value/use-on-change.ts
```diff
@@ -6,15 +6,14 @@ export function useOnChange<T>(
     value: MotionValue<T> | number | string,
     callback: Subscriber<T>
 ) {
-    useEffect(
-        () => (isMotionValue(value) ? value.onChange(callback) : undefined),
-        [value]
+    useEffect(() =>
+        isMotionValue(value) ? value.onChange(callback) : undefined
```
nvh, line 10: Is this on change returning the unsubscribe handler?

nvh (approved):

src/value/use-on-change.ts
InventingWithMonster: Yeah

PR #760, opened 2020-09-12
Let useDomEvent handle any EventTarget

dminkovsky:
Very neat little hook. This change allows passing in a ref `window` or `document`. `addDomEvent` is primed for it above. This change would fix the error on [this line](https://codesandbox.io/s/framer-motion-image-lightbox-47bhc?file=/src/Image.tsx:278-351). Any reason this is bad?

### DIFF
--- a/src/events/use-dom-event.ts
+++ b/src/events/use-dom-event.ts
@@ -34,7 +34,7 @@ export function addDomEvent(
  * @public
  */
 export function useDomEvent(
-    ref: RefObject<Element>,
+    ref: RefObject<EventTarget>,
     eventName: string,
     handler?: EventListener | undefined,
     options?: AddEventListenerOptions


dminkovsky:
Probably extreme, out of scope bike shedding here, but looking through lib.dom.d.ts, it's a shame I'm not seeing a way to make this safe/generic on various descendants of `EventTarget`, so that handler parameters get the correct event type.

Every `EventTarget` descendant has a method like

```
addEventListener<K extends keyof DocumentEventMap>(type: K, listener: (this: XMLDocument, ev: DocumentEventMap[K]) => any, options?: boolean | AddEventListenerOptions): void;
```

but there's no correlation between the event map type and the parent target type. So you could write something like

```
export default function useDomEvent<
    T extends EventTarget,
    M extends GlobalEventHandlersEventMap,
    K extends keyof M
>(
    ref: RefObject<T>,
    eventName: K,
    handler?: (this: T, ev: M[K]) => void,
): void {
    useEffect(() => {
        const element = ref.current;

        if (handler && element) {
            return addDomEvent(element, eventName, handler, options);
        }
    }, [ref, eventName, handler, options]);
}
```

but that is way intense.

InventingWithMonster:
You closed this PR but it’s a great change, I didn’t know about this type, you ok for me to merge?

dminkovsky:
Oh yeah definitely feel free to merge. I closed by accident somehow trying to delete a comment I accidentally posted. Wish this could be improved further somehow, but the types in lib.dom.d.ts don't seem to support it.

PR #761, opened 2020-09-15
Only setForcedRenderCount if component is mounted

### LINKED ISSUE #625: [BUG] Possible Memory Leak with AnimatePresence?

Nope, not one of those.

**2. Describe the bug**

I'm seeing evidence of a memory leak when using `AnimatePresence` in a container that leaves and reenters the DOM mid-animation. I think the scenario that causes the problem is something like this.

* Container mounts `AnimatePresence` along with component to show or hide.
* Show/hide animate starts.
* Before the animation completes, the container component itself leaves the view.
* React fires a warning: 

I use it in an interface where the user can quickly "Next" through a series of pages and it tends to pop up when we're testing a feature and hitting the button rapid-fire. Just now, we started seeing it while testing a form that modifies an animated component and occasionally interrupts it.

https://codesandbox.io/embed/framer-motion-examples-kx11r?fontsize=14&hidenavigation=1&theme=dark

This is a simple container that mounts an `AnimatePresence` component. The animated component has a long transition tied to `childVisible`. When `childVisible` becomes false, an event hook fires that removes the container before the animation has completed.

**4. Steps to reproduce**

1. Go to https://codesandbox.io/embed/framer-motion-examples-kx11r?fontsize=14&hidenavigation=1&theme=dark
2. Open up the in-window console to see React Warnings
3. Let the animation load, click "Hide it" and the animation will start but be interrupted by the removal of the container
4. "Hide It" will become "Restart It" -- try clicking these a few times at different stages of the animation and you'll eventually see the warning.

It seems to be dependent on timing and it doesn't happen all the time. I was able to reproduce it using this sandbox and so was a co-worker.

**5. Expected behavior**

I'd expect it to not try calling the internal `setState` on the unmounted component. It seems like something needs a `useEffect` that, on unmounting, stops the animation. Seems like I should be able to do this manually by passing controls from `useAnimation` but I'm not sure how to mix this with the existing `animate` options sent to each animated component. 

**6. Video or screenshots**

image

**7. Environment details**

We're seeing this on Chrome 83.

### `AnimatePresence` isn't working

Have all of its immediate children got a unique `key` prop that **remains the same for that component every render**?

```jsx
// Bad: The index could be given to a different component if the order of items changes
<AnimatePresence>
  {items.map((item, index) => <Component key={index} />)}
</AnimatePresence>
```

```jsx 
// Good: The item ID is unique to each component
<AnimatePresence>
  {items.map((item, index) => <Component key={item.id} />)}
</AnimatePresence>
```

Is the `AnimatePresence` correctly outside of the controlling conditional? `AnimatePresence` must be rendered whenever you expect an `exit` animation to run - it can't do so if it's unmounted!

```jsx
// Bad: AnimatePresence is unmounted - exit animations won't run
{isVisible && (
  <AnimatePresence>
    <Component />
  </AnimatePresence>
)}
```

```jsx
// Good: Only the children are unmounted - exit animations will run
<AnimatePresence>
  {isVisible && <Component />}
</AnimatePresence>
```

subvertallchris:
This appears to fix #625. I've tested it locally and it does the trick (🎉 ) but I'm unsure of how to add tests for it. I tested it with the component I put in the original codesandbox and ran my Cypress suite, which used to butt heads against this constantly and is now quite happy.

### DIFF
--- a/src/utils/use-force-update.ts
+++ b/src/utils/use-force-update.ts
@@ -1,9 +1,20 @@
-import { useState, useCallback } from "react"
+import { useState, useCallback, useRef, useEffect } from "react"
 
 export function useForceUpdate() {
+    const unloadingRef = useRef(false)
     const [forcedRenderCount, setForcedRenderCount] = useState(0)
 
-    return useCallback(() => setForcedRenderCount(forcedRenderCount + 1), [
-        forcedRenderCount,
-    ])
+    useEffect(() => {
+        return () => {
+            unloadingRef.current = true
+        }
+    }, [])
+
+    return useCallback(() => {
+        if (unloadingRef.current) {
+            return
+        }
+
+        setForcedRenderCount(forcedRenderCount + 1)
+    }, [forcedRenderCount])
 }


subvertallchris:
Fork of bug report sandbox with this commit: https://codesandbox.io/s/framer-motion-examples-forked-z9dzx?fontsize=14&hidenavigation=1&theme=dark

subvertallchris:
Sorry to be a pain, I know you guys are busy, but @InventingWithMonster, think you could take a look?

InventingWithMonster:
Yep, been on holiday, will take a look today

InventingWithMonster (approved):

InventingWithMonster:
Nice one thanks for the PR!

subvertallchris:
Would you consider backporting this and releasing as 1.11.x? (Not sure what version that’s up to.) I can’t upgrade to 2.6 yet due to #772 so this would be a big help. Happy to provide a PR for that, too.

PR #762, opened 2020-09-15
Create a reference to Animate's safeToRemove on VisualElement.config

hemlok:
In Framer, screens with VisualElements that have been assigned a `layoutId` in case a shared layout transition is performed, prevent screens that are removed with a different transition type from being fully removed from the DOM. This is because they've registered to the PresenceContext, and haven't called `safeToRemove()`. 

This isn't a problem for shared layout transitions, because we flush a callback to all children telling them to call `safeToRemove()` at the end of their animation, or immediately if they're not being animated.

Since framer uses `exit` transitions on containers for other transition types, rather than flushing a callback to every child of the exiting screen through the SharedLayout batcher (which would require us to perform some unnecessary dom manipulations and measurements on each outgoing component), this PR adds a reference to safeToRemove to the `visualElement` config. This is then called in framer on each exiting child without forking the batcher logic, or adding unnecessary overhead.

Fixes containers not being removed in Framer here: https://github.com/framer/FramerStudio/pull/6348

### DIFF
--- a/api/framer-motion.api.md
+++ b/api/framer-motion.api.md
@@ -843,7 +843,7 @@ export function useAnimation(): AnimationControls;
 export function useCycle<T>(...items: T[]): CycleState<T>;
 
 // @public
-export function useDomEvent(ref: RefObject<Element>, eventName: string, handler?: EventListener | undefined, options?: AddEventListenerOptions): void;
+export function useDomEvent(ref: RefObject<EventTarget>, eventName: string, handler?: EventListener | undefined, options?: AddEventListenerOptions): void;
 
 // @public
 export function useDragControls(): DragControls;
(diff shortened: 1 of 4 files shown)


api/framer-motion.api.md
```diff
@@ -843,7 +843,7 @@ export function useAnimation(): AnimationControls;
 export function useCycle<T>(...items: T[]): CycleState<T>;
 
 // @public
-export function useDomEvent(ref: RefObject<Element>, eventName: string, handler?: EventListener | undefined, options?: AddEventListenerOptions): void;
+export function useDomEvent(ref: RefObject<EventTarget>, eventName: string, handler?: EventListener | undefined, options?: AddEventListenerOptions): void;
```
nvh, line 846: Why did this change? Also the optional argument and undefined seem a bit double to me.

nvh (approved):

api/framer-motion.api.md
InventingWithMonster: @niels I think this is a rebuilt API from another external PR that changed the types to more accurately reflect valid arguments (like window and document)

InventingWithMonster (approved):

api/framer-motion.api.md
hemlok: Yep that's it! Threw me for a minute too!

PR #765, opened 2020-09-17
Fix: Don't unset keys with 0 values

hemlok:
- Animations from a current value to 0 were being unset because the target value was falsy
- Includes package.json bumps from releases that couldn't be pushed to master with my permissions

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "framer-motion",
-    "version": "2.6.13",
+    "version": "2.6.14-rc.0",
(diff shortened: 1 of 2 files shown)


hemlok:
This is the underlying bug behind: https://github.com/framer/company/issues/19676

nvh (approved):
Can these values be `null` at some point?

hemlok:
> Can these values be `null` at some point?

I don't know off of the top of my head, so I wanted to ensure that if that was the case for a reason I'm not aware of, that that logic doesn't get broken here.

PR #771, opened 2020-09-22
Duration-based springs

InventingWithMonster:
This PR adds duration-based springs to Motion.

Rather than using time-unbounded `stiffness` `mass` and `damping`, which are great for interactive springs, this allows springs to be defined using `duration` and/or `bounce` for timeline-compatible springs:

```
transition={{ type: "spring", duration: 0.7, bounce: 0.35 }}
```

`bounce` is a value between `0` (no overshoot) and `1` (extreme bounciness)

Demo: https://codesandbox.io/s/framer-motion-duration-based-springs-api-rc-z8fbu?file=/src/Example.tsx

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 Framer Motion adheres to [Semantic Versioning](http://semver.org/).
 
+## [2.7.0] 2020-09-22
+
+### Added
+
+-   Duration-based springs.
+
 ## [2.6.15] 2020-09-18
 
 ### Fix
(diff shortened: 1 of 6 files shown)


nvh:
Where's the pop motion PR that this is pointing to?

InventingWithMonster:
@nvh It's in an on-going `9` branch but the new file is here https://github.com/Popmotion/popmotion/blob/9/packages/popmotion/src/animations/utils/find-spring.ts

jonastreub (approved):
👏

PR #775, opened 2020-09-23
Removing support for legacy pointer events

InventingWithMonster:
Pointer events enjoy [full browser support](https://caniuse.com/pointer) so this PR removes support for mouse and touch fallbacks. This reduces the overall bundle size and reduces codebase complexity.


src/events/__tests__/types.test.tsx
```diff
@@ -1,9 +1,36 @@
 import "../../../jest.setup"
-import { Point } from "../types"
+import { Point2D } from "../../types/geometry"
+
+function relativeTo(idOrElem: string | HTMLElement) {
```
InventingWithMonster, line 4: This function was only used by tests, but as part of the old `Point` namespace couldn't be codesplit. So here I've refactored it out into the relative test.

api/framer-motion.api.md
```diff
@@ -224,9 +224,9 @@ export interface DraggableProps extends DragHandlers {
 // @public (undocumented)
 export interface DragHandlers {
     onDirectionLock?(axis: "x" | "y"): void;
-    onDrag?(event: MouseEvent | TouchEvent | PointerEvent, info: PanInfo): void;
```
InventingWithMonster, line 227: This IMO isn't a major API change as we used `PointerEvent`s as first choice and all major browsers support this.

InventingWithMonster:
Closing until iOS Safari 12 dies out.

image

PR #778, opened 2020-09-24
Removing Point namespace

InventingWithMonster:
The `Point` namespace only contains one method used in the codebase and can't be codesplit. This PR splits out the two functions into their respective points of usage and replaces the `Point` type with the other `Point2D` type used elsewhere in the codebase.

### DIFF
--- a/src/gestures/PanSession.ts
+++ b/src/gestures/PanSession.ts
@@ -1,11 +1,11 @@
-import { EventInfo, Point } from "../events/types"
+import { EventInfo } from "../events/types"
 import { isTouchEvent, isMouseEvent } from "./utils/event-type"
 import { extractEventInfo } from "../events/event-info"
 import sync, { getFrameData, cancelSync } from "framesync"
 import { secondsToMilliseconds } from "../utils/time-conversion"
 import { addPointerEvent } from "../events/use-pointer-event"
 import { distance, pipe } from "popmotion"
-import { TransformPoint2D } from "../types/geometry"
+import { Point2D, TransformPoint2D } from "../types/geometry"
 
 export type AnyPointerEvent = MouseEvent | TouchEvent | PointerEvent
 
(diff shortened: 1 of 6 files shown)


src/gestures/PanSession.ts
```diff
@@ -1,12 +1,12 @@
...
+import { EventInfo } from "../events/types"
 import { isTouchEvent, isMouseEvent } from "./utils/event-type"
 import { extractEventInfo } from "../events/event-info"
 import sync, { getFrameData, cancelSync } from "framesync"
 import { secondsToMilliseconds } from "../utils/time-conversion"
 import { addPointerEvent } from "../events/use-pointer-event"
 import { unblockViewportScroll } from "../gestures/drag/utils/block-viewport-scroll"
 import { distance } from "popmotion"
-import { TransformPoint2D } from "../types/geometry"
+import { Point2D, TransformPoint2D } from "../types/geometry"
```
nvh, line 9: This can be imported as types:
```suggestion
import type { Point2D, TransformPoint2D } from "../types/geometry"
```

src/gestures/PanSession.ts
```diff
@@ -308,16 +308,20 @@ export class PanSession {
 
 function transformPoint(
     info: EventInfo,
-    transformPagePoint?: (point: Point) => Point
+    transformPagePoint?: (point: Point2D) => Point2D
 ) {
     return transformPagePoint ? { point: transformPagePoint(info.point) } : info
 }
 
+function subtractPoint(a: Point2D, b: Point2D): Point2D {
```
nvh, line 316: I guess if it's single use it fits fine here, but I think it belongs in the geometry utils. (Don't feel strongly about this though)

nvh (approved):
Great thing to remove the duplication 👏 
I think we should import the `Point2D` as type everywhere

Issue #310, opened 2019-09-01
[BUG] [iOS13] whileTap doesn't animate back to initial props when interrupted with scrolling


gusso:
## Description
`whileTap` should animate back to initial element props when tap is over, but that doesn't happen when starting a scroll from the tap target.

## Environment
• iOS 13.1 beta (17A5821e) (also on iOS 13.0 builds)
• Safari

## Demo
https://codesandbox.io/s/runtime-night-f0o8i
![Kapture 2019-09-01 at 7 19 04](https://user-images.githubusercontent.com/782754/64072567-ed474400-cc88-11e9-858f-4bda6d9bc5a8.gif)

gusso:
Seems connected to https://github.com/framer/motion/issues/281, adding `touch-action: none` to the tap target stops the scrolling from happening and enforces the tap always finishes.

FuchsbergerTbs:
Is this a problem related to framer motion or is it a temporary bug in safari on iOS 13? On iOS 12 it works like a charm.
I'm asking because the `touch-action: none` solution doesn't cut it for me because I want to build a table-like layout where every cell is clickable, needless to say that `touch-action: none` doesn't work here because scrolling wouldn't work anymore.

GitIgnoreMaybe:
It doesn't work on Android either. Generally it would be great if those events e.g. whileTap are not triggered in the first place if ontouchmove is true.

IljaDaderko:
Happens on chrome (macOS) as well, setting `touch-action` to anything that is "pan" for an element that has `whileTap` seems to trigger this behaviour.

IljaDaderko:
Elements that have `whileTap` fire `onPointerDown` event, but after scrolling and releasing a finger they don't fire `onPointerUp` event, only when you click somewhere on the screen again.

IljaDaderko:
Something like this works as a workaround, but it is very involved i.e. you would need to dispatch "scrollEnded" event from your scrollable view to let all your components they need to animate back to normal position, ideally we still want `onPointerUp` event to fire:

```tsx
function Card({ onTap, children }: Props) {
  const animation = useAnimation();

  useSubscribe('scrollEnded', () => {
    animation.start({ scale: 1 });
  });

  return (
    <motion.div
      onPointerDown={() => {
        animation.start({ scale: 0.95 });
      }}
      onPointerUp={() => {
        animation.start({ scale: 1 });
      }}
      onTap={onTap}
      animate={animation}
    >
      {children}
    </motion.div>
[...]

Issue #185, opened 2019-07-11
[BUG] Stop scrolling from interfering with dragging


danprince:
**Is your feature request related to a problem? Please describe.**
If you render a list of `drag="x"` components and try to scroll down on a touch device, a small amount of horizontal drag happens on each element that your finger touches.

**Describe the solution you'd like**
A generic solution could be to allow the variant styles to override the styles from the drag.

```tsx
let variants = {
  // this would override whatever x value the drag gesture is providing
  static: { x: 0 },
  swiping: { },
};

let Swipeable = () => {
  let [isSwiping, setIsSwiping] = useState(0);

  let onDrag = (event, info) => setIsSwiping(info.offset.x > 10);

  return (
    <motion.div drag="x" animate={isSwiping ? "swiping" : "static"} onDrag={onDrag} />
  );
}
```

**Describe alternatives you've considered**
I also tried overriding by passing `transform` styles directly to the `motion.div` element.

A more specific version of this might be adding a `dragThreshold` prop that would specify the offsets that the gesture must pass before the `onDragStart` event is called.

Another option might be having some way to cancel the animation (but not the event) from the `onDrag` handler.

InventingWithMonster:
Are you finding that scrolling is happening at all when you drag on a component? It should be blocked - ideally we'd only block if you touched in the same direction as defined in `drag` but there's been issues getting this to work reliably.

You can combine `dragDirectionLock` with a drag axis so at least, for now, it shouldn't move if the user is swiping in the opposite direciton 

```jsx
<motion.div dragDirectionLock drag="x" />
```

danprince:
Thanks! That's at least a step in the right direction :+1:

I had been testing by scrolling in screen space below the list, but that's actually a much bigger issue if/when the list occupies the whole screen. Guessing there's not a way to unblock scrolling right now then?

InventingWithMonster:
Not at the moment sadly. I believe there might be a possible fix. But the problem ultimately is that to detect whether a user is dragging and in which direction we need to block the touch event from scrolling, or at least start blocking it before scrolling. But there's a distance a user can swipe within a single event that initiates scrolling. Once it's started, it's unblockable. Mobile browsers 👍

danprince:
What about allowing the user to control whether `blockViewportScroll()` is called, via a prop?

In my case, there's no horizontal scroll space available, so a drag on the x axis isn't going to affect it. Scrolling on the y axis shouldn't interfere because of the direction lock, right?

[...]

PR #779, opened 2020-09-24
Fix per-axis viewport scroll blocking

InventingWithMonster:
Fixes https://github.com/framer/motion/issues/185
Fixes https://github.com/framer/motion/issues/310

By using e.preventDefault after a event has started, it's impossible to block the browser from touch-scrolling. So we do it regardless of the dragging axis.

Currently if you, for instance, set drag to `"x"`, it's impossible to start dragging vertically from that element. Not helpful for carousels etc. By using the info in the `drag` prop we can set a CSS `touch-action` rule that blocks the browser from triggering scroll in the other direction.

QA
- Run make dev
- Go to Drag-block-viewport-conditionally
- Turn on mobile view 
- Start dragging vertically from the white box - page should scroll
- Start dragging horizontally from white box - box should drag


jonastreub (approved):
Fixes the issue 👏

jonastreub:
Finished the QA successfully as well.

Issue #288, opened 2019-08-21
[BUG] Propagation stopped only by constant not controls


dcecile:
**Describe the bug**
The [docs on propagation](https://www.framer.com/api/motion/animation/#propagation) say...

> If a `motion` component has children, changes in variant will flow down through the component hierarchy. These changes in variant will flow down until a child component defines its own `animate` property.

But defining an `animate` property that's set to animation controls doesn't stop propagation.

**To Reproduce**
1. Set up a parent with controls
2. Set up a child with controls (and same variant names)
3. Change the parent controls (`.start()`)
4. See the effect on the child

**Expected behavior**
Parent should not affect child because child has its own `animate` property.

**Screenshots**
See CodeSandbox: https://codesandbox.io/s/variants-tmgqc?module=%2Fsrc%2FExample.tsx

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser: Chrome
 - Version: 76

PR #780, opened 2020-09-24
Don't propagate variants if animate is AnimationControls

InventingWithMonster:
Fixes https://github.com/framer/motion/issues/288

This PR disables variant propagation if `animate` is an instance of `AnimationControls`.

There was an exception for `AnimationControls` but this doesn't seem to be covered by tests or have any explanation. I can't think of why you'd want this and it goes against the docs as quoted in the ticket.

As a compromise I've changed the behaviour of the `inherit` prop so now it can be used to either force a component to inherit or not, whereas before it could only disable inheritance.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 Framer Motion adheres to [Semantic Versioning](http://semver.org/).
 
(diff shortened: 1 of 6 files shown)


jonastreub (approved):
LGTM

Issue #364, opened 2019-10-15
[BUG] Not working on IE


mackan92:
The website my team are developing is using Framer Motion. It seems that it won't work on Internet Explorer. I've tested it with IE 11. Searching for bug reports lead me to nothing. 

When I tried to go in to your website on IE, [https://www.framer.com/motion/](https://www.framer.com/motion/), animations did not appear to be working either.

Has Framer Motion support for IE? If no, do you plan to implement it in a near future?

InventingWithMonster:
No support for IE11 now or ever I'm afraid. Time-permitting I hope to have a browser support list up in the docs, though it's basically all modern browsers.

msheakoski:
I wouldn't bother supporting IE11, it's a dead end. The new Chromium-based Edge supports Windows 7 and newer. I'm sure Microsoft will be pushing it heavily to existing IE users, in updates, notifications, and new installs of Windows. For corporate users, there is an "IE mode" built in for legacy sites.

thebuilder:
Seems like the issue on frame.com/motion is caused by arrow functions - Is the problem you are using any methods in the library that can't be polyfill?

msheakoski:
@thebuilder I had to support IE11 on a project that used Pose, so I imagine the polyfills needed are similar. Here is what made IE11 work:
```typescript
import "core-js/features/array/find";
import "core-js/features/object/assign";
import "core-js/features/promise";
import "core-js/features/string/ends-with";
import "core-js/features/symbol/for";
import "core-js/features/weak-set";
```

mackan92:
Thank you for your answers. My project also has to support IE. I added the polyfills and managed to make Framer Motion partly work on IE, but there were still a lot of issues. I will look at another option, probably remove animations completely on IE and keep Framer Motion for the modern browsers.

thebuilder:
@mackan92 any information about what/how you added the polyfills? What issues are you seeing? Broken animations, or script errors?
Are you loading the the ES version of Framer Motion, or the ES5 version?

https://unpkg.com/browse/framer-motion@1.6.14/dist/

Most likely Webpack is loading the `es` version and just uses that uncompiled.

thebuilder:
https://github.com/react-spring/react-spring/ has had the same issue with shipping ES6 bundle - They solved it by having you import the `.cjs` file instead. It's also

anilanar:
[...]

PR #781, opened 2020-09-24
IE11 support

InventingWithMonster:
Fixes https://github.com/framer/motion/issues/364

Some enterprise customers have a requirement to support IE11. With polyfills, everything in Motion is supportable with the exception of `Proxy`. Motion uses `Proxy` to generate `motion` components for all HTML, SVG, and hypothetical web component elements.

This PR adds a new function, `createDomMotionComponent`. This can be used to create DOM `motion` components manually.

```javascript
import { createDomMotionComponent } from "framer-motion"

const motion = {
  div: createDomMotionComponent("div")
}
```


tsconfig.json
```diff
@@ -11,7 +11,7 @@
         "emitDecoratorMetadata": true,
         "declaration": true,
         "declarationDir": "types",
-        "outDir": "tsc",
+        "outDir": "lib",
```
InventingWithMonster, line 14: This is a holdover from the previous approach but it also makes sense as this folder is transpiled, not TypeScript

jonastreub (approved):

magicspon:
Beeeeeelting...

PR #783, opened 2020-09-25
Fix: Manual `from` setting

InventingWithMonster:
This PR respects the `from` setting in `transition`, unblocking the Framer `borderRadius` ticket.


jonastreub (approved):
LGTM 👏

Issue #807, opened 2020-10-13
[BUG] layout animation stuttering when the parent component's state is updated


ttillberg:
**2. Describe the bug**

Updating the state of a component containing a `layout` element during a transition causes the animation to stutter.

Observations:

- Testing the bug against different versions of framer-motion it seems the issue appeared between 2.4.3 and 2.5.0.
- The bug doesn't occur with simple `animate` etc. only `layoutId` and `layout` elements 

**3. IMPORTANT: Provide a CodeSandbox reproduction of the bug**

https://codesandbox.io/s/vigilant-murdock-wri5r?file=/src/App.tsx

**4. Steps to reproduce**

1. Go to https://codesandbox.io/s/vigilant-murdock-wri5r?file=/src/App.tsx
2. Observe the animation is stuttering (pausing shortly every 250ms when the state gets updated )
3. Downgrade to framer-motion 2.4.3 to observe a fluid animation as expected

**5. Expected behavior**

Changing the state or re-rendering the parent component of a motion component shouldn't affect the animations of the latter

**6. Video or screenshots**

**7. Environment details**

Chrome 85, Safari 13, Firefox  OSX

mattgperry:
It looks like the logic around when we start a new layout animation changed around then. I'll take a look.

PR #821, opened 2020-10-20
Comparing origin/target before starting new layout animation

mattgperry:
Fixes https://github.com/framer/motion/issues/807

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,13 +2,19 @@
 
 Framer Motion adheres to [Semantic Versioning](http://semver.org/).
 
-## [2.9.3] 2020-10-19
+## [2.9.4] 2020-10-20
+
+### Fixed
+
+-   Only starting a new layout animation on an axis when an animation isn't in progress, the target has changed, and the origin is different to the currently-calculated viewport box.
+
+## [2.9.3] 2020-10-20
 
 ### Added
 
 -   Exporting `Target` type.
 
-## [2.9.2] 2020-10-19
+## [2.9.2] 2020-10-20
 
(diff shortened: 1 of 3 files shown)


hemlok (approved):
Makes sense 👌

johnnybenson:
Is this okay to merge?

rijk:
Here's another sandbox: https://codesandbox.io/s/condescending-keller-s5u28, as you can see when closing the modal it also keeps restarting the animation from 1 opacity, and the exit animation never finishes. Hope this change fixes that as well. In my app I have a similar modal with `layoutId` that because of a state change never reaches opacity: 1 but remains stuck on opacity: 0.962834. I tried to reproduce that here but couldn't unfortunately.

rijk:
@mattgperry sorry to spam, but this one is super annoying and comes up a lot. Basically every state change (loading data, measuring bounds, etc) that happens during a layout transition trips it up. There is not really a way to work around it, except by delaying the state updates, which in some cases (like a `ResizeObserver`) is not even possible. So it's kind of a show stopper for `layout` in those cases. If this fix is good, would it be possible for you to merge it?

ttillberg:
@mattgperry are you considering merging / using this fix in framer-motion 3 ? It's still a big problem for applications relying heavily on realtime state updates.

rijk:
I second that. It is occurring on almost every open/close transition I have in my app, making me have to work around it by e.g. delaying the loading of data.

Honason:
This one is important to me too. Without the fix I had to disable framer animations completely.

rijk:
Yes! 🥳🙏

PR #831, opened 2020-10-27
Enable framer magic motion transitions

hemlok:
- Add's a duplicate `SharedLayoutContext` called `FramerTreeContext`. This enables `Measure` to add the VisualElement to both the existing SharedLayoutContext (created as a batcher, or by any AnimateSharedLayout), but now also to `FramerTreeContext` which ensures that any motion element anywhere in the Framer tree still gets added to SharedLayoutTree's children map, so that they are flushed animate callbacks when there is a magic motion transition.

### DIFF
--- a/api/framer-motion.api.md
+++ b/api/framer-motion.api.md
@@ -260,6 +260,11 @@ export interface FeatureProps extends MotionProps {
 // @public (undocumented)
 export type ForwardRefComponent<T, P> = ForwardRefExoticComponent<PropsWithoutRef<P> & RefAttributes<T>>;
 
+// Warning: (ae-internal-missing-underscore) The name "FramerTreeContext" should be prefixed with an underscore because the declaration is marked as @internal
+// 
+// @internal (undocumented)
+export const FramerTreeContext: import("react").Context<SyncLayoutBatcher | SharedLayoutSyncMethods>;
+
 // @public (undocumented)
 export type GestureHandlers = PanHandlers & TapHandlers & HoverHandlers;
 
--- a/src/components/AnimateSharedLayout/SharedLayoutContext.ts
+++ b/src/components/AnimateSharedLayout/SharedLayoutContext.ts
@@ -101,6 +101,6 @@ export const SharedLayoutContext = createContext<
 /**
  * @internal
  */
-export const FramerTreeContext = createContext<
+export const FramerTreeLayoutContext = createContext<
     SyncLayoutBatcher | SharedLayoutSyncMethods
 >(createBatcher())
--- a/src/components/AnimateSharedLayout/SharedLayoutContext.ts
+++ b/src/components/AnimateSharedLayout/SharedLayoutContext.ts
@@ -97,3 +97,10 @@ export function isSharedLayout(
 export const SharedLayoutContext = createContext<
     SyncLayoutBatcher | SharedLayoutSyncMethods
 >(createBatcher())
+
+/**
+ * @internal
+ */
+export const FramerTreeContext = createContext<
+    SyncLayoutBatcher | SharedLayoutSyncMethods
+>(createBatcher())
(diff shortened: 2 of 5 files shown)


api/framer-motion.api.md
```diff
@@ -260,6 +260,11 @@ export interface FeatureProps extends MotionProps {
 // @public (undocumented)
 export type ForwardRefComponent<T, P> = ForwardRefExoticComponent<PropsWithoutRef<P> & RefAttributes<T>>;
 
+// Warning: (ae-internal-missing-underscore) The name "FramerTreeContext" should be prefixed with an underscore because the declaration is marked as @internal
+// 
+// @internal (undocumented)
+export const FramerTreeContext: import("react").Context<SyncLayoutBatcher | SharedLayoutSyncMethods>;
```
nvh, line 266: Should we prefix this with an underscore?

src/components/AnimateSharedLayout/SharedLayoutContext.ts
```diff
@@ -97,3 +97,10 @@ export function isSharedLayout(
 export const SharedLayoutContext = createContext<
     SyncLayoutBatcher | SharedLayoutSyncMethods
 >(createBatcher())
+
+/**
+ * @internal
+ */
+export const FramerTreeContext = createContext<
```
nvh, line 104: After seeing these again next to each other, I think `FramerTreeLayoutContext` might be a better name. WDYT @mattgperry ?

src/motion/features/layout/use-snapshot-on-unmount.ts
```diff
@@ -2,18 +2,24 @@ import { VisualElement } from "../../../render/VisualElement"
...
 import {
     SharedLayoutContext,
+    FramerTreeContext,
     isSharedLayout,
 } from "../../../components/AnimateSharedLayout/SharedLayoutContext"
 import { useIsomorphicLayoutEffect } from "../../../utils/use-isomorphic-effect"
 
 export function useSnapshotOnUnmount(visualElement: VisualElement) {
     const syncLayout = useContext(SharedLayoutContext)
+    const framerTreeSync = useContext(FramerTreeContext)
```
nvh, line 12: Maybe?
```suggestion
    const syncTreeLayout = useContext(FramerTreeContext)
```

src/motion/features/layout/Measure.tsx
```diff
@@ -6,10 +6,12 @@ import {
     SharedLayoutSyncMethods,
     isSharedLayout,
     SharedLayoutContext,
+    FramerTreeContext,
 } from "../../../components/AnimateSharedLayout/SharedLayoutContext"
 
 interface SyncProps extends FeatureProps {
     syncLayout: SharedLayoutSyncMethods | SyncLayoutBatcher
+    framerTreeSync: SharedLayoutSyncMethods | SyncLayoutBatcher
```
nvh, line 14:
```suggestion
    syncTreeLayout: SharedLayoutSyncMethods | SyncLayoutBatcher
```

nvh (approved):
Can we create tests for this?

src/motion/features/layout/Measure.tsx
mattgperry: I'd prefer this to mirror the context name and for it to be Framer-specific

mattgperry (approved):

api/framer-motion.api.md
hemlok: I don't have a strong opinion here but we have several other exports that are @internal that aren't prefixed with an underscore, so in the interest of keeping things consistent I would prefer not to add an underscore.

hemlok:
> Can we create tests for this?

@mattgperry are there any existing integration tests that I could build on top of that could easily confirm that both contexts where called where expected?

mattgperry:
@hemlok No we don't have tests around this yet :(

hemlok:
Cool, let's merge as is, and test in Framer 👌

PR #843, opened 2020-11-14
Fix parameter type of useTransform

### LINKED ISSUE #840: [BUG] Typescript error when using useTransform to combine 2 useTransforms
**2. Describe the bug**
I have 2 useTransforms (x, y) which I want to combine into a string to use as a CSS transform (not sure of what other way to do this — top/left isn't performant)
image

```
  const progress = useSpring(0, {spring});
  const x = useTransform(progress, [0, 1], [0, 100]);
  const y = useTransform(progress, [0, 1], [0, 100]);
  const t = useTransform([x, y], ([x:, y]) => `translate(${x}px, ${y}px)`);
```

But when I pass them into a 3rd useTransform to combine, typescript complains

```
TypeScript error in /Users/tarng/Code/playspace/src/av/Stream.tsx(21,26):
No overload matches this call.
  Overload 1 of 3, '(input: MotionValue<unknown>, transformer: SingleTransformer<unknown, unknown>): MotionValue<unknown>', gave the following error.
    Argument of type 'MotionValue<number>[]' is not assignable to parameter of type 'MotionValue<unknown>'.
      Type 'MotionValue<number>[]' is missing the following properties from type 'MotionValue<unknown>': current, prev, timeDelta, lastUpdated, and 19 more.  TS2769

    19 |   const x = useTransform(progress, [0, 1], [0, 100]);
    20 |   const y = useTransform(progress, [0, 1], [0, 100]);
  > 21 |   const t = useTransform([x, y], ([x, y]) => `translate(${x}px, ${y}px)`);
       |                          ^
```

if i try this on codepen without strict typescript, it throws this error:

image

https://codesandbox.io/s/framer-motion-2-animatesharedlayout-animate-between-different-components-forked-qmgro?file=/src/App.js

**4. Steps to reproduce**

Load the prototype, see typescript or type error

**5. Expected behavior**

It should run

**6. Video or screenshots**
N/A
**7. Environment details**
Chrome, macos

kena0ki:
Fixes #840 

I added `MotionValue<number>[]` and `MotionValue<string>[]` to the parameter type of third useTransform using union type.

 \* Changes about FramerTreeLayoutContext are resulted from previous commit.

### DIFF
--- a/src/value/use-transform.ts
+++ b/src/value/use-transform.ts
@@ -160,12 +160,19 @@ export function useTransform<I, O>(
  * @public
  */
 export function useTransform<I, O>(
-    input: MotionValue<string | number>[],
+    input:
+        | MotionValue<string>[]
+        | MotionValue<number>[]
+        | MotionValue<string | number>[],
     transformer: MultiTransformer<I, O>
 ): MotionValue<O>
 
 export function useTransform<I, O>(
-    input: MotionValue<I> | MotionValue<string | number>[],
+    input:
+        | MotionValue<I>
+        | MotionValue<string>[]
+        | MotionValue<number>[]
+        | MotionValue<string | number>[],
     inputRangeOrTransformer: InputRange | Transformer<I, O>,
     outputRange?: O[],
     options?: TransformOptions<O>
--- a/api/framer-motion.api.md
+++ b/api/framer-motion.api.md
@@ -260,10 +260,10 @@ export interface FeatureProps extends MotionProps {
 // @public (undocumented)
 export type ForwardRefComponent<T, P> = ForwardRefExoticComponent<PropsWithoutRef<P> & RefAttributes<T>>;
 
-// Warning: (ae-internal-missing-underscore) The name "FramerTreeContext" should be prefixed with an underscore because the declaration is marked as @internal
+// Warning: (ae-internal-missing-underscore) The name "FramerTreeLayoutContext" should be prefixed with an underscore because the declaration is marked as @internal
 // 
 // @internal (undocumented)
-export const FramerTreeContext: import("react").Context<SyncLayoutBatcher | SharedLayoutSyncMethods>;
+export const FramerTreeLayoutContext: import("react").Context<SyncLayoutBatcher | SharedLayoutSyncMethods>;
 
 // @public (undocumented)
 export type GestureHandlers = PanHandlers & TapHandlers & HoverHandlers;
@@ -911,7 +911,7 @@ export function useTransform<I, O>(input: MotionValue<I>, transformer: SingleTra
 // Warning: (ae-forgotten-export) The symbol "MultiTransformer" needs to be exported by the entry point index.d.ts
 // 
 // @public
-export function useTransform<I, O>(input: MotionValue<string | number>[], transformer: MultiTransformer<I, O>): MotionValue<O>;
+export function useTransform<I, O>(input: MotionValue<string>[] | MotionValue<number>[] | MotionValue<string | number>[], transformer: MultiTransformer<I, O>): MotionValue<O>;
 
 // @public
 export function useViewportScroll(): ScrollMotionValues;


o-alexandrov:
Instead of union of three different options:
- MotionValue<string>[]
- MotionValue<number>[]
- MotionValue<string | number>[]

You can just define only
- (MotionValue<string> | MotionValue<number>)[]

kena0ki:
@o-alexandrov 
Thanks for the review. But for your definition, following code results in error,
```
  const str = useMotionValue("0px");
  const num = useMotionValue(0);
  useTransform([str, num], () => 0); // should be OK
```

and similar test case already exists.
https://github.com/framer/motion/blob/dc571a6625dafa1d195d569db6b84c449fcc4140/src/value/__tests__/use-transform.test.tsx#L39-L50

o-alexandrov:
@kena0ki hello, thank you for the reply.
Are you sure you made the proposed adjustments?

I admit I typed the first message above from a phone not checking anything :D
But I just tested and have no issues, after making modifications to your code.

Typings work:
works

And if we try to cause an expected error, it throws:
expected-error

kena0ki:
@kena0ki
Are you sure `strict` or `strictFunctionTypes` in your tsconfig is true?
The issue only happens in strict mode.

kena0ki:
@o-alexandrov
Are you sure `strict` or `strictFunctionTypes` in your tsconfig is true?
The issue only happens in strict mode.

o-alexandrov:
Yeap, let me please share a screen recording as well.
Notes:
- did a window reload so there's no possibility of VSCode using old tsconfig settings
- showed .git changes to prevent a possibility of making it up

I don't have a permission to push to your branch, so I demonstrated via screenshots above and a video below.
[Screen Recording 2020-11-15 at 2.36.30 AM.mov.zip](https://github.com/framer/motion/files/5542029/Screen.Recording.2020-11-15.at.2.36.30.AM.mov.zip)

kena0ki:
@o-alexandrov
Sorry, you are right. The above code doesn't cause error.
The code results in error (and I tested) is following.
```
  const strOrNum = useMotionValue(true?"0px":0); // MotionValue<string | number>
  useTransform([strOrNum, strOrNum], () => 0); // should be OK (at least current definition)
```
If this case is ignorable, I agree your suggestion is better.

o-alexandrov:
Thank you for the clarification @kena0ki 
I also checked this case, for some reason under strict the case you highlighted in the last message doesn't pass.
Of course, I'm just a random fan of framer-motion, so it's not my decision to make, but I suggest:
- to accept the proposed solution in this MR for the typings

There are two options of defining those types then, either:
```ts
| MotionValue<string>[]
| MotionValue<number>[]
| MotionValue<string | number>[]
```
or:
```ts
(
| MotionValue<string>
| MotionValue<number>
| MotionValue<string | number>
)[]
```

src/value/use-transform.ts
```diff
@@ -160,12 +160,19 @@ export function useTransform<I, O>(
  * @public
  */
 export function useTransform<I, O>(
-    input: MotionValue<string | number>[],
+    input:
```
o-alexandrov, line 163: you might consider assigning this typings to a type variable and reuse it in the next function below

src/value/use-transform.ts
```diff
@@ -160,12 +160,19 @@ export function useTransform<I, O>(
...
+        | MotionValue<number>[]
+        | MotionValue<string | number>[],
     transformer: MultiTransformer<I, O>
 ): MotionValue<O>
 
 export function useTransform<I, O>(
-    input: MotionValue<I> | MotionValue<string | number>[],
+    input:
+        | MotionValue<I>
+        | MotionValue<string>[]
```
o-alexandrov, line 173: here

kena0ki:
Thank you for the suggestions. 
I think it depends on preference, so I want to leave the decision to maintainers. Until then, I'll leave the code as it is.
(Moreover another solution might be suggested, like just replacing to generics.)

Issue #846, opened 2020-11-15
[BUG] Can't install with latest version of React 17.0.1


joshbedo:
Trying to install framer motion with React 17.0.1 and can't. Tried using force command but still won't install because of dependency issues. Does this not work with the latest version of React? Doesn't seem like they changed anything just performance updates.

<img width="574" alt="Screen Shot 2020-11-15 at 6 32 08 PM" src="https://user-images.githubusercontent.com/607239/99200016-f084fd80-2770-11eb-9d45-bba65d2e5fbf.png">

Gerigot:
I think the propblem here is that in the peerDependencies the react version is written lik this ^16.8, that will not take the 17th or bigger in consideration.
I'll make a PR to correct this

Amasa-Delano:
It would be cool to get a new release of the package with this fix in it.

hafffe:
@mattgperry any reason why 2.9.5 isn't released to npm yet?

mattgperry:
Apologises for the delay, just publishing this now

PR #847, opened 2020-11-16
changed peerDependencies to support react 17

Gerigot:
As stated in the issue #846 it was impossible to use this library with react 17.

I've modified the peerDependencies to make it work

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,8 +35,8 @@
         "start-dev-server": "webpack-dev-server --config=dev/webpack/config.js"
     },
     "peerDependencies": {
-        "react": "^16.8",
-        "react-dom": "^16.8"
+        "react": ">=16.8",
+        "react-dom": ">=16.8"
     },
     "devDependencies": {
         "@cypress/webpack-preprocessor": "^4.1.0",
@@ -91,6 +91,7 @@
         "yarn-deduplicate": "^1.1.1"
     },
     "dependencies": {
+        "@emotion/is-prop-valid": "^0.8.2",
         "framesync": "^4.1.0",
         "hey-listen": "^1.0.8",
         "popmotion": "9.0.0-rc.20",


mattgperry:
Excellent thanks for this

package.json
```diff
@@ -91,6 +91,7 @@
         "yarn-deduplicate": "^1.1.1"
     },
     "dependencies": {
+        "@emotion/is-prop-valid": "^0.8.2",
```
o-alexandrov, line 94: @Gerigot 
Could you please clarify whether duplicating this from `optionalDependencies` was necessary?

o-alexandrov: this is probably a mistake and should be removed

Gerigot: Oh I see now I'm sorry yes it was a mistake probably done automatically by my IDE, if it's a problem I'll remove it.

o-alexandrov: I'm not using `preact` in projects w/ framer-motion and was just passing by :)
Since you probably have a working state w/ React 17, could you please test whether it still installs w/o this dep in `dependencies` and remove afterwards?

Gerigot: I'm an idiot. 
I was "fixing" (removing) the dependency and I screw the things up with git. so I've created a new pull request with the fix.

Amasa-Delano: I'm confused. I have a Next.js project that uses React 17. When I do `npm i framer-motion` I still get the peer dependency error. I thought the problem was fixed? Is there something I need to do to get the latest, fixed version of your library to install? I forced the install with `npm i framer-motion --force` but was hoping this wouldn't be necessary...thanks!

OlivierFortier: I have the same exact problem as [Amasa-Delano](https://github.com/framer/motion/pull/847#discussion_r531385531) .

Gerigot: I see that this pull request has been merged in the main branch 11 days ago, but no version of the library has been created since then.
The problem should be that  the last version on NPM doesn't have this fix. 
You'll have to wait till a version will be created or for now download and build it by yourself.
I'm not the owner of the project, so I don't know if is planned to create a new version and when will that happen.

Amasa-Delano: thanks for your quick response and help.

PR #857, opened 2020-11-17
3.0

mattgperry:
This PR introduces some new features and some breaking API changes to Motion.

# Added

## `whileDrag`

Like the existing `whileTap` and `whileHover`, `whileDrag` allows you to animate to a visual state while the drag gesture is active.

```jsx
<motion.div whileDrag={{ opacity: 0.8 }} />
```

## New drag CSS

Automatically adding Webkit-prefixed `user-select: none` and `touch-callout: none` CSS rules to draggable elements.

## Variant lists for `while-` props and `exit`

`whileHover`, `whileTap`, `whileDrag` and `exit` all now accept variant lists.

```jsx
<motion.div exit={["a", "b"]} />
```

# Removed

The following deprecated APIs have been completely removed:

- `useInvertedScale`. Use the `layout` prop instead.
- `yoyo`, `flip` and `loop` options. Use `repeat` and `repeatType` options instead.

# Changed

## Animation state

This is the major change of 3.0, which codifies and expands the existing behaviour of `whileHover` and `whileTap` to all animation props.

When a component's animation props (`animate`, `whileHover`, `exit` etc) change, or their active state (hover start/end) changes, it generates a new set of animations.

When a value no longer exists on any active prop, a new animation is created from its setting in `style`, or its originally-measured value from the DOM.

It also introduces the behaviour that when an animation prop is active, and the values in that prop change, we correctly handle that (whereas the current behaviour is to only animate when the active status of that prop changes). For example, if a hover gesture is active, and `whileHover` changes from `{ opacity: 0 }` to `{ opacity: 1 }`, that is now handled correctly.

Because this is largely a codification and expansion of existing functionality, it *shouldn't* cause any changes in the vast majority of projects. But as an explicit change to very old behaviour, it should be kept in mind when upgrading.


rijk:
It would be great if #821 could be merged in this PR as well.

PR #863, opened 2020-11-23
Reset only translate and scale when drag start

### LINKED ISSUE #833: [BUG] Motion element keeps gradually increasing in scale when dragged
Probably best to show in a sandbox:
https://codesandbox.io/s/gifted-dream-z5onb

When I click on the element with the shown configuration, the element keeps gradually increasing in scale as you tap the element (it's a small change, worth checking the inspector to see the effect better).

kena0ki:
Fixes #833 (and some type improvement around `HTMLVisualElement.transform`)

I changed to reset only translate and scale instead of whole transform when drag starts.
Translate and scale are recalculated using delta after drag starts, but rotate and skew are not recalculated so there is no need to be reset. And if they are reset, the size of the element used for delta value is wrongly measured.

### DIFF
--- a/api/framer-motion.api.md
+++ b/api/framer-motion.api.md
@@ -345,6 +345,7 @@ export class HTMLVisualElement<E extends HTMLElement | SVGElement = HTMLElement>
     rebaseTargetBox(force?: boolean, box?: AxisBox2D): void;
     removeValue(key: string): void;
     render(): void;
+    resetSomeTransform(props: TransformKey[]): void;
     resetTransform(): void;
     setAxisTarget(axis: "x" | "y", min: number, max: number): void;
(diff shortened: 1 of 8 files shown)


kena0ki:
Kapture 2020-11-23 at 18 57 01
https://codesandbox.io/s/wild-rgb-npcj3?file=/src/App.js

mattgperry:
Thanks for the PR but I am probably going to have to fix this one myself. These transforms are reset specifically so we can accurately measure the layout of the element. If we don't reset rotate and skew, this layout won't be correctly measured.

kena0ki:
Thank you for the comment.
I don't know how you fix with rotate and skew being reset but I hope this PR helps somehow.

PR #865, opened 2020-11-25
Fix SVG transforms

mattgperry:
Ensuring SVG transforms are applied after dimensions are measured

### DIFF
--- a/src/render/dom/use-render.ts
+++ b/src/render/dom/use-render.ts
@@ -7,7 +7,7 @@ import { buildSVGProps } from "./utils/build-svg-props"
 import { SVGVisualElement } from "./SVGVisualElement"
 import { isSVGComponent } from "./utils/is-svg-component"
 
-export function render<Props>(
+export function useRender<Props>(
     Component: string | React.ComponentType<Props>,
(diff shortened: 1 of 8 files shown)


src/render/dom/use-render.ts
```diff
@@ -7,7 +7,7 @@ import { buildSVGProps } from "./utils/build-svg-props"
 import { SVGVisualElement } from "./SVGVisualElement"
 import { isSVGComponent } from "./utils/is-svg-component"
 
-export function render<Props>(
+export function useRender<Props>(
```
nvh, line 10: Why the rename to useRender? It's not really a hook, right?

nvh (approved):
Don't have a lot of context here, but seems well-tested

src/render/dom/use-render.ts
mattgperry: Technically the one we have defined for the DOM isn't a hook in that it doesn't call any React hooks. But I want to make it explicit that it does adhere to the rules of hooks and therefore they could be called from it, either for us and this DOM `useRender` or an externally-defined one.

PR #869, opened 2020-11-27
Improving start-up performance

mattgperry:
Following the discovery of canvas work from MotionValue initialisation, this PR:

1. Lazy-initialises any MotionValue that was previously being created on page load
2. Directly applies the init value of a `MotionValue` to `current` rather than running the more expensive `updateAndNotify` (which also schedules a velocity check for the following frame)


hemlok (approved):
This looks good to me, but I would love if we could avoid merging this until we've successfully deployed the RC to production so we have one less new thing to qa/worry about!

PR #872, opened 2020-11-28
Add early return to snapToCursor not to update element axis when shouldn't

### LINKED ISSUE #855: [BUG] using "useDragControls" to "snapToCursor" does not respect the element's drag axis
**2. Describe the bug**
The useDragControls' snapToCursor flag should allow the targeted element to immediately move to the cursor.  Prior to framer-motion@2.4.0, the element would be positioned correctly on the specified drag axis. Post framer-motion@2.4.0, the element snaps to the cursor correctly, but the element's locked axis is not respected. 

For example, if the element has a drag="x", the element should only be able to move on the "x" axis. The issue here is that the element's "y" axis is also updated when snapToCursor is enabled.

https://codesandbox.io/s/framer-motion-usedragcontrols-forked-mv8qs?file=/src/App.js

**4. Steps to reproduce**
1. Go to https://codesandbox.io/s/framer-motion-usedragcontrols-forked-mv8qs?file=/src/App.js
2. Click just above the progress bar 
3. Button snaps to the cursor, but the button's "y" axis is also updated.

**5. Expected behavior**
The button should snap to the cursor based on the "x" axis. The "y" axis should remain unchanged.

**6. Video or screenshots**
Initial state:
image

After clicking just above the progress bar:
image

**7. Environment details**
Issue was introduced framer-motion@^2.4.0
Issue occurs on Chrome and Safari.

Providing the code JS code block from the codesandbox
```
export default function App() {
  const handleX = useMotionValue(0);
  const progressScaleX = useTransform(handleX, [0, 300], [0, 1]);
  const dragControls = useDragControls();

  return (
    <div
      className="slider-container"
      onMouseDown={(e) => dragControls.start(e, { snapToCursor: true })}
    >
      <div className="bar">
        <motion.div className="progress" style={{ scaleX: progressScaleX }} />
      </div>
      <motion.div
        className="handle"
        drag="x"
        dragMomentum={false}
        dragElastic={0}
        dragConstraints={{ left: 0, right: 300 }}
        dragControls={dragControls}
        style={{ x: handleX }}
      />
    </div>
  );
}

```

kena0ki:
Fixes #855

I added an early return to `snapToCursor` as `updateAxis` does.


mattgperry (approved):
Nice addition, thanks!

PR #875, opened 2020-12-02
Layout projection improvements

mattgperry:
This PR makes two improvements to layout projection:

## `withoutTransform`

This is a new method on `VisualElement` that will synchronously run a callback **after** unsetting the `transform` of an element and all of its ancestors and **before** restoring it.

This allows an element to be cleanly measured without the transform of any layout projected parent.

It's used here by `AnimateSharedLayout` to ensure a tree is measured only when its ancestors are reset, for instance those in another ASL tree or `LayoutTree` tree.

It's also used by `drag` components to ensure the current transform of a parent isn't taken into account for the layout measurement. This is currently an experimental feature but it means we can do things like resize a container as we drag an element out of it.

## Tree scale

This was currently being calculated by keeping a cumulative `treeScale` during `updateLayoutDeltas`. This has been changed to calculate a new tree scale from all of a component's ancestors in case the immediate parent isn't being projected.

### DIFF
--- a/src/utils/geometry/delta-apply.ts
+++ b/src/utils/geometry/delta-apply.ts
@@ -225,11 +225,17 @@ export function applyTreeDeltas(
     const treeLength = treePath.length
     if (!treeLength) return
 
+    // Reset the treeScale
     treeScale.x = treeScale.y = 1
+
     for (let i = 0; i < treeLength; i++) {
         const { delta } = treePath[i]
+
+        // Incoporate each ancestor's scale into a culmulative treeScale for this component
         treeScale.x *= delta.x.scale
         treeScale.y *= delta.y.scale
+
+        // Apply each ancestor's calculated delta into this component's recorded layout box
         applyBoxDelta(box, delta)
     }
 }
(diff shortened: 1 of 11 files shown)


hemlok:
> This has been changed to calculate a new tree scale from all of a component's ancestors in case the immediate parent isn't being projected.

src/utils/geometry/delta-apply.ts
```diff
@@ -217,9 +217,19 @@ export function removeBoxTransforms(
...
     const treeLength = treePath.length
+    if (!treeLength) return
+
+    treeScale.x = treeScale.y = 1
     for (let i = 0; i < treeLength; i++) {
-        applyBoxDelta(box, treePath[i].delta)
+        const { delta } = treePath[i]
+        treeScale.x *= delta.x.scale
+        treeScale.y *= delta.y.scale
+        applyBoxDelta(box, delta)
```
hemlok, line 233: Since we're mutating treeScale and not doing anything with it, I'm not sure I understood what was going on here until I read the tests, I think because it's a separate file, so maybe it could benefit from a comment?

hemlok (approved):
👍  LGTM

mattgperry:
> Does this mean that you can have motion layers now not tagged with layout between layers that are, and they will all still scale correct as expected? 🎉

Yeah that's right. This was actually the first way we did this but I tried to move it so we didn't have to loop through all the parents. It didn't really pan out though so only some stuff was optimised and we ended up with this bug.

PR #877, opened 2020-12-03
Adding support for SVG fallback

mattgperry:
This PR checks props for SVG fallback values so removing a value from `animate` behaves the same across SVG and HTML components.


hemlok (approved):
Nice 👍

Issue #828, opened 2020-10-22
[BUG] `rotate` no longer works after `layout` is added to an item


lintonye:
`rotate` stops working after `layout` prop is added and the layout is updated

CodeSandbox: https://codesandbox.io/s/purple-dawn-h1x1k?file=/src/App.js

** Steps to reproduce**

1. Click any card in the deck

** Expected behavior**

The cards should be arranged in a fan:

![image](https://user-images.githubusercontent.com/1065299/96907937-ca5d8d80-1450-11eb-84dc-a9eb49e96511.png)

** Actual behavior **

If applicable, add a video or screenshots to help explain the bug.
![image](https://user-images.githubusercontent.com/1065299/96907976-db0e0380-1450-11eb-9f18-1c851e8db1a0.png)

** Environment details**

framer-motion: 2.9.4

armandocin:
Hi guys!

Same issue with `translateX` animation

https://codesandbox.io/s/framer-motion-2-animating-shared-layouts-forked-oplj5

In this example removing `layout` from `motion.li` in `Content` component will make it work, but in other scenarios removing that attribute may cause distortions.

Please help!

lintonye:
@armandocin  You can just use `x` instead of `translateX`.

armandocin:
> @armandocin You can just use `x` instead of `translateX`.

I realize that just right before reading your comment. xD

It works, thanks for the reply!!

mattgperry:
This is fixed in `5.0.0-beta.1` - will leave this open until merged

PR #889, opened 2020-12-12
Support rotation in layout animations

mattgperry:
Fixes https://github.com/framer/motion/issues/828


hemlok (approved):
Works great! 👍 I really like that it's just quietly supported by ASL now instead of a prop too 😊

Issue #836, opened 2020-11-04
`whileFocus` gesture


stuartflanagan:
**Is your feature request related to a problem? Please describe.**
Kind of. A lot of designs I see have focus states similar to hover states. There is no focus gesture so we need to fallback to a non animated variant for focus. Or add more react state to handle focus if we want to reuse say a `whileHover` state animation

**Describe the solution you'd like**
A gesture `whileFocus` to be added to the possible gestures

**Describe alternatives you've considered**
CSS fallbacks. A lot of the time results in us not using motion for hover states so we can replicate them for focus states.

stuartflanagan:
👋

dahannes:
I would also love to see this feature along with `whileDisabled`

oielbanna:
Is anyone working on this? I've been needing this feature for a while now so Im down to work on it

dahannes:
@oielbanna I had some time to look into this today. Please see the above MR. 
1. `whileFocus` is pretty straight forward and works just fine
2. `whileDisable` has been implemented via MutationObserver to listen for changes of the disabled attribute.
It also works fine, but I did not manage to test this correctly, because `VisualElement.element` is undefined when run in tests.

FYI: @mattgperry

oielbanna:
@dahannes
I contacted Matt and he said I should go for it so I worked on whileFocus as well. I'm almost done, only missing the docs and examples. 
Did you branch off master? I think your PR might not be up to date with the latest changes. Would you like to work on this together?

oielbanna:
One thing I was thinking about while working on this is we could add a console.warn when users try to use this with elements that cant get focus or other related rules. For example <a> cant receive focus if it doesnt have href.

oielbanna:
@stuartflanagan https://github.com/framer/motion/pull/894 closes this feature request!

stuartflanagan:
Thanks a heap for this it will reduce a lot of duplication!

stuartflanagan:
Thank you for implementing this @oielbanna

oielbanna:
@stuartflanagan my pleasure :)

PR #892, opened 2020-12-14
Handle focus and disabled styles with framer motion

dahannes:
Closes https://github.com/framer/motion/issues/836

TODOs:
- Shim MutationObserver for tests and test whileDisabled correctly
Why is `visualElement.element` undefined when running tests?


mattgperry:
I'm going to close this PR for now as it's based of 2 and we now have `whileFocus` functionality. I like the idea of `whileDisabled` but we'll probably not add it just yet, I'd rather think of a way to let user-defined `while-` props going forward.

PR #913, opened 2020-12-30
Using pre-existing types in createDomMotionComponent to get the same proxied motion object types

LBRDan:
Hello there
I'm currently working in a javascript proxyless env. Searching inside the source code (as i cannot use proxy-polyfill) i noticed the `createDomMotionComponent` function being exposed.
Unfortunately, even as the function name suggests the creation of purely motion dom components, the resulting return types seems to be a little off compared to the one obtained via classic motion exposed object.
I took the existing types of motion, removing "typeof motion.custom" as it's doesn't fit with the wanted "createDom..." behavior

Please let me know if there is a better way to do this, or even if i'am missing something

### DIFF
--- a/src/render/dom/index.ts
+++ b/src/render/dom/index.ts
@@ -11,7 +11,7 @@ import { Animation } from "../../motion/features/animation"
 import { AnimateLayout } from "../../motion/features/layout/Animate"
 import { MeasureLayout } from "../../motion/features/layout/Measure"
 import { MotionFeature } from "../../motion/features/types"
-import { HTMLElements } from "./utils/supported-elements"
+import { HTMLElements, SVGElements } from "./utils/supported-elements"
 
 /**
  * I'd rather the return type of `custom` to be implicit but this throws
@@ -22,6 +22,8 @@ export type CustomDomComponent<Props> = React.ForwardRefExoticComponent<
     React.PropsWithoutRef<Props & MotionProps> &
         React.RefAttributes<SVGElement | HTMLElement>
 >
+
+type MotionComponents = HTMLMotionComponents & SVGMotionComponents
 
 const allMotionFeatures = [
     MeasureLayout,
@@ -52,10 +54,10 @@ const domBaseConfig = {
  * @public
  */
 export function createMotionProxy(defaultFeatures: MotionFeature[]) {
-    type CustomMotionComponent = { custom: typeof custom }
-    type Motion = HTMLMotionComponents &
-        SVGMotionComponents &
-        CustomMotionComponent
+    type CustomMotionComponent = {
+        custom: typeof custom
+    }
+    type Motion = MotionComponents & CustomMotionComponent
 
     const config: MotionComponentConfig<HTMLElement | SVGElement> = {
         ...domBaseConfig,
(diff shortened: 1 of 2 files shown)


src/render/dom/index.ts
```diff
@@ -104,9 +105,13 @@ export const motion = /*@__PURE__*/ createMotionProxy(allMotionFeatures)
  *
  * @public
  */
-export function createDomMotionComponent(key: string) {
+export function createDomMotionComponent(key: HTMLElements) {
```
mattgperry, line 108: This should remain `string` as it can be used to create custom HTML elements.

src/render/dom/index.ts
```diff
@@ -104,9 +105,13 @@ export const motion = /*@__PURE__*/ createMotionProxy(allMotionFeatures)
...
+    type motionTypes = HTMLMotionComponents & SVGMotionComponents
```
mattgperry, line 109: I think these types can be defined outside the function? And should start with a capital letter.

src/render/dom/index.ts
```diff
@@ -104,9 +105,13 @@ export const motion = /*@__PURE__*/ createMotionProxy(allMotionFeatures)
...
+    type motionTypesKeys = keyof motionTypes
+    type motionTypesValues = motionTypes[motionTypesKeys]
```
mattgperry, line 111: `MotionTypes[keyof MotionTypes]` is probably fine

mattgperry (changes requested):

mattgperry:
Hey @LBRDan, thanks for this PR! 

Have you got some examples of some types that are "off"? It would be good to write some tests that fail without your changes, for instance some test that simply renders a component using some of these broken types, so this doesn't break again in future.

src/render/dom/index.ts
LBRDan: Sure it is, changing this

src/render/dom/index.ts
LBRDan: Should I keep string only or just merging types like `HTMLElements | string` to give user some hints?

src/render/dom/index.ts
LBRDan: Sure it is, i tried to keep the same pattern for internal types you adopted in `createMotionProxy` by creating function scoped types
Where should I create these types ? `src/render/dom/types.ts` ?
Also seems good to have a single way for reference these types and adapt the types inside createMotionProxy (possibly)

LBRDan:
@mattgperry thanks for your time
`createMotionComponent` is a sort of forwardedRef HoC and gives back a Forwarded MotionComponent. It accepts two generics, that are not explicitly defined neither in `createDomMotionComponent` or `createMotionProxy`
In `createMotionProxy` however, the returned Proxy is forced `as Motion` that ties together HTMLElements key with HTMLMotionProps and ReactHTML attributes, consisting in a key-typed

Actually, I must admit that i don't know how to test this with a test framework, either because even createMotionProxy is used with hinting `as Motion` to Typescript...

Let me know if i can help in any other way

src/render/dom/index.ts
mattgperry: Er yeah you could merge and add SVGElements too

src/render/dom/index.ts
mattgperry: You can create them in this file, I think I only did that with createMotionproxy as the types themselves relied on incoming arguments (otherwise I messed up 😅)

mattgperry:
@LBRDan Because the tests are in typescript you can just do, well lets say for example a broken type was `onTap`.

So a test like this:

```
const Component = createDomMotionComponent('div')

render(<Component onTap={() => {}} />)

expect(true).toBe(true)
```

Should fail on `main` but pass with your changes, because TS would complain about the types. It would also help maintainers know which props were a problem.

mattgperry (approved):
Thanks for the changes!

mattgperry:
@LBRDan As a heads up I've had to remove ` as MotionComponents[keyof MotionComponents]` in [this PR](https://github.com/framer/motion/pull/924/files) as when running `make api` it added a ton of dynamic types to the generated API file. The test you provided in this PR is still passing so hopefully it was robust.

LBRDan:
Sorry @mattgperry i just saw this 
Maybe if you want to, I can include a subtype inside type files to avoid automatic dynamic type gen from api-extractor

Please let me know if I can help to work this out

mattgperry:
Ah thanks - I think I figured it out with this line: https://github.com/framer/motion/pull/924/files#diff-db6c742d25310f9a6dad488a97bf9a3265d4618ad5823df6474cd83513554a0aR109

PR #923, opened 2021-01-05
Feature/restore css prefixes

### LINKED ISSUE #920: [FEATURE] Restore CSS Vendor Prefix
Some time ago (not sure on which version came off), Motion used stylefire for css related stuff and element style assignment. One of the many perks of using stylefire was having a larger support on older platforms that are in need of vendorized/prefixed css properties (eg. transform->WebkitTransform stuff)

**Additional context**
Seems like some legacy platforms (using some patched webkit version) expose an unsupported unprefixed css property as supported.
Eg:
Some of the Samsung Smart Tv's (Orsay) Webkit build, acts like it supports (using a test dom element as a testing probe [stylefire way]) the plain `transform` property, but actually does ignore everything throwed in it and works fine using `-webkit-transform`. Even the good old stylefire fails at detecting this.

Restore this lost autoprefix behaviour inside motion

If you want, I already have a draft that I can submit via PR including this modification, but for the same reason described in the Additional Context paragraph, it fails to run some tests because the CSS property name picking function gives priority to prefixed one (surely supported) while the test make assertion about the exact style match
This can ben tedious to convert, also because the test environment uses a syntetic dom implementation via `jsdom` (using the .style implementation (inside `ElementCSSInlineStyle-impl.js`) based off `cssom` and `cssstyle`) therefore it is always pointing to the latest available living standard

LBRDan:
Hello there, this working draft resolves #920 (Please, read the issue for the concerns)

Not sure about lots of things:
- I kept `HTMLVisualElements` style separated on render method, calling prefixer on every render
In older version of motion, seems to be implemented directly inside of what is not called `buildHTMLStyles`, but i don't want to mess up something by touching something in there
Current client-tests still passes as they refer to `buildHTMLStyles` directly and other tests pass as jest DOM abstraction lib `jsdom` uses `cssom` and `cssstyle` (Refer back to #920) so prefixer does not find any prefixed property needs (Maybe, in the future, changing to headless browser test may be lead to client test failure as `prefixer` gives priority to vendorized (and certainly supported) css properties)
- The only test wrote is client-side, as old prefixer code uses dashCase only for SSR
This can lead to problems in some ways, as (eg) `Object.assign(this.element.style, {"transform-origin": "..."})` does not sound correct
- `prefixer` changes may be unsupported in some envs?
Using `key in testElement.style` may require additional iterator support that i suspect is missing from older browser/older ES implementation... sorry i cannot find something that says it does or it doesn't work

Please, let me know if this is possible to be done, and if so, if that can be done better than my garbage

### DIFF
--- a/src/render/dom/HTMLVisualElement.ts
+++ b/src/render/dom/HTMLVisualElement.ts
@@ -36,6 +36,7 @@ import {
 import { isMotionValue } from "../../value/utils/is-motion-value"
 import { MotionProps } from "../../motion"
 import { isCSSVariable } from "./utils/is-css-variable"
+import prefixer from "./utils/prefixer"
 
 export type LayoutUpdateHandler = (
     layout: AxisBox2D,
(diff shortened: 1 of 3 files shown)


LBRDan:
Actually, i am completely wrong
All the things that uses `HTMLVisualElement` cannot (as right as it is) pass the tests (for obvious reason)

PR #925, opened 2021-01-06
Add MotionContext to exports

jsfowles:
I am working on project that utilizes framer-motion, but I'm hitting an issue because this context is not exposed. Can we just export this variable publicly, or is there a reason for it to be left unexposed?

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -80,6 +80,7 @@ export {
 } from "./motion/context/MotionConfigContext"
 export { PresenceContext } from "./components/AnimatePresence/PresenceContext"
 export { LayoutGroupContext } from "./components/AnimateSharedLayout/LayoutGroupContext"
+export { MotionContext } from "./motion/context/MotionContext"
 
 /**
  * Types


mattgperry:
I mostly try not to expose internals unless we use them in Framer, what's the issue you've hit?

SimplyComplexable:
Hi @mattgperry  (any relation?) I'm working with Jake on the project this is for. Basically, we're building some primitives around the framer-motion spec that makes some of our most commonly used animations easier to implement. The reason we need to have access to this context is we're trying to default the behavior of "children" components if it doesn't have a parent, but also respect the parent variant if there _is_ a parent component wrapping it. This is more-or-less impossible without directly hooking into the `VariantContext`. 
We've tried implementing our own context, but it's impossible to match the inheritance happening within the actual context without creating wrappers for every motion component (and even then I'm not sure). Even if that route is possible, it would be a ton of overhead just to get around this context not being exposed. 
Honestly, this single variable is extremely useful for a variety of use-cases beyond this one. I can't think of any real risks, so I'm just not sure what the downside would be. If you felt more comfortable just exposing a `useVariantContext` so the values can't actually be changed just consumed we can update the PR to reflect that.

SimplyComplexable:
Actually just saw the latest commit which is exactly what I was hoping for. We'll close this.

PR #929, opened 2021-01-08
Adding correct default types for filter

mattgperry:
This PR adds `filter` type, which generates correct defaults for `brightness`, `contrast`, `saturation` and `opacity` filters, which should all be `1` or `100%` (depending on the input type) rather than `0`.

What this doesn't do, and will have to be fixed in Framer, is when we read from the DOM `filter` is always returned as numerical type as in `brightness(1)` vs `brightness(100%)`. This means if someone defines a `filter` in `style` as `brightness(100%)` and animates to `brightness(50%)` it'll actually animate from `brightness(1%)` coerced from the read value of `brightness(1)`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 Framer Motion adheres to [Semantic Versioning](http://semver.org/).
 
+## [3.2.0] Unreleased
+
+### Added
+
+-   Support for hex `alpha`, ie `#FFF0` and `#FFFFFF00`.
+-   Support for default `filter` values. For example, when animating from `brightness(50%)`, the animation will start from `brightness(100%)`rather than`brightness(0%)`.
+
 ## [3.1.5] 2020-01-08
 
 ### Fixed
--- a/dev/examples/Animation-filter.tsx
+++ b/dev/examples/Animation-filter.tsx
@@ -1,0 +1,24 @@
+import * as React from "react"
+import { motion } from "@framer"
+
+/**
+ * An example of animating the filter property.
+ */
+
+const style = {
+    width: 100,
+    height: 100,
+    background: "white",
+    opacity: 1,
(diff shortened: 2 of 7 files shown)


api/framer-motion.api.md
```diff
@@ -185,8 +185,11 @@ export interface BoxDelta {
...
...types").ForwardRefComponent<SVGPatternElement, import("./types").SVGMotionProps<SVGPatternElement>> | import("./types").ForwardRefComponent<SVGPolygonElement, import("./types").SVGMotionProps<SVGPolygonElement>> | import("./types").ForwardRefComponent<SVGPolylineElement, import("./types").SVGMotionProps<SVGPolylineElement>> | import("./types").ForwardRefComponent<SVGRadialGradientElement, import("./types").SVGMotionProps<SVGRadialGradientElement>> | import("./types").ForwardRefComponent<SVGRectElement, import("./types").SVGMotionProps<SVGRectElement>> | import("./types").ForwardRefComponent<SVGStopElement, import("./types").SVGMotionProps<SVGStopElement>> | import("./types").ForwardRefComponent<SVGSVGElement, import("./types").SVGMotionProps<SVGSVGElement>> | import("./types").ForwardRefComponent<SVGTextPathElement, import("./types").SVGMotionProps<SVGTextPathElement>> | import("./types").ForwardRefComponent<SVGTSpanElement, import("./types").SVGMotionProps<SVGTSpanElement>> | import("./types").ForwardRefComponent<SVGUseElement, import("./types").SVGMotionProps<SVGUseElement>> | import("./types").ForwardRefComponent<SVGViewElement, import("./types").SVGMotionProps<SVGViewElement>>;
```
mattgperry, line 192: I believe this is fixed in another PR

hemlok (approved):
Woooo! 🚀

PR #941, opened 2021-01-22
Use a ref with relative constraints for drag constraints

hemlok:
Using ref with relative drag constraints makes it easier to update relative drag constraints without triggering a react render, especially when the constraints have to be passed through props.

### DIFF
--- a/src/utils/is-bounding-box-ref.ts
+++ b/src/utils/is-bounding-box-ref.ts
@@ -1,0 +1,13 @@
+import { RefObject } from "react"
+import { BoundingBox2D } from "types/geometry"
+
+export const isBoundingBoxRefObject = (
+    ref: any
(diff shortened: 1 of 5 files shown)


src/utils/is-bounding-box-ref.ts
```diff
@@ -0,0 +1,13 @@
...
+import { BoundingBox2D } from "types/geometry"
+
+export const isBoundingBoxRefObject = (
+    ref: any
+): ref is RefObject<BoundingBox2D> => {
+    return (
+        typeof ref === "object" &&
+        ref.hasOwnProperty("current") &&
+        ref.current.hasOwnProperty("top") &&
+        ref.current.hasOwnProperty("left")
```
hemlok, line 11: Not suuuuuure if this is the best approach, if you can think of a better heuristic lmk

Issue #933, opened 2021-01-13
[BUG] Nested MotionConfigs with duplicate features throws error 


harisraharjo:
Hi,
I'm using AnimatePresence to enable page transition with react-router-dom.
However I encountered this error:
"Encountered two children with the same key, `animation`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version"

CodeSandbox:
https://codesandbox.io/s/loving-glitter-iy64g?file=/src/App.js

Steps to reproduce:

Go to the sandbox linked above.
Go to / 

It seems like this error is firing because of nested MotionConfig inside PageA. If I remove the outer MotionConfig and change the import from m to motion, the error goes away.

Environment details:
Using Chrome Version 87.0.4280.141 (Official Build) (64-bit) on Windows

mattgperry:
For me, later: The solution here is to dedupe `features` with the same key.

PR #949, opened 2021-01-28
Support duplicate features in nested configs

987poiuytrewq:
To [reduce the bundle size](https://www.framer.com/api/motion/guide-reduce-bundle-size/) (introduced in `2.2.0`), it would be helpful to be able to use multiple nested `MotionConfig`s. Then each one can import only the features that it needs. 

However there is an issue in doing this, detailed in framer/motion#933. Importing the same feature twice causes a key error.

Simply de-duping the list of features stops this from happening and allows you to nest configs.

Fixes https://github.com/framer/motion/issues/933

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 Framer Motion adheres to [Semantic Versioning](http://semver.org/).
 
+## [3.2.2] Unreleased
+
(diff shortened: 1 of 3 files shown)


mattgperry (approved):
Great work, thanks!

PR #957, opened 2021-02-02
Generate stagger order without re-render

mattgperry:
This removes forced re-renders via context to generate the correct stagger order

### DIFF
--- a/src/render/dom/html-visual-element.ts
+++ b/src/render/dom/html-visual-element.ts
@@ -20,6 +20,8 @@ export const htmlConfig: VisualElementConfig<
     HTMLMutableState,
     DOMVisualElementOptions
 > = {
+    treeType: "dom",
+
     readValueFromInstance(domElement, key) {
         if (isTransformProp(key)) {
             return getDefaultValueType(key)?.default || 0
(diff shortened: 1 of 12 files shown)


src/render/dom/html-visual-element.ts
```diff
@@ -40,6 +42,10 @@ export const htmlConfig: VisualElementConfig<
         vars: {},
     }),
 
+    sortNodePosition(a, b) {
+        return a.compareDocumentPosition(b) & 2 ? 1 : -1
```
hemlok, line 46: I wonder if a brief comment explaining this bitwise operation would be valuable? I read the spec and vaguely understand what's going on here but I will certainly forget 😅

src/render/dom/use-dom-visual-element.ts
```diff
@@ -86,27 +78,7 @@ export function useDomVisualElement<E>(
         )
     }
 
-    /**
-     * Create motion context for children to ensure children re-render if new variants
-     * are being passed. This is quite a blunt instrument and will lead to a ton of
-     * unneccessary rerenders. In the future we might prefer something like AnimateSharedLayout
-     * where incoming or re-rendering children trigger a rerender in the immediate
-     * variant parent.
-     */
```
hemlok, line 95: Yesssss 🔪 💀

hemlok (approved):
Code LGTM, but the code sandbox is failing 👀

hemlok:
I wonder if this will impact: https://github.com/framer/company/issues/20829 btw

mattgperry:
@hemlok Yeah this is a missing export from the branch we're trying to merge into.

I don't think this will affect the linked, but it will positively affect performance

Issue #190, opened 2019-07-12
[FEATURE] dragElastic bounds


sami616:
It would be great to be able to define how much 'elasticity' to apply to each `dragConstraint` edge using `dragElastic`

ie- `dragElastic={{ top: 0, bottom: 0.5 }}`

Trying to limit drag to only down in this demo: https://codesandbox.io/s/stoic-violet-h3bwn

InventingWithMonster:
Yeah I think this would be a cool feature. I'd accept a PR on this if anyone wants to give it a shot!

natemoo-re:
Rather than accepting a constraint-like object, could dragElastic accept a MotionValue? That could cover this use case and others. My use-case is to disallow overdrag past certain coordinates.

To that end, on a fork I was able to get dragElastic to read a MotionValue. But I ran into some glitchy behavior... Since a dragElastic of 0 restricts target movement to dragConstraints, setting dragElastic to 0 while a target is outside of its dragConstraints causes rapid position jumping.

Conceptually I guess I was thinking of dragElastic as a “dragResistance” type value rather than something spring-based.

@InventingWithMonster any thoughts on a path forward here? I’m happy to work on a PR!

vdjurdjevic:
Any news on this?

vdjurdjevic:
@InventingWithMonster I can submit pull request for this if you agree?

sami616:
Hi all, i wondered if there was a work around for this in the time being? Im building a drawer component which i want users to be able to drag to close, it would be great to disable dragging in a particular direction!

vdjurdjevic:
@sami616 The only workaround I found is to use react-spring :D I submitted a pull request for this, but got no response at all, not sure if it will be considered.

PR #549, opened 2020-05-14
Added support for constraint like elastic drag config

vdjurdjevic:
Fixes https://github.com/framer/motion/issues/190

PR #984, opened 2021-02-18
Adding dragElastic as an object

mattgperry:
Replaces https://github.com/framer/motion/pull/549
Fixes https://github.com/framer/motion/issues/190

Allows `dragElastic` to be defined as an object.

```
<motion.div drag dragElastic={{ top: 0.5 }} />
```


hemlok (approved):
LGTM 👍

PR #1002, opened 2021-02-24
Moving forwardMotionProps from props to config

mattgperry:
- Adding `forwardMotionProps` as a config option to `motion()`.
- Making `forwardMotionProps` default to `true` for the deprecated `motion.custom()`
- Moving `Component` through as a pre-bound option rather than through the `motion` render path.

### DIFF
--- a/api/framer-motion.api.md
+++ b/api/framer-motion.api.md
@@ -353,22 +353,22 @@ export interface LayoutProps {
 }
 
 // @public (undocumented)
-export const m: (<Props>(Component: string | import("react").ComponentClass<Props, any> | import("react").FunctionComponent<Props>) => import("./motion").CustomDomComponent<Props>) & import("./types").HTMLMotionComponents & import("./types").SVGMotionComponents & {
-    custom: <Props>(Component: string | import("react").ComponentClass<Props, any> | import("react").FunctionComponent<Props>) => import("./motion").CustomDomComponent<Props>;
+export const m: (<Props>(Component: string | import("react").ComponentClass<Props, any> | import("react").FunctionComponent<Props>, { forwardMotionProps }?: import("./motion").DomMotionComponentConfig) => import("./motion").CustomDomComponent<Props>) & import("./types").HTMLMotionComponents & import("./types").SVGMotionComponents & {
+    custom: <Props>(Component: string | import("react").ComponentClass<Props, any> | import("react").FunctionComponent<Props>, { forwardMotionProps }?: import("./motion").DomMotionComponentConfig) => import("./motion").CustomDomComponent<Props>;
 };
 
+// Warning: (ae-forgotten-export) The symbol "DomMotionComponentConfig" needs to be exported by the entry point index.d.ts
 // Warning: (ae-forgotten-export) The symbol "HTMLMotionComponents" needs to be exported by the entry point index.d.ts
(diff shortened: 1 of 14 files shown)


api/framer-motion.api.md
```diff
@@ -353,22 +353,22 @@ export interface LayoutProps {
...
 // Warning: (ae-forgotten-export) The symbol "SVGMotionComponents" needs to be exported by the entry point index.d.ts
 // 
 // @public
-export const motion: (<Props>(Component: string | React.ComponentClass<Props, any> | React.FunctionComponent<Props>) => CustomDomComponent<Props>) & HTMLMotionComponents & SVGMotionComponents & {
-    custom: <Props>(Component: string | React.ComponentClass<Props, any> | React.FunctionComponent<Props>) => CustomDomComponent<Props>;
+export const motion: (<Props>(Component: string | React.ComponentClass<Props, any> | React.FunctionComponent<Props>, { forwardMotionProps }?: DomMotionComponentConfig) => React.ForwardRefExoticComponent<React.PropsWithoutRef<Props & MotionProps> & React.RefAttributes<SVGElement | HTMLElement>>) & HTMLMotionComponents & SVGMotionComponents & {
+    custom: <Props>(Component: string | React.ComponentClass<Props, any> | React.FunctionComponent<Props>, { forwardMotionProps }?: DomMotionComponentConfig) => React.ForwardRefExoticComponent<React.PropsWithoutRef<Props & MotionProps> & React.RefAttributes<SVGElement | HTMLElement>>;
```
nvh, line 366: This suddenly became real noisy. Any way to prevent that?

src/motion/__tests__/custom.test.tsx
```diff
@@ -54,12 +54,33 @@ describe("motion()", () => {
...
+        expect(animate).toEqual({ x: 100 })
+        expect(foo).toBe(true)
+    })
+
+    test("forwards MotionProps if deprecated motion.custom used", () => {
+        let animate: any
+        let foo: boolean = false
+        const BaseComponent = React.forwardRef(
+            (props: Props, ref: RefObject<HTMLDivElement>) => {
+                animate = (props as any).animate
```
nvh, line 74: Do we need the any cast here? Can't we make it part of `Props`?

api/framer-motion.api.md
mattgperry: Yeah I'll take another look at it - some of this is the config but really it was from removing the explicit return.

src/motion/utils/use-visual-element.ts
```diff
@@ -27,7 +20,6 @@ type VisualElementRef = MutableRefObject<VisualElement | null>
 
 export function useVisualElement<E>(
     createVisualElement: CreateVisualElement<E>,
-    Component: string | ComponentType,
```
nvh, line 30: I don't really understand how we can remove this Component argument now?

mattgperry: This is now getting provided to these functions when we build the component, rather than threading it through with every individual render.

mattgperry: Specifically with this function, it's getting bound to the provided `createVisualElement` function, which is nice as this is meant to be the abstracted function that deals with platform-specific (in this case DOM) functionality

nvh (approved):

PR #1008, opened 2021-02-28
Fix tree-shaking under Webpack

frysztak:
Currently `framer-motion` doesn't tree-shake under Webpack. This is very noticable in component libraries such as Chakra-UI, where `framer-motion` is used very sparingly, but has a [large effect](https://github.com/chakra-ui/chakra-ui/issues/2693#issuecomment-758249931) on final bundle size.

In this minimal example:

```tsx
import { AnimationFeature, m as motion, MotionConfig } from 'framer-motion';

function App() {
  return (
     <MotionConfig features={[AnimationFeature]}>
       <motion.div
         className="container"
         initial={{ scale: 0 }}
         animate={{ rotate: 180, scale: 1 }}
         transition={{
           type: "spring",
           stiffness: 260,
           damping: 20
         }}
       />
     </MotionConfig>
  );
}
```

`framer-motion` takes 26.57 kB gzipped.

This PR does a couple of things:

1. Disables bundling entire library into a single ES module.
  I don't know about other bundlers, but Webpack [doesn't like](https://github.com/webpack/webpack/issues/9337#issuecomment-524966201) bundled libraries. 
  This effectively enabled tree-shaking and reduces bundle size to 20.05 kB gzipped.
  
2. Splits `motion.ts` into `motion-proxy.ts` . 
  Originally, `motion-minimal.ts` imported `motion.ts` in order to use `createMotionProxy`. Moving this function into a separate file allows Webpack to prune `motion.ts`.
  This reduces bundle size to 16.66 kB gzipped.
  
3. Moves `gestureProps` to `types.ts`.
  `gestureProps` were imported in `valid-prop.ts`, which inadvertly caused entire gesture feature to be bundled.
  This reduces bundle size to 14.91 kB gzipped.
  

In this minimal example, we saved a little over 11 kB gzipped.

### DIFF
--- a/api/framer-motion.api.md
+++ b/api/framer-motion.api.md
@@ -183,8 +183,11 @@ export function createDomMotionComponent<T extends keyof MotionComponents>(key: 
 // @internal
 export function createMotionComponent<P extends {}, E>({ defaultFeatures, createVisualElement, useRender, }: MotionComponentConfig<E>): React.ForwardRefExoticComponent<React.PropsWithoutRef<P & MotionProps> & React.RefAttributes<E>>;
 
-// @public
-export type CustomDomComponent<Props> = React.ForwardRefExoticComponent<React.PropsWithoutRef<Props & MotionProps> & React.RefAttributes<SVGElement | HTMLElement>>;
+// Warning: (ae-forgotten-export) The symbol "React" needs to be exported by the entry point index.d.ts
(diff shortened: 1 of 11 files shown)


mattgperry:
This is fantastic - I didn't understand why the tree-shaking wasn't working. Thanks for this PR, I'll try to merge and publish tomorrow.

PR #1009, opened 2021-03-01
Fixing perspective as style

mattgperry:
Currently, setting `style={{ perspective: 100 }}` outputs `transform: perspective(100px)`. This PR fixes it so it outputs `perspective: 100px`, and `transform` perspective must still be set as `transformPerspective`.

### DIFF
--- a/src/render/dom/utils/transform.ts
+++ b/src/render/dom/utils/transform.ts
@@ -8,17 +8,18 @@ export const transformAxes = ["", "X", "Y", "Z"]
  * An ordered array of each transformable value. By default, transform values
  * will be sorted to this order.
  */
-const order = ["perspective", "translate", "scale", "rotate", "skew"]
+const order = ["translate", "scale", "rotate", "skew"]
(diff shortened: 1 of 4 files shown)


hemlok (approved):
Yesssssss 🙏

src/render/dom/utils/transform.ts
```diff
@@ -8,17 +8,18 @@ export const transformAxes = ["", "X", "Y", "Z"]
...
  * will be sorted to this order.
  */
-const order = ["perspective", "translate", "scale", "rotate", "skew"]
+const order = ["translate", "scale", "rotate", "skew"]
 
 /**
  * Generate a list of every possible transform key.
  */
-export const transformProps = ["transformPerspective", "x", "y", "z"]
+export const transformProps = ["x", "y", "z"]
```
nvh, line 16: Why don't we create a set for this directly?

nvh (approved):

PR #1010, opened 2021-03-01
Fix path in motion-proxy

Evalon:
There is wrong path in motion-proxy that trigger typescript when library imported.


mattgperry:
Nice catch, thanks

PR #1011, opened 2021-03-02
Fixing crossfade for unused components

mattgperry:


src/components/AnimateSharedLayout/utils/crossfader.ts
```diff
@@ -148,7 +148,11 @@ export function createCrossfader(): Crossfader {
         stop: () => progress.stop(),
         getCrossfadeState(element) {
             updateCrossfade()
-            return element === options.lead ? leadState : followState
+            if (element === options.lead) {
+                return leadState
+            } else if (element === options.follow) {
+                return followState
+            }
```
hemlok, line 155: So this generally makes sense to me but I'm not sure how it effects the Framer bug? 

AFAIK AnimateLayoutTrees only triggers crossfades between the lead and follow so I'm confused how we could be trying to animate a element that isn't the follow?

mattgperry: Well this is more defensive than what we probably should do is ensure all non-follow components have no crossfader. But the gist is, if something that isn't follow/lead has reference to a crossfader and it renders for any reason it'll pull the follow's crossfaded state and render it

hemlok: Yeah I think I get it, let's get it in and test in framer 👌

hemlok (approved):

Issue #1003, opened 2021-02-24
[FEATURE] useVelocityTransform 


djfarly:
**Is your feature request related to a problem? Please describe.**
I would like to create a new MotionValue based on another MotionValues velocity.
Concrete use case: Apply a skew transform to images on scroll that is stronger the faster the user scrolls. (See condesandbox below)
This seems to doable in user space by targeting `getVelocity()` on the parent MotionValue. The problem is, that there never is a zero velocity call to the useTransform-transformer since the value itself is not changing anymore. Thus in my example if you scroll fast and stop abruptly - the images get stuck in a skewed position.

**Describe the solution you'd like**
An API like this would be great:
```ts
  const { scrollY } = useViewportScroll();
  const skewY = useVelocityTransform(scrollY, (scrollYVelocity) => {
    return scrollYVelocity * 0.00015;
  });
```

As an alternative MotionValues could expose their velocity as a MotionValue directly:
```ts
  const { scrollY } = useViewportScroll();
  const scrollYVelocity = scrollY.velocityMotionValueOrSomething; // 🤨
  const skewY = useTransform(scrollYVelocity, (scrollYVelocity) => {
    return scrollYVelocity * 0.00015;
  });
```

I tried to implement the first API in user space by replicating [the "next-next-frame velocity check" from the MotionValue internals](https://github.com/framer/motion/blob/main/src/value/index.ts#L302-L325). This works, but when setting the velocity to zero, it is quite jumpy. My guess is, that happens because it gains new velocity from being set to zero, thus retriggering the check.

```ts
function useVelocityTransform(
  val: MotionValue<number>,
  transformer: (velocity: number) => number,
) {
  const transformed = useTransform(val, () => transformer(val.getVelocity()));

  const handlePostRender = React.useCallback(() => {
    sync.postRender(handleClearFrame);
  }, []);

  const handleClearFrame = React.useCallback(
    ({ timestamp, delta }) => {
	  // if there has been no update since the last frame, assume velocity 0
      if ((transformed as any).lastUpdated - timestamp + delta !== 0) {
        transformed.set(0);
      }
    },
    [transformed],
  );

  React.useEffect(() =>
    val.onChange(() => {
      sync.postRender(handlePostRender);
    }),
  );

  return transformed;
}
```

Idk if this is very correct. It definitely feels quite hacky. ;)
One way to get rid of the jumpyness – and the way I am using this solution atm is to `useSpring` the resulting transformed value to basically smooth it out.

```ts
[...]

PR #1013, opened 2021-03-03
useVelocity

mattgperry:
This PR adds a `useVelocity` hook. It accepts a motion value and returns a new one that updates with its velocity.

```javascript
const x = useMotionValue(0)
const xVelocity = useVelocity(x)
const scale = useTransform(xVelocity, [0, 1000], [1, 2], { clamp: false })
```

Fixes https://github.com/framer/motion/issues/1003

Demo: https://codesandbox.io/s/framer-motion-310-usevelocity-nw1dn?file=/src/Example.tsx

### DIFF
--- a/src/value/index.ts
+++ b/src/value/index.ts
@@ -66,7 +66,7 @@ export class MotionValue<V = any> {
      *
      * @internal
      */
-    public velocityUpdateSubscribers = subscriptionManager<Subscriber<V>>()
+    public velocityUpdateSubscribers = subscriptionManager<Subscriber<number>>()
 
     /**
      * Functions to notify when the `MotionValue` updates and `render` is set to `true`.
@@ -111,7 +111,7 @@ export class MotionValue<V = any> {
      * @internal
      */
     constructor(init: V) {
-        this.current = init
+        this.prev = this.current = init
         this.canTrackVelocity = isFloat(this.current)
     }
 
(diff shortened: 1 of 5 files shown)


src/value/use-velocity.ts
```diff
@@ -0,0 +1,25 @@
...
+export function useVelocity(value: MotionValue<number>): MotionValue<number> {
+    const velocity = useMotionValue(value.getVelocity())
+
+    useEffect(() => {
+        return value.velocityUpdateSubscribers.add((newVelocity) => {
+            velocity.set(newVelocity)
+        })
+    }, [value])
+
+    return velocity
```
nvh, line 24: Nice that this is so simple 💪🏻

src/value/use-on-change.ts
```diff
@@ -6,9 +6,9 @@ export function useOnChange<T>(
     value: MotionValue<T> | number | string,
     callback: Subscriber<T>
 ) {
-    useEffect(() =>
-        isMotionValue(value) ? value.onChange(callback) : undefined
-    )
+    useEffect(() => {
+        if (isMotionValue(value)) return value.onChange(callback)
+    }, [callback])
```
nvh, line 11: Is this a driveby fix?

src/value/index.ts
```diff
@@ -61,6 +61,13 @@ export class MotionValue<V = any> {
      */
     private updateSubscribers = subscriptionManager<Subscriber<V>>()
 
+    /**
+     * Functions to notify when the velocity updates.
+     *
+     * @internal
+     */
+    public velocityUpdateSubscribers = subscriptionManager<Subscriber<number>>()
```
nvh, line 69: Why do we need a separate set of velocitySubscribers? Is that an optimisation because the velocity does not change as much?

src/value/use-velocity.ts
```diff
@@ -0,0 +1,25 @@
...
+ * ```
+ *
+ * @public
+ */
+export function useVelocity(value: MotionValue<number>): MotionValue<number> {
+    const velocity = useMotionValue(value.getVelocity())
+
+    useEffect(() => {
+        return value.velocityUpdateSubscribers.add((newVelocity) => {
+            velocity.set(newVelocity)
```
nvh, line 20: We could theoretically just subscribe to the value and then call `getVelocity` on that, right?

nvh (approved):

djfarly:
Tyvm @mattgperry – this api solves my use case perfectly and is much smarter that what I came up with. You work fast. 😬

src/value/use-velocity.ts
mattgperry: That was my preferred solution but there is a caveat that led me to putting it inside `MotionValue`. Every time a `MotionValue` is updated it schedules a velocity check at the end of the subsequent frame, so if it isn't updated in the meantime we will correctly report the velocity as `0` if requested.

It's this "if requested"  bit that's the problem. We'd have to duplicate that kind of scheduling here if we were to make sure the velocity is updated correctly.  By making explicit velocity subscribers we can ensure they get called at the same time velocity is updated, no matter the exact mechanism we use to perform that now or in the future.

src/value/use-on-change.ts
mattgperry: Sort of, it didn't look right to me but I also haven't heard of any breakages that I'm fixing.

src/value/use-velocity.ts
nvh: Makes sense 👍🏻

src/value/index.ts
mattgperry: Hopefully explained in the other comment

Issue #927, opened 2021-01-06
[BUG] "Dynamic start" doesn't seem to be supported in 3.x


ewanmellor:
The documentation at https://www.framer.com/api/motion/animation/ says:

> start can also accept a function that can dynamically start each component and the controls are bound to with a different animation definition.
> Custom data can be sent to this function via the component's custom prop.

This doesn't seem to be supported right now.  I found a line in `src/render/VisualElement/utils/animation.ts` that says `// TODO: Remove any and handle TargetResolver`, which I think is referring to this feature.

It seems to work in 2.9.5, and obviously it's documented.  Is this meant to be supported in 3.x?

KirdesMF:
Seems related to #888 

https://codesandbox.io/s/crimson-cache-y9j59?file=/src/App.tsx

jonathan-oberlander:
I also had an animation sequence break after updating to version 3.x. Here is simplified version. n is undefined, animation jumps straight to second step. This code is directly inspired from API documentation. Any thoughts?

```javascript
//
const animation = useAnimation()

useEffect(() => {
    const control = async () => {
      await animation.start((n) => ({
        y: -24,
        transition: {
          delay: n * 0.15,
        },
      }))
      await animation.start({
        opacity: 0,
        width: '45%',
        transition: {
          duration: 0.3,
        },
      })
    }
    control()
}, [anim])

return (
<>
  {['A', 'B', 'C', 'D'].map((letter, i) => (
   <motion.div
     custom={i}
     animate={animation}
     key={i}
   >
     {letter}
   </motion.div>
  )}
</>
)
```

Issue #888, opened 2020-12-11
[BUG] Dynamic controls does not fired


KirdesMF:
Dynamic Controls doesn't work, I tried with Framer and Framer motion
But with dynamics variants, it works
Maybe I'm missing something?

Codesandbox : https://codesandbox.io/s/gracious-shirley-6t03m?file=/src/App.tsx

1. Go to the codesanbox
2. Animation should works as expected
3. reverse the two useEffect.
4. Animation wont fired

jshikanova:
Something like this? https://codesandbox.io/s/pedantic-cache-kx0yb?file=/src/App.tsx
I'm not sure how to make it work with Framer tho

I think you can get similar kind of animation with stagger:
https://www.framer.com/api/motion/animation/#orchestration
https://codesandbox.io/s/framer-motion-stagger-fade-in-up-9gky2

KirdesMF:
the issue is not to have the same animation. The issue is that useAnimation with Framer does not work if you dont use a variants

https://codesandbox.io/s/gracious-shirley-6t03m?file=/src/App.tsx

When I tried with motion, I think I used the `useAnimation` from Framer and not framer-motion.

Anyway thanks for your help

PR #1014, opened 2021-03-03
Fixing dynamic controls.start

mattgperry:
Fixes https://github.com/framer/motion/issues/927
Fixes https://github.com/framer/motion/issues/888

### DIFF
--- a/src/render/utils/animation.ts
+++ b/src/render/utils/animation.ts
@@ -41,7 +41,7 @@ export type MakeTargetAnimatable = (
 export function animateVisualElement(
     visualElement: VisualElement,
     definition: AnimationDefinition,
-    options?: AnimationOptions
+    options: AnimationOptions = {}
 ) {
     visualElement.notifyAnimationStart()
     let animation: Promise<any>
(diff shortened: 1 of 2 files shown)


src/render/utils/animation.ts
```diff
@@ -54,8 +54,15 @@ export function animateVisualElement(
...
     } else {
-        // TODO: Remove any and handle TargetResolver
-        animation = animateTarget(visualElement, definition as any, options)
+        const resolvedDefinition =
+            typeof definition === "function"
+                ? (resolveVariant(
+                      visualElement,
+                      definition,
+                      options.custom
+                  ) as TargetAndTransition)
```
nvh, line 63: Why is this cast needed?

nvh (approved):

src/render/utils/animation.ts
mattgperry: We're hijacking a resolution function that could conceivably return `undefined` but in this instance it actually couldn't

nvh: I think we can resolve that by adding extra type overloads for this function

PR #1026, opened 2021-03-10
Reducing canvas memory usage

mattgperry:
This PR makes it possible for a `motion` component to render in static mode without instantiating a `VisualElement`, significantly reducing memory consumption.

A `VisualElement` is an abstraction around the renderer, by default the DOM. It enables us to performantly interact with the renderer and to communicate throughout the tree, away from React and its lifecycles. We create one for every `motion` component.

Static mode is a flag we can switch on in the Framer canvas and in server rendering where `motion` components behave more like normal elements in that a new `style` tag is created every render and no animations/gestures are mounted.

This PR changes the architecture slightly so that the initial map of values is created externally from the `VisualElement` and can itself either be used to render the `motion` component via React in static mode or to instantiate the `VisualElement` in normal mode.

This will save hundreds of megabytes of memory consumption from the Framer canvas in large prototypes. In a future refactor I'll change the `VisualElement` factory function to a `class` which will cut memory consumption in normal mode also.

This architecture also gets us pretty close to allowing the `VisualElement` itself to be lazy loaded, so we can bring the `m` component down from 13kb to something closer to 5.


dev/tests/svg.tsx
```diff
@@ -21,7 +21,7 @@ export const App = () => {
                 data-testid="rotate"
                 style={{ rotate: 45 }}
             />
-            <motion.rect
+            {/* <motion.rect
```
hemlok, line 24: This need to stay commented out?

PR #1030, opened 2021-03-14
Adding LazyMotion

mattgperry:
Fixes: https://github.com/framer/motion/issues/254

This PR adds a new `LazyMotion` component, that allows a subset of functionality to be loaded, sync or async. This enables the initial render to be just `4.8kb`.

## Sync

```jsx
<LazyMotion features={domAnimations} />
```

## Async

```jsx
<LazyMotion features={() => import("./path/to/feature").then(res => res.default)} />
```

## Notes

There's a lot here but the majority is just splitting out types from files to help code splitting in Webpack.


src/components/LazyMotion/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+        loadFeatures(loadedFeatures)
+    }
+
+    useEffect(() => {
+        if (isLazyBundle(features)) {
+            features().then(({ renderer, ...loadedFeatures }) => {
+                loadFeatures(loadedFeatures)
+                loadedRenderer.current = renderer
+                setIsLoaded(true)
+            })
```
hemlok, line 65: Any chance you need to catch here or is it up to the user to make sure that the features promise is handled accordingly?

hemlok: (I think it's fine if it's the latter)

src/motion/features/layout/Animate.tsx
```diff
@@ -61,6 +62,8 @@ class Animate extends React.Component<AnimateProps> {
         visualElement.enableLayoutProjection()
         this.unsubLayoutReady = visualElement.onLayoutUpdate(this.animate)
         visualElement.layoutSafeToRemove = () => this.safeToRemove()
+
+        addScaleCorrection(defaultScaleCorrectors)
```
hemlok, line 66: What's this change doing? Did we remove scale correction from somewhere else and are now adding it back to satisfy the bundlers? Hard to see in the big pr with lots of files moving around

src/motion/features/layout/Measure.tsx
```diff
@@ -28,6 +27,16 @@ class Measure extends React.Component<SyncProps> {
...
+
+        visualElement.onUnmount(() => {
+            if (isSharedLayout(syncLayout)) {
+                syncLayout.remove(visualElement as any)
+            }
+
+            if (isSharedLayout(framerSyncLayout)) {
+                framerSyncLayout.remove(visualElement as any)
+            }
+        })
```
hemlok, line 39: Same here, why is this change required?

src/motion/features/use-features.tsx
```diff
@@ -1,39 +1,62 @@
...
                     <Component
-                        key={key}
+                        key={name}
                         {...props}
-                        visualElement={visualElement as any}
+                        visualElement={visualElement}
                     />
                 )
+            } else {
+                // TODO: Lazy load
```
hemlok, line 57: Is this todo still valid? It looks to me like we are handling lazy loading already?

src/motion/features/layout/Animate.tsx
mattgperry: Basically this is dynamically loading scale correction with layout animations rather than having it synchronously loaded into the main bundle where it might never be used.

src/motion/features/layout/Measure.tsx
mattgperry: Same as above really, this ones just as bad really as these contexts call `createBatcher` by default. So it's not just a matter of bundling in empty contexts, the batcher also gets included in the minimal bundle.

src/motion/features/use-features.tsx
mattgperry: Oh yeah thanks, no it isn't. I had a thought that I'm going to leave for a later version if there's any demand, that we could provide a seperate endpoint for each feature which we'd then request here dynamically. So rather than loading during the first `useEffect` we could wait until a component wants a specific feature. But from my initial prodding this was generating a lot of duplication.

src/motion/features/layout/Measure.tsx
hemlok: Gotcha 👌

src/components/LazyMotion/index.tsx
mattgperry: Ah yeah this is actually why I didn't handle the loading in the initial iteration. I'll have a quick ponder.

hemlok (approved):

src/components/LazyMotion/index.tsx
mattgperry: Oh yeah I've had a play around and its probably better to figure this out within the dynamic function. A user could conceivably create a custom retry function here.

hemlok: Makes sense to me 👌

PR #1033, opened 2021-03-15
Adding support for isStatic in useMotionValue

mattgperry:


hemlok (approved):
LGTM 👍 

We will need to also update Page because it appears to be using the `motionValue` constructor instead of the hook: https://github.com/framer/FramerStudio/blob/2a4e4ce8511c658cf8dbe12629c3f309140b33c6/src/library/src/components/Page.tsx#L535-L538

I can take that if you want!

mattgperry:
@hemlok that'd be amazing if you could, I published this on `3.10.4`

Issue #1050, opened 2021-03-21
Having a build issue (Q/A)


curavcy:
Using Next,js to start, But cannot with these problem. Anyone recognize it before? 
Already deleted **package.lock.json** installed again. 
Using Node v14.15.5, Npm v6.14.11. 

```
Error: Package subpath './package.json' is not defined by "exports" in
    C:\Users\...\node_modules\framer-motion\package.json
    Code: ERR_PACKAGE_PATH_NOT_EXPORTED
```

mattgperry:
Hmm this sounds daft (because where else is it reading exports from in the first place?) but looking at this: https://github.com/react-native-community/cli/issues/1168 if something defines exports, as we do now, package.json needs to have a specific export for it to be importable. I’ll take a look at this later today.

mattgperry:
@curavcy can you try `4.0.1-rc.0` for me? I've attempted a fix

PR #1053, opened 2021-03-22
Adds package.json to exports field

mattgperry:
Fixes https://github.com/framer/motion/issues/1050

Adding `exports` to `package.json` actually breaks direct imports (ie `import from "framer-motion/package.json"`)

This PR adds a specific export for `package.json`. Seems backwards, but is apparently legit.


hemlok (approved):
😂

PR #1057, opened 2021-03-23
Changing layout projection from tree traversal to flat array

mattgperry:
Currently, all `VisualElement`s are kept in a tree structure. Once every frame, layout projection calculations are updated by iterating from the top `VisualElement` of every tree.

This kind of traversal is costly as it can involve many nested loops as the depth of the tree increases.

This PR changes the layout tree data structure to be a flat array. **Only** `VisualElement`s that have layout projection enabled are added to this array. The array is resorted by `VisualElement.depth` when an iteration is triggered and a `VisualElement` has been added/removed in the meantime.

Traversing a flat array brings its own performance benefits but this is more crucial with the coming changes to relative viewport box resolution, which will necessitate two-pass traversal.


api/framer-motion.api.md
```diff
@@ -964,6 +962,10 @@ export interface VisualElement<Instance = any, RenderState = any> extends Lifecy
     isVisible?: boolean;
     // (undocumented)
     layoutSafeToRemove?: () => void;
+    // Warning: (ae-forgotten-export) The symbol "FlatTree" needs to be exported by the entry point index.d.ts
```
nvh, line 965: Should we export FlatTree?

src/components/AnimateSharedLayout/utils/batcher.ts
```diff
@@ -18,7 +19,7 @@ export function createBatcher(): SyncLayoutBatcher {
     return {
         add: (child) => queue.add(child),
         flush: ({ measureLayout, layoutReady, parent } = defaultHandler) => {
-            const order = Array.from(queue).sort((a, b) => a.depth - b.depth)
+            const order = Array.from(queue).sort(sortByDepth)
```
nvh, line 22: This should be named `compareByDepth`

src/render/index.ts
```diff
@@ -846,9 +839,10 @@ export const visualElement = <Instance, MutableState, Options>({
             isEnabled && restoreTransform(instance, renderState)
         },
 
-        updateLayoutProjection() {
-            isProjecting() && updateLayoutProjection()
```
nvh, line 850: These calls were removed, is that intentional?

src/render/utils/flat-tree.ts
```diff
@@ -0,0 +1,35 @@
+import { addUniqueItem, removeItem } from "../../utils/array"
+import { sortByDepth } from "./sort-by-depth"
+
+export interface DepthChild {
```
nvh, line 4: This type does not have to be exported, and I think can be named better. In the FramerStudio codebase we have a pattern to prefix it with `With...` so something like `WithDepth` could work. Or `ChildWithDepth` if you want to be explicit.

src/render/utils/flat-tree.ts
```diff
@@ -0,0 +1,35 @@
...
+    depth: number
+}
+
+export class FlatTree {
+    children: DepthChild[] = []
```
nvh, line 9: This should probably be private

src/render/utils/sort-by-depth.ts
```diff
@@ -0,0 +1,4 @@
+import { VisualElement } from "../types"
+
+export const sortByDepth = (a: VisualElement, b: VisualElement) =>
```
nvh, line 3: Why is the type here a `VisualElement` and not a `DepthChild`?

src/render/utils/flat-tree.ts
```diff
@@ -0,0 +1,35 @@
...
+    numChildren = 0
```
nvh, line 10: It does not look like this is used anywhere else then inside the `forEach` I think it should just move to a local const there

src/render/utils/flat-tree.ts
```diff
@@ -0,0 +1,35 @@
...
+        this.isDirty = true
+    }
+
+    remove(child: DepthChild) {
+        removeItem(this.children, child)
+        this.isDirty = true
+    }
+
+    forEach(callback: (child: DepthChild) => void) {
+        this.isDirty && this.order()
```
nvh, line 25: Not checking the length every `forEach` call looks like a premature optimisation to me. We can remove the `order` function completely by doing:

```suggestion
       const numChildren = this.children.length
       this.isDirty && this.children.sort(compareByDepth)
```

src/render/utils/__tests__/flat-tree.ts
```diff
@@ -0,0 +1,13 @@
...
+
+describe("FlatTree", () => {
+    test("Correctly sorts by depth on iteration", () => {
+        const tree = new FlatTree()
+        tree.add({ depth: 1 })
+        tree.add({ depth: 0 })
+        expect(tree.children).toBe([{ depth: 1 }, { depth: 0 }])
+
+        tree.forEach(() => {})
+        expect(tree.children).toBe([{ depth: 0 }, { depth: 1 }])
```
nvh, line 11: I think it would be better here to actually check the calls to `forEach` instead of the side-effect of it

nvh (approved):
Added some comments for improvements, but looks good to me in general

src/render/index.ts
mattgperry: Yeah this does get called now by the `DepthTree.forEach`

api/framer-motion.api.md
mattgperry: Anything to make api-extractor happy 😅

PR #1060, opened 2021-03-24
Layout projection: Relative box resolution & Webpack updates

mattgperry:
This PR introduces a two-pass layout projection calculation, one to resolve relative projection boxes into viewport boxes, and then a second to calculate projection transforms as normal.

This PR also updates Webpack and related libraries to the latest versions. 

**Note:** We're only adding the mechanism to resolve relative bounding boxes. The next step is to go through and choose in which situations we use relative vs viewport


src/render/index.ts
```diff
@@ -842,7 +858,38 @@ export const visualElement = <Instance, MutableState, Options>({
...
+                    }
+                }
+
+                parentProjection = foundParentProjection
+            }
+
+            if (parentProjection) {
+                calcRelativeBox(projection, parentProjection)
+            } else {
+            }
```
hemlok, line 892: Can this else be removed?

src/render/utils/state.ts
```diff
@@ -18,6 +18,13 @@ export interface TargetProjection {
      */
     isTargetLocked: boolean
 
+    /**
+     * The parent-relative box we want to derive the viewport box from, if defined.
+     * This is currently all relative to the top/left of the parent box,
+     * but could be expanded in the future.
+     */
+    relativeTarget?: AxisBox2D
```
hemlok, line 26: I'm having a hard time seeing where this get's set?

Issue #1023, opened 2021-03-09
[BUG] 3.4.0 broke layout animations


rijk:
**2. Describe the bug**

From 3.4.0, layout animations have started behaving radically differently.

**3. IMPORTANT: Provide a CodeSandbox reproduction of the bug**

https://codesandbox.io/s/framer-motion-330-issue-with-layout-animations-iepm0?file=/src/App.js

**4. Steps to reproduce**

1. Click on one of the squares
2. Click again to close the modal
3. Observe the border radius animating back from 0 to 20, and the item disappearing only when the animation is fully done

Now upgrade framer-motion from 3.3.0 to 3.4.0:
1. Close the modal
2. Observe the border radius **not** animating back, and a flash when the animation is done and the modal is removed from DOM and the thumbnail is not yet visible

**5. Expected behavior**

A smooth transition like in 3.3.0 and before.

rijk:
Here's another very strange one, that so far I'm struggling to reproduce since it's wired deep into the app router etc. 

I've slowed down the transitions to be able to better see what's going on and made a recording. Let me know if you think this is the same issue, or would like a separate repro of it.

This is what happens in the recording:

1. I open modal by clicking square
2. Crossfade animation from square to modal
3. Close modal by clicking backdrop
4. Crossfade back to square
5. Open again
6. Refresh (the modal open state is stored in the URL, it should remain open after refreshing)
7. **Problem:** Note that initially the modal appears fine, and then suddenly a layout animation appears to kick in, making it fade out, while the step behind it still appears to be in its final crossfade position. This doesn't happen on 3.2.0.

https://user-images.githubusercontent.com/159500/110473238-9c66a600-80de-11eb-8d65-fd5ebe77bdc6.mov

rijk:
Hi @mattgperry , this is what I'm seeing in your forked Sandbox with 4.1.0 when I click the thumbnail:


https://user-images.githubusercontent.com/159500/113273257-c3c72200-92dc-11eb-8fa7-fde2309a87a0.mov

rijk:
After some trial and error, I see that it strongly depends on the type of transition:

**Works:**

```typescript
const transition = {
  type: "spring",
  stiffness: 400,
  damping: 50
};
```

**Also works:**

```typescript
const transition = {
  duration: 1
};
```

**Doesn't work:**

```typescript
const transition = {
  type: "spring",
  duration: 1
};
```

Issue #1022, opened 2021-03-09
[BUG] MotionConfig transition doesn't affect layout animations


rijk:
**2. Describe the bug**

When wrapping a tree in a `<MotionConfig>`, this has no effect on the automatic layout animations. Matt confirmed this to be a bug on Twitter: https://twitter.com/mattgperry/status/1366087239785144321

**3. IMPORTANT: Provide a CodeSandbox reproduction of the bug**

https://codesandbox.io/s/framer-motion-2-animating-shared-layouts-forked-3qu8z?file=/src/App.js

**4. Steps to reproduce**

Trigger a layout animation by clicking one of the elements.

**5. Expected behavior**

A layout animation with a duration of 5.

Issue #997, opened 2021-02-21
[BUG]


f0rr0:
Give a clear and concise description of what the bug is.

`AnimateSharedLayout` with custom durations does not animate back to initial state correctly. The initial layout is stuck at `opacity: 0`

A CodeSandbox minimal reproduction will allow us to quickly follow the reproduction steps. **Without one, this bug report won't be accepted.**

https://codesandbox.io/s/framer-motion-animatesharedlayout-app-store-demo-forked-6vm71?file=/src/Item.js

1. The linked codesandbox is from the official API docs. The sandbox has been updated to the latest release of framer and some custom durations have been added to `Item` motion components.
2. Click on a `Card` item and it animates to an expanded state correctly.
3. Click on it again or press back to go back to the original state.
4. The transition happens correctly but the original item is suddenly set to `opacity: 0` and is stuck there now.
5. The expansion animation still functions correctly.

`opacity` should not be set to 0 at the end of the transition.

Issue #721, opened 2020-08-21
[FEATURE] different transition for AnimateSharedLayout and mount animations


chrisdrackett:
In the beta of 2.0 I had a transition for `AnimateSharedLayout` that was different from the transition on my motion elements. This let me have one animation for the shared layout and another for mount. Just curious if this is still possible as it seems that `AnimateSharedLayout` no longer has a transition prop?

PR #1069, opened 2021-03-29
Layout projection: Animate relative projection boxes

mattgperry:
Fixes https://github.com/framer/motion/issues/1073
Fixes https://github.com/framer/motion/issues/1023
Fixes https://github.com/framer/motion/issues/1022
Fixes https://github.com/framer/motion/issues/997
Fixes https://github.com/framer/motion/issues/636
Fixes https://github.com/framer/motion/issues/721

Also adds `transition.crossfade` and `transition.layout` options.


dev/examples/AnimateSharedLayout-lists.tsx
```diff
@@ -24,7 +29,15 @@ const List = ({ list, onItemClick, backgroundColor }: ListProps) => {
...
                         onClick={() => onItemClick(id)}
-                        transition={{ duration: 3 }}
+                        transition={{
+                            default: {
+                                type: "spring",
+                                duration: 5,
+                            },
+                            crossfade: {
+                                duration: 0.4,
+                            },
```
hemlok, line 39: v cool

src/components/AnimateSharedLayout/utils/stack.ts
```diff
@@ -101,7 +101,7 @@ export function layoutStack(): LayoutStack {
             })
 
             if (
-                state.follow &&
+                state.lead !== prevState.follow &&
```
hemlok, line 104: Can you explain this change? I'm having a hard time understanding what it solves.

mattgperry: Yeah this is basically when we're animating the lead back to a follow, then when thats finished AnimatePresence rerenders, removes the lead, and makes the old follow the new lead. I'll add a comment tho

src/motion/features/layout/Animate.tsx
```diff
@@ -109,6 +113,38 @@ class Animate extends React.Component<AnimateProps> {
...
+        const projectionParent = visualElement.getProjectionParent()
+
+        if (projectionParent) {
+            let prevParentViewportBox = projectionParent.prevViewportBox
+            let parentLayout = projectionParent.getLayoutState().layout
+
+            if (prevParent) {
+                if (targetBox) {
+                    parentLayout = prevParent.getLayoutState().layout
+                }
```
hemlok, line 131: I think a comment explaining what's going on here might be helpful (for me 😅) down the road!

src/motion/features/layout/Animate.tsx
```diff
@@ -210,19 +245,18 @@ class Animate extends React.Component<AnimateProps> {
...
+                frameTarget.max,
+                isRelative
             )
         }
 
         // Synchronously run a frame to ensure there's no flash of the uncorrected bounding box.
         frame()
 
-        // Ensure that the layout delta is updated for this frame.
-        visualElement.updateLayoutProjection()
```
hemlok, line 224: It's not entirely clear to me why this isn't needed anymore?

hemlok: Is this because all children no longer have a delta since they're always relatively resolved?

mattgperry: The `flushSync` calls essentially do this now. This was a bit brute force anyway - it was a sync call that would update the entire tree below it. For each element calling it. Instead we do this at the end of the batcher, just once for the whole tree.

hemlok: Oh awesome

mattgperry: Oh and it was incompatible with the two-pass anyway, as it would schedule the second pass for the next frame step. Which is why we manually flush these out via framesync now. It ensures all trees first get relative boxes updated and then all projections updated, and then rendered, just once

hemlok (approved):
This looks great 🎉

src/motion/index.tsx
```diff
@@ -12,7 +12,6 @@ import { useMotionRef } from "./utils/use-motion-ref"
 import { useCreateMotionContext } from "../context/MotionContext/create"
 import { loadFeatures } from "./features/definitions"
 import { isBrowser } from "../utils/is-browser"
-export { MotionProps }
```
rijk, line 15: Would this change also influence imports like this?

```typescript
import { MotionProps } from 'framer-motion'
```

If so, it could be good to add a note about this to the changelog as well.

mattgperry: It shouldn’t, are you seeing any?

PR #1084, opened 2021-04-01
useMotionRef not updating callback

RuudBurger:
When using a callback as ref the useMotionRef callback is never updated because it isn't in the deps list.

```jsx
const MyComponent = ({ somethingThatChanges }) => {

  const callbackRef = useCallback(() => {
    // This only triggers once, even if `callbackRef` updates.
    console.log(somethingThatChanges);
  }, [somethingThatChanges])

  return <motion.div ref={callbackRef} ... />
}
```

### DIFF
--- a/src/motion/utils/use-motion-ref.ts
+++ b/src/motion/utils/use-motion-ref.ts
@@ -31,6 +31,6 @@ export function useMotionRef<Instance, RenderState>(
                 }
             }
         },
-        [visualElement]
+        [visualElement, visualState, externalRef]
     )
 }


RuudBurger:
@mattgperry Just gonna tag you, never got a response on my other PR as well. But this is a simple fix so should be able to quickly merge it.

thebuilder:
Didn't see this PR, but I've just made a similar change in #1119.

Not sure if the `visualState` needs to be included - seems to be an internal value? Might be the most correct approach to not break any rules of hooks.

mattgperry:
Thanks for this PR, I've merged the mentioned PR just because visualState is a constant between renders.

Issue #1088, opened 2021-04-07
[BUG] Transition from rem-value to 0rem using variants breaks starting with 4.1.1


souldreamer:
**1. Read the [FAQs](#faqs) 👇**

**2. Describe the bug**

When using variants to transition the height of an element from a non-zero rem value to 0rem (using the old way of actually specifying the unit for 0) we get an error. Strangely enough, it doesn't happen for 'em', 'vh', 'px' values. (see comments in App.tsx lines 7-14. Also doesn't happen if we don't specify the unit for 0. Though sometimes it just ignores the height completely and leaves it at 0.

**3. IMPORTANT: Provide a CodeSandbox reproduction of the bug**

[Reproduction at https://codesandbox.io/s/confident-meadow-kimzu](https://codesandbox.io/s/confident-meadow-kimzu)

**4. Steps to reproduce**

1. Go to codesandbox repro
2. Hover over the "Container" text
3. Mouse out from the div
4. Either see that "hi" text has no height (which is an error in and of itself), or repeat steps 2&3 a few times until you get an error overlay.

**5. Expected behavior**

Expected it to not break and interpret 0rem as 0 correctly. And always apply the variant's height correctly.

**6. Video or screenshots**

**7. Environment details**

Chrome
works correctly in v4.1.0, but breaks in 4.1.1 and 4.1.2

PR #1089, opened 2021-04-07
Animating 0 for all unit types

mattgperry:
Fixes https://github.com/framer/motion/issues/1088


src/animation/utils/transitions.ts
```diff
@@ -240,7 +238,9 @@ export function isZero(value: string | number) {
 export function getZeroUnit(
     potentialUnitType: string | number
 ): string | number {
-    return getValueAsType(0, findDimensionValueType(potentialUnitType))
+    return typeof potentialUnitType === "number"
+        ? 0
+        : getAnimatableNone("", potentialUnitType)
```
hemlok, line 243: So the trick was to use less code? 😅

mattgperry: Haha yeah it turned out nicely

hemlok (approved):

PR #1095, opened 2021-04-15
Disable hover on drag

mattgperry:
This PR disables hover events while a drag event is active. In Framer this will ensure we don't trigger hover state changes while a drag gesture is active.


nvh (approved):

tpict:
Is it possible to opt out of this behavior on the element being hovered? Since this change was merged, you can't have an element animate in response to the user dragging something into it (without re-implementing `whileHover`, anyway)

PR #1096, opened 2021-04-16
Relative projection fixes

mattgperry:
This PR introduces two fixes to layout projection.

1. It is more defensive about when it performs a layout animation relatively. If either an `originBox` or `targetBox` is explicitly defined **without** reference to a different parent component, we revert to viewport projection. This ensures we don't calculate a relative position to the wrong component.
2. The layout animation batcher in `AnimateSharedLayout` schedules a job to overwrite the `prevViewportBox` snapshot with the live `projection` data. This ensures that any nested `AnimateSharedLayout` components attempting to calculate a relative position aren't doing so against an out-of-date snapshot.

### DIFF
--- a/src/components/AnimateSharedLayout/utils/batcher.ts
+++ b/src/components/AnimateSharedLayout/utils/batcher.ts
@@ -1,4 +1,4 @@
-import { flushSync } from "framesync"
+import sync, { flushSync } from "framesync"
 import { VisualElement } from "../../../render/types"
 import { compareByDepth } from "../../../render/utils/compare-by-depth"
 import { Presence, SyncLayoutBatcher, SyncLayoutLifecycles } from "../types"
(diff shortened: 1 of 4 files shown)


src/components/AnimateSharedLayout/utils/batcher.ts
```diff
@@ -62,7 +62,21 @@ export function createBatcher(): SyncLayoutBatcher {
...
 
+            /**
+             * Schedule a callback at the end of the following frame to assign the latest projection
+             * box to the prevViewportBox snapshot. Once global batching is in place this could be run
+             * synchronously. But for now it ensures that if any nested `AnimateSharedLayout` top-level
+             * child attempts to calculate its previous relative position against a prevViewportBox
+             * it will be against its latest projection box instead, as the snapshot is useless beyond this
+             * render.
+             */
+            sync.postRender(() => order.forEach(assignProjectionToSnapshot))
```
nvh, line 73: i there a test for the bug that this fixes? i'm afraid it will return in some shape or form once we do the global batching

src/motion/features/layout/Animate.tsx
```diff
@@ -153,9 +153,11 @@ class Animate extends React.Component<AnimateProps> {
             }
 
             if (prevParentViewportBox) {
-                isRelative = true
-                origin = calcRelativeOffset(prevParentViewportBox, origin)
-                target = calcRelativeOffset(parentLayout, target)
+                if (prevParent || (!prevParent && !(originBox || targetBox))) {
```
nvh, line 156: Can we extract these in a well named constant that explains what we're checking for?

nvh (approved):

src/components/AnimateSharedLayout/utils/batcher.ts
mattgperry: I'll have to revisit this with a test as I think it'll be quite difficult to implement and we want to ship this out asap

Issue #956, opened 2021-02-02
[BUG] Multiple drag fails to have correct positionning in v3


arelaxend:
Dear contributors,

**2. Describe the bug**

When there are several drags imbricated drag, the position of the parent drag impacts the position of child imbricated drags, whenever it is `layout` or  `AnimateSharedLayout` or not.

**3. IMPORTANT: Provide a CodeSandbox reproduction of the bug**

Please check the following codeSandbox, https://codesandbox.io/s/framer-motion-drag-with-constraints-s5qiq.

**4. Steps to reproduce**

Use framer version 3.2.1.

1. Drag the parent `drag`area
2. Then drag one of the child
3. Depending on the position of the parent drag, the child drag position change.

**5. Expected behavior**

For version below v3, for exemple 2.9.5 the behavior looks correct. That is, the child drag position remains.

A.

PR #1103, opened 2021-04-19
Nested drag fixes

mattgperry:
Fixes https://github.com/framer/motion/issues/956

This PR adds a robust testing environment around, and fixes, nested drag. 

There are three remaining skipped tests, these are for when we're dragging one layer on the `"x"`  axis and a child on the `"y"` simultaneously **and** one or the other is layout-projected. I think this is quite a rare case compared to the amount of fixes this introduces so I'd rather look next at integrating this into `Scroll` first.

### DIFF
--- a/src/components/AnimateSharedLayout/utils/batcher.ts
+++ b/src/components/AnimateSharedLayout/utils/batcher.ts
@@ -1,4 +1,8 @@
 import sync, { flushSync } from "framesync"
+import {
+    batchResetAndMeasure,
+    withoutTreeTransform,
+} from "../../../render/dom/projection/utils"
 import { VisualElement } from "../../../render/types"
 import { compareByDepth } from "../../../render/utils/compare-by-depth"
 import { Presence, SyncLayoutBatcher, SyncLayoutLifecycles } from "../types"
@@ -7,7 +11,6 @@ import { Presence, SyncLayoutBatcher, SyncLayoutLifecycles } from "../types"
  * Default handlers for batching VisualElements
  */
 const defaultHandler: SyncLayoutLifecycles = {
-    measureLayout: (child) => child.updateLayoutMeasurement(),
     layoutReady: (child) => child.notifyLayoutReady(),
 }
 
@@ -19,24 +22,14 @@ export function createBatcher(): SyncLayoutBatcher {
 
     return {
         add: (child) => queue.add(child),
-        flush: ({ measureLayout, layoutReady, parent } = defaultHandler) => {
+        flush: ({ layoutReady, parent } = defaultHandler) => {
             const order = Array.from(queue).sort(compareByDepth)
 
-            const resetAndMeasure = () => {
-                /**
-                 * Write: Reset any transforms on children elements so we can read their actual layout
-                 */
-                order.forEach((child) => child.resetTransform())
-
-                /**
-                 * Read: Measure the actual layout
-                 */
-                order.forEach(measureLayout)
-            }
-
             parent
-                ? parent.withoutTransform(resetAndMeasure)
-                : resetAndMeasure()
+                ? withoutTreeTransform(parent, () => {
+                      batchResetAndMeasure(order)
+                  })
+                : batchResetAndMeasure(order)
 
             /**
              * Write: Notify the VisualElements they're ready for further write operations.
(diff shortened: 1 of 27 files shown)


src/render/utils/flat-tree.ts
```diff
@@ -18,10 +18,7 @@ export class FlatTree {
 
     forEach(callback: (child: WithDepth) => void) {
         this.isDirty && this.children.sort(compareByDepth)
-
-        const numChildren = this.children.length
-        for (let i = 0; i < numChildren; i++) {
-            callback(this.children[i])
-        }
+        this.isDirty = false
```
mattgperry, line 21: 🤦‍♂️

src/components/AnimateSharedLayout/utils/batcher.ts
```diff
@@ -19,24 +22,14 @@ export function createBatcher(): SyncLayoutBatcher {
...
-                order.forEach(measureLayout)
-            }
-
             parent
-                ? parent.withoutTransform(resetAndMeasure)
-                : resetAndMeasure()
+                ? withoutTreeTransform(parent, () => {
+                      batchResetAndMeasure(order)
+                  })
+                : batchResetAndMeasure(order)
```
nvh, line 32: Why is this a ternary if we don't use the return value?
And even if the return value was used, I think it would be much clearer to just have an if-statement, because of the function calls and callbacks.

src/gestures/drag/VisualElementDragControls.ts
```diff
@@ -395,27 +385,19 @@ export class VisualElementDragControls {
...
-
-        if (dragMomentum || this.elastic) {
-            const { velocity } = info
-            this.animateDragEnd(velocity)
-        }
+        const { velocity } = info
+        this.animateDragEnd(velocity)
 
-        onDragEnd?.(event, info)
+        this.props.onDragEnd?.(event, info)
```
nvh, line 391: Why not extract this from the props path the top of this function?

nvh: Never mind, I did not see the full context and that this is the only access to `this.props` that we do

src/render/dom/projection/convert-to-relative.ts
```diff
@@ -0,0 +1,41 @@
+import { calcRelativeOffset } from "../../../motion/features/layout/utils"
+import { AxisBox2D } from "../../../types/geometry"
+import { eachAxis } from "../../../utils/each-axis"
+import { removeBoxTransforms } from "../../../utils/geometry/delta-apply"
+import { VisualElement } from "../../types"
+
+export function convertToRelativeProjection(
```
nvh, line 7: Can we add a comment what the meaning of the return type of this function is?

src/render/dom/projection/relative-set.ts
```diff
@@ -0,0 +1,24 @@
+import { calcRelativeOffset } from "../../../motion/features/layout/utils"
+import { eachAxis } from "../../../utils/each-axis"
+import { VisualElement } from "../../types"
+
+export function setCurrentViewportBox(visualElement: VisualElement) {
+    const projectionParent = visualElement.getProjectionParent()
+
+    if (!projectionParent) {
+        visualElement.rebaseProjectionTarget()
+    } else {
```
nvh, line 10: I think this would read better as an early return

src/render/index.ts
```diff
@@ -693,40 +692,30 @@ export const visualElement = <Instance, MutableState, Options>({
...
 
         setCrossfader(newCrossfader) {
             crossfader = newCrossfader
         },
 
+        isProjecting: () => projection.isEnabled && layoutState.isHydrated,
+
         /**
          * Start a layout animation on a given axis.
-         * TODO: This could be better.
```
nvh, line 717: 🥇

src/render/utils/flat-tree.ts
nvh: 😅

src/render/dom/projection/utils.ts
```diff
@@ -0,0 +1,96 @@
...
+ */
+export function withoutTreeTransform(
+    visualElement: VisualElement,
+    callback: () => void
+) {
+    const { parent } = visualElement
+    const { isEnabled } = visualElement.projection
+    isEnabled && visualElement.resetTransform()
+
+    parent ? withoutTreeTransform(parent, callback) : callback()
```
nvh, line 47: Same comment about ternary without return usages here.

nvh (approved):
I'm not going to pretend I fully understand what's going on, but it's well tested. Left some minor style remarks.

PR #1106, opened 2021-04-21
Fixing layout projection tests

mattgperry:
This fixes the tests that the nested drag PR broke.

### DIFF
--- a/cypress.json
+++ b/cypress.json
@@ -1,4 +1,4 @@
 {
     "baseUrl": "http://0.0.0.0:9990",
-    "video": false
+    "video": true
 }
--- a/cypress/integration/drag-svg.ts
+++ b/cypress/integration/drag-svg.ts
@@ -154,8 +154,16 @@ describe("Drag SVG & Layout", () => {
     it("Drags the element by the defined distance", () => {
         cy.visit("?test=drag-svg&layout=true")
             .get("[data-testid='draggable']")
-            .wait(200)
-            .trigger("pointerdown", 0, 0, { force: true })
+            .wait(50)
+            .should(($draggable: any) => {
+                const draggable = $draggable[0] as HTMLDivElement
+                const { left, top } = draggable.getBoundingClientRect()
+
+                expect(left).to.equal(30)
+                expect(top).to.equal(30)
+            })
+            .wait(50)
+            .trigger("pointerdown", 50, 50, { force: true })
             .trigger("pointermove", 60, 60, { force: true }) // Gesture will start from first move past threshold
             .wait(50)
             .trigger("pointermove", 200, 300, { force: true })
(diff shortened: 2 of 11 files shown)


cypress.json
```diff
@@ -1,4 +1,4 @@
 {
     "baseUrl": "http://0.0.0.0:9990",
-    "video": false
+    "video": true
```
nvh, line 3: Should this still be enabled?

src/motion/features/layout/Measure.tsx
```diff
@@ -61,11 +60,9 @@ class Measure extends React.Component<SyncProps> {
     }
 
     componentDidUpdate() {
-        const { syncLayout, visualElement } = this.props
+        const { syncLayout } = this.props
 
         if (!isSharedLayout(syncLayout)) syncLayout.flush()
-
-        setCurrentViewportBox(visualElement)
```
nvh, line 68: Why is this no longer needed?

src/render/index.ts
```diff
@@ -808,6 +809,7 @@ export const visualElement = <Instance, MutableState, Options>({
          * needs to be performed.
          */
         notifyLayoutReady(config) {
+            setCurrentViewportBox(element)
```
nvh, line 812: Ah, did it move to here?

cypress/integration/drag-svg.ts
```diff
@@ -154,8 +154,16 @@ describe("Drag SVG & Layout", () => {
...
+            .wait(50)
+            .should(($draggable: any) => {
+                const draggable = $draggable[0] as HTMLDivElement
+                const { left, top } = draggable.getBoundingClientRect()
+
+                expect(left).to.equal(30)
+                expect(top).to.equal(30)
+            })
+            .wait(50)
+            .trigger("pointerdown", 50, 50, { force: true })
```
nvh, line 166: Did this test just change significantly which is why all the numbers changed?

nvh (approved):

cypress.json
mattgperry: Yeah I use it a lot for debugging and I don't notice a speed difference when running the tests so it's usually just a pain to keep changing this setting

src/render/index.ts
mattgperry: Yeah that's right. Specifically what this changes is we now only run this when the element's layout is definietly ready to be used in projection calculations.

cypress/integration/drag-svg.ts
mattgperry: So I've manually verified that these tests are working the same but it's more that we've changed that the pointer's position on an element gets calculated on pointer down, rather than when the pan initiates, the avoid pointer slippage.

PR #1108, opened 2021-04-22
Nested drag fixes: Meme edition

mattgperry:
This adds tests for, and fixes, children of nested drag elements

### DIFF
--- a/cypress/integration/drag-nested.ts
+++ b/cypress/integration/drag-nested.ts
@@ -159,7 +159,7 @@ function testNestedDrag(parentLayout: boolean, childLayout: boolean) {
 describe("Nested drag", () => {
     it("Parent: layout, Child: layout", () => testNestedDrag(true, true))
     it("Parent: layout", () => testNestedDrag(true, false))
-    it("Child: layout", () => testNestedDrag(false, true))
+    it.skip("Child: layout", () => testNestedDrag(false, true))
     it("Neither", () => testNestedDrag(false, false))
 })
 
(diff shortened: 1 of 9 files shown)


src/utils/geometry/delta-apply.ts
```diff
@@ -242,7 +242,7 @@ export function applyTreeDeltas(
         applyBoxDelta(box, delta)
 
         // If this is a draggable ancestor, also incorporate the node's transform to the layout box
-        if (!isRelative && node.getProps().drag) {
+        if (node.getProps().drag) {
```
mattgperry, line 245: This was only applying drag transforms to children that were being projected absolutely. So children were jumping as they weren't being transformed in the same way.

cypress/integration/drag-nested.ts
```diff
@@ -5,12 +5,12 @@ interface BoundingBox {
...
 
-function expectBbox(element: HTMLElement, expectedBbox: BoundingBox) {
+function expectBbox(element: HTMLElement, expectedBbox: Partial<BoundingBox>) {
     const bbox = element.getBoundingClientRect()
     expect(bbox.left).to.equal(expectedBbox.left)
     expect(bbox.top).to.equal(expectedBbox.top)
-    expect(bbox.width).to.equal(expectedBbox.width)
-    expect(bbox.height).to.equal(expectedBbox.height)
+    expectedBbox.width && expect(bbox.width).to.equal(expectedBbox.width)
+    expectedBbox.height && expect(bbox.height).to.equal(expectedBbox.height)
```
hemlok, line 13: maybe this is a good candidate for non-null assertions?

hemlok (approved):
Make sense 👌

PR #1111, opened 2021-04-23
Fixing jumping children in nested Page/Scroll

mattgperry:


src/render/utils/is-draggable.ts
```diff
@@ -0,0 +1,6 @@
+import { VisualElement } from "../types"
+
+export function isDraggable(visualElement: VisualElement) {
+    const { drag, _dragX } = visualElement.getProps()
+    return drag && !_dragX
```
nvh, line 5: What does the _dragX mean?

nvh (approved):

src/render/utils/is-draggable.ts
mattgperry: This is the external `_dragX` and `_dragY` prop - if we're outputting drag to those we can consider this component not really draggable

mattgperry: (We always use both hence the single check)

PR #1118, opened 2021-04-26
Allowing individual layers to reset transforms during layout measurements

mattgperry:
This allows us to reset the transform of a component so its children can be measured without its effects.


src/components/AnimateSharedLayout/index.tsx
```diff
@@ -152,7 +152,10 @@ export class AnimateSharedLayout extends React.Component<
         /**
          * Write: Reset rotation transforms so bounding boxes can be accurately measured.
```
nvh, line 153: This comment should be updated

nvh (approved):

Issue #742, opened 2020-08-28
[BUG] Input Fields are not behaving as expected when used with react-hook-form


Elia-d:
**2. Describe the bug**

When using ```motion.input``` with react-hook-form, input fields are getting unregistered after resetting the form.

As you would see in the CodeSandbox provided below, using either normal input fields or using an input field with ```React.forwardRef``` both work as expected with no problems.

However, when using ```motion.input```, react-hook-form does not recognize the input fields after resetting the form (even when I tell it not to unregister inputs when removed from the DOM).

I also thought putting a key on the input field might probably fix the problem, but it didn't!

**3. IMPORTANT: Provide a CodeSandbox reproduction of the bug**

https://codesandbox.io/s/competent-wu-fk7dp?file=/src/App.js

**4. Steps to reproduce**

1. Go to the CodeSandbox provided above
2. Fill the form
3. Submit the form (form will reset after submitting)
4. Fill the form again
5. At this point when you try to submit the form, the framer-motion input fields will show a required error even though they are valid (because they are no longer recognized by react-hook-form).
6. If you look in the console before and after resetting the form, you will see that the Refs of the motion input fields do not exist on the ```fieldsRef``` object anymore!

**5. Expected behavior**

Motion input fields should behave like a regular forwardRef input field

**7. Environment details**

Os: macOs
Browser: Chrome 85

mattgperry:
I think this functionality of `react-hook-form` is relying on a very specific peculiarity of how React handles callback refs: https://reactjs.org/docs/refs-and-the-dom.html#caveats-with-callback-refs

Motion isn't using `ref` incorrectly either and I think this is a bit of an edge case to actively support.

PR #1119, opened 2021-04-27
Update ref if externalRef changes

thebuilder:
Trigger an update if a `ref` callback function changes - This ensures everything is kept in sync.

I've gotten a bug report on this being an issue: https://github.com/thebuilder/react-intersection-observer/issues/466

See also: #742 

This could potentially cause an infinite rerender if the `ref` is not defined as `useCallback` method, but that would be a bug on the consumer side.
The correct implementation is listed in the Hooks FAQ: https://reactjs.org/docs/hooks-faq.html#how-can-i-measure-a-dom-node

### DIFF
--- a/src/motion/utils/use-motion-ref.ts
+++ b/src/motion/utils/use-motion-ref.ts
@@ -31,6 +31,6 @@ export function useMotionRef<Instance, RenderState>(
                 }
             }
         },
-        [visualElement]
+        [visualElement, externalRef]
     )
 }


mattgperry (approved):

mattgperry:
Sorry to hear that! I was under the impression this is covered in React docs? What bug has this surfaced?

thebuilder:
> 
> 
> Hey folks! This patch introduced a subtle breaking change into our application because of an instance where we were passing an inline function into the `ref` attribute of a `motion.div` element. I'm honestly not sure whether this is a mea culpa on our end or not, but I would suggest a tweak of Framer's (otherwise excellent) docs to clarify that callback functions should always be stable for `motion` elements.

The general recommendation is to not use inline functions for `ref` values, since they will get called twice on each render. 
https://reactjs.org/docs/refs-and-the-dom.html#caveats-with-callback-refs

When coupled with a component that uses a callback function for the node ref, like framer-motion, then you as the consumer must also ensure a stable `ref` function. Otherwise you get into the infinite render loop. But without this change, it's impossible to change the `ref` after mounting the component, since it would never get triggered again. 

It could be highlighted in the documentation that you must use a `useRef` or `useCallback` for getting the `ref`.

mattgperry:
Rolling this back with https://github.com/framer/motion/commit/2f6c163ddedb7c895be13850a4eee4d6c39e38af

This was my fault - there was a reason I didn't have `externalRef` in the dependencies and it was because of this issue in the linked docs. We use the ref callback to manage the lifecycle of the element and its animations. A fix that would support changing external refs would need to be much more involved to ensure it doesn't unmount existing visual elements.

thebuilder:
The caveat only applies to using an inline function directly. In this case, where it's passed to a `useCallback`, it's a bug that it's working. You will also be seeing bug reports from devs trying to dynamically set the `ref`, where it doesn't work as expected, eg: https://github.com/thebuilder/react-intersection-observer/issues/466

Might someone from the React team be able to weigh in with the best approach? Ideally React would be able to detect the infinite render and give a useful Warning message.

mattgperry:
I don’t think there’s so much a problem with supporting changing refs or letting external code break in a way it would have without framer-motion, but I know the reason the animations were breaking and it’s because Motion doesn’t correctly support the changing ref, the one we’re generating. Our code assumes this remains constant for the lifecycle of the element. So any change would need to refactor this.

thebuilder:
Fair enough. Ref's are one of those seemingly simple things that can give a lot of headaches.

PR #1141, opened 2021-05-10
Ensure projection box can't get locked by drag

mattgperry:
This adds a new callback to `PanSession`, `onSessionEnd`, that will call regardless of whether the drag technically started. Between this and moving `unlockViewportBox` from `stop` to `cancel` it should be even less likely the viewport box will be locked by drag.


hemlok (approved):
LGTM

PR #1151, opened 2021-05-15
add gradientTransform attribute to camel case attr array

### LINKED ISSUE #1150: [FEATURE] Supporting SVG gradientTransform Animations
It would be awesome if in the future framer-motion could animate the gradientTransform property of SVGs, like this:

```
const transition = { repeat: Infinity, repeatType: "mirror" }

 <motion.linearGradient
          id="gradient3"
          initial={{
            gradientTransform: "rotate(20)"
          }}
          animate={{
            gradientTransform: "rotate(180)"
          }}
          transition={{ ...transition, duration: 1 }}
        >
...
</ motion.linearGradient>
```

What this SHOULD do is create an animation that rotates the linear gradient from 20 degrees to 180 degrees back and forth infinitely, but currently this code does a 20-degree rotation initially and then never animates anything. Either I'm not understanding how to implement this correctly or framer-motion simply doesn't support gradientTransform animations yet.

Thanks!

KirdesMF:
It should resolve #1150 
See comments

### DIFF
--- a/src/render/svg/utils/camel-case-attrs.ts
+++ b/src/render/svg/utils/camel-case-attrs.ts
@@ -20,4 +20,5 @@ export const camelCaseAttributes = new Set([
     "stdDeviation",
     "tableValues",
     "viewBox",
+    "gradientTransform",
 ])


mattgperry:
Awesome thanks for this!

Issue #1152, opened 2021-05-16
[BUG] SVG Attributes 


KirdesMF:
This issue is related to #1150

First, I don't know if it's really needed to fill a bug report. Let me know if there is something wrong.
I should have checked when I made the PR #1151

there is others svg attributes that should be in the camelCase array

I realised that after trying to animate a svg pattern https://codesandbox.io/s/magical-williams-i8vv2?file=/src/styles.css

![image](https://user-images.githubusercontent.com/54989552/118400936-001fc900-b664-11eb-85d8-32068fae93dd.png)

Here is the list of camelCase svg attribute

- allowReorder
- attributeName
- attributeType
- autoReverse
- baseFrequency
- baseProfile
- calcMode
- clipPathUnits
- contentScriptType
- contentStyleType
- diffuseConstant
- edgeMode
- externalResourcesRequired
- filterRes
- filterUnits
- glyphRef
- gradientTransform
- gradientUnits
- kernelMatrix
- kernelUnitLength
- keyPoints
- keySplines
- keyTimes
- lengthAdjust
- limitingConeAngle
- markerHeight
- markerUnits
- markerWidth
- maskContentUnits
- maskUnits
- patternContentUnits
- patternTransform
- patternUnits
- pointsAtX
- pointsAtY
- pointsAtZ
- preserveAlpha
- preserveAspectRatio
- primitiveUnits
- repeatCount
- repeatDur
- requiredExtensions
- requiredFeatures
- specularConstant
- specularExponent
- spreadMethod
- startOffset
- stdDeviation
- stitchTiles
- surfaceScale
- systemLanguage
- tableValues
- textLength
- viewBox
- viewTarget
- xChannelSelector
- yChannelSelector

rijk:
I ran into this issue while trying to animate `startOffset`, it is still present in `framer-motion@6.3.3`.

willhindson:
Also a problem for me — it's animating the non-property "start-offset" instead of startOffset

PCPbiscuit:
I've created a PR which adds missing most used animated camel case svg attributes

rijk:
That will not get merged, [other people have done that already](https://github.com/framer/motion/pull/1153) and it was rejected due to bundle size.

If we were to solve this, it would have to be with a [solution like this](https://github.com/framer/motion/pull/1153#issuecomment-1135736249) that doesn't require a hardcoded list of attributes.

PR #1153, opened 2021-05-16
update list of camel case svg attributes

KirdesMF:
#1152 
it should be good now


KirdesMF:
https://codesandbox.io/s/framer-motion-simple-animation-forked-j53vx?file=/src/Example.tsx

mattgperry:
Thanks for this. I'm wondering though, given how obscure all these attr names are, if there's a way we could get rid of the array altogether.

rijk:
@mattgperry could we merge this in the meantime? Because it's currently not possible to animate these properties.

rijk:
Hm, I found that monkey patching this array didn't fix the issue for me.. Debugging, it seems `renderSVG` is not called at all? 🤔

mattgperry:
@rijk Is this still true with the latest version of Motion? Which element and attribute were you animating? 

Closing this out though as I'd rather these quite obscure props weren't animatable than add the extra bundlesize.

rijk:
Hi @mattgperry ! Yup, I am using `6.3.3` and it was not working. I was trying to animate `startOffset` for the scroll-triggered marquee text [on this site](https://made-baron.vercel.app). But when I tried to do this:

```jsx
<motion.textPath href="#path" startOffset={offset}>
  {children}
</motion.textPath>
```

it didn't work because it incorrectly converts the attribute to `start-offset`. So I had to resort to this:

```ts
useEffect(() => {
  return pathOffset.onChange((v) => {
    path.current?.setAttribute('startOffset', v)
  })
}, [])
```

which works but I think doesn't perform as well.

rijk:
Very annoying that the SVG attributes [are so inconsistent](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute).. I was thinking you could maybe assume camelCase when in an SVG context, but `glyph-orientation-vertical` vs `glyphRef`, `marker-start` vs `markerWidth`.. there is no logic to the naming it seems they just flipped a coin 🤔

**Edit:** ahh, the dash-case ones are apparently what they call "[Presentation attributes](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute#presentation_attributes)". That is why they are not camelCased:

>  SVG presentation attributes are CSS properties that can be used as attributes on SVG elements.

rijk:
So knowing this, I think a way to do it without bloating the bundle with a hardcoded list is to default to camelCase for SVG, and only use dash-case (not sure what the official name for that is) for these CSS presentational attributes;

```js
const svg = document.createElement('svg')
const key = attribute in svg.style ? property : camelCase(attribute)
```

Or more to the point of [the code](https://github.com/framer/motion/blob/f39eb0730d75ada35437e20dd8cff8284030f7eb/packages/framer-motion/src/render/svg/visual-element.ts#L28):

```js
key = camelToDash(key) in domElement.style ? camelToDash(key) : key
```

mattgperry:
That's very clever. I wonder how it can be done with SSR in mind. I think this is a good starting place though.

rijk:
Aren’t these functions (get/set current animation value) by their nature only run on the client?

mattgperry:
Yeah but off the top of my head we also output JSX to SSR the initial state

PR #1157, opened 2021-05-17
Remove hover event supression during layout measurements

mattgperry:
From testing, hover events don't fire if an element moves from under the element. I remember there was an effect like this in Framer but fundamentally they don't. So in this PR we remove this, if this kind of effect causes more issues we can find an alternative fix for the actual problem.


hemlok (approved):
👍

PR #1162, opened 2021-05-18
Update 'static.framer.com' references in Markdown files

koenrh:
We are moving away from serving unmanaged files on the `framer.com` domain, part of which is migrating content from `static.framer.com` to `misc.framerstatic.com`. The old links will keep on working through a 302 redirect, but let's update all the links anyway.

Ref: https://github.com/framer/company/issues/20653


mattgperry (approved):

mattgperry:
Nice one thanks for this

PR #1181, opened 2021-06-08
Fix prematurely removing elements that are still crossfading out

### LINKED ISSUE #1085: [BUG] State updates still interrupting layout animations
**Describe the bug**

This was reported in #732 and fixed in #1018 (v3.10.1), but it seems there still is an issue.

https://codesandbox.io/s/9eh9p?file=/src/Modal.tsx

**Steps to reproduce**

1. Click on a blue square
2. ☑️ Open animation no longer interrupted by state updates
3. Close modal
4. ❌ Close animation is interrupted by state update, comment icon seems to start new animation with default transition, border radius does not return to 10:

https://user-images.githubusercontent.com/159500/113399830-297ce200-93a1-11eb-91c4-82be7f784584.mov

**Expected behavior**

Comment out the `useEffect` in `Modal.tsx` to see the expected behaviour:

https://user-images.githubusercontent.com/159500/113399880-3b5e8500-93a1-11eb-82f3-a0cd622af027.mov

rijk:
I'll add some specific tests for this tomorrow, but I wanted to get this out already so you can take a look. I've ran all existing tests and the e2e tests.

Fixes #1085. 

For a succinct reproduction, see [this Sandbox](https://codesandbox.io/s/framer-motion-state-update-during-layout-transition-issue-9eh9p?file=/src/Test.tsx).

Basically what currently happens is elements in a AnimateSharedLayout + AnimatePresence tree are not animated out nicely, they just disappear, leaving only the following item to complete the crossfade. This can be faked by adding an `exit` animation with the same duration as the crossfade, but that's a workaround for something that should work out of the box.

This changeset defers calling `visualElement.notifyLayoutAnimationComplete()` from `animate` if the element is currently being crossfaded. Instead, the complete call is done from the AnimateSharedLayout stack when the crossfade is completed. I've slightly changed some signatures there, as there was no way yet to know when the crossfade was done; opted to add a Promise return to `stack.animate()`.

### DIFF
--- a/src/components/AnimatePresence/PresenceChild.tsx
+++ b/src/components/AnimatePresence/PresenceChild.tsx
@@ -39,11 +39,8 @@ export const PresenceChild = ({
             custom,
             onExitComplete: (childId: number) => {
                 presenceChildren.set(childId, true)
-                let allComplete = true
-                presenceChildren.forEach((isComplete) => {
-                    if (!isComplete) allComplete = false
-                })
-
+                const completed = [...presenceChildren.values()]
+                const allComplete = completed.every(Boolean)
                 allComplete && onExitComplete?.()
             },
             register: (childId: number) => {
--- a/src/components/AnimateSharedLayout/utils/crossfader.ts
+++ b/src/components/AnimateSharedLayout/utils/crossfader.ts
@@ -4,7 +4,6 @@ import {
     linear,
     mix,
     mixColor,
-    PlaybackControls,
     progress as calcProgress,
 } from "popmotion"
 import { animate } from "../../../animation/animate"
@@ -16,8 +15,8 @@ import { motionValue } from "../../../value"
 export interface Crossfader {
     isActive(): boolean
     getCrossfadeState(element: VisualElement): ResolvedValues | undefined
-    toLead(transition?: Transition): PlaybackControls
-    fromLead(transition?: Transition): PlaybackControls
+    toLead(transition?: Transition): Promise<void>
+    fromLead(transition?: Transition): Promise<void>
     setOptions(options: CrossfadeAnimationOptions): void
     reset(): void
     stop(): void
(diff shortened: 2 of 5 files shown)


src/components/AnimatePresence/PresenceChild.tsx
```diff
@@ -39,11 +39,8 @@ export const PresenceChild = ({
             custom,
             onExitComplete: (childId: number) => {
                 presenceChildren.set(childId, true)
-                let allComplete = true
-                presenceChildren.forEach((isComplete) => {
-                    if (!isComplete) allComplete = false
-                })
-
+                const completed = [...presenceChildren.values()]
+                const allComplete = completed.every(Boolean)
```
rijk, line 43: Just small refactor, no functional change.

src/components/AnimateSharedLayout/utils/crossfader.ts
```diff
@@ -71,42 +70,47 @@ export function createCrossfader(): Crossfader {
         finalCrossfadeFrame = null
         let hasUpdated = false
```
rijk, line 72: Everything below is mostly just whitespace changes for the Promise return. Only new line is the `resolve()` from the `onComplete()`.

rijk:
I have noticed in testing that a non-linear crossfade (like `type: 'spring'`) is marked as completed too soon. Don't know if that's related to this change though, might be an existing issue. I'll investigate further.

src/components/AnimatePresence/PresenceChild.tsx
```diff
@@ -39,11 +39,8 @@ export const PresenceChild = ({
             custom,
             onExitComplete: (childId: number) => {
                 presenceChildren.set(childId, true)
-                let allComplete = true
-                presenceChildren.forEach((isComplete) => {
-                    if (!isComplete) allComplete = false
-                })
-
+                const completed = [...presenceChildren.values()]
```
mattgperry, line 42: Doesn't values() already return a new array?

mattgperry:
Yeah I think this is a good approach, my only brief thought is what if the crossfade completes before the layout animation? Is this not inverting that problem?

src/components/AnimatePresence/PresenceChild.tsx
rijk: No, it returns a `MapIterator`, which behaves like an array in some cases but doesn't have `forEach`, `map`, `filter`, `some`, `every`, etc.

rijk:
Hmm interesting, you're right, I hadn't thought of that. When I use a transition like this:

```
transition={{
  crossfade: { duration: 1 },
  default: { duration: 5 },
}}
```

it does seem to interrupt (restart) the layout animation as well. I guess this is the same problem as [I've noted above](https://github.com/framer/motion/pull/1181#issuecomment-858622430).

What would be a better way to approach this?

mattgperry:
I'm not certain the architecture lends itself but some kind of `Promise.all`. But I also notice there's more difference here. For instance `onNotifyLayoutAnimationComplete` currently gets called when a layout animation completes, whereas in this PR it's only when a crossfade `fromLead` finishes. In the instances where this doesn't run the promise will resolve immediately which isn't correct either and could have unintended consequences of its own.

rijk:
Yeah, I was mistaking the crossfade for the whole thing, including the layout animation. But it's just the opacity switch. Ideally we need to make sure the crossfade is added to the list of animation promises [in Animate](https://github.com/rijk/motion/blob/fix/crossfade-exit/src/motion/features/layout/Animate.tsx#L321) I guess.

rijk:
But the opacity is actually not all; because the main goal is to keep the element in the tree until the layout animation is completed. I don't actually care about the crossfade, the main problem is that the element is removed too soon, triggering a new layout animation while the previous one is still running. That is the main problem I was trying to solve.

rijk:
Ahh hold on, I think I know what the problem is. This should actually be a simple bugfix. What happens is:

1. the axis animations are added to `Promise.all()` and all is well
2. a state change happens, triggering layout animation check
3. no (new) animations; `Promise.all()` resolves immediately
4. `visualElement.notifyLayoutAnimationComplete()` is called — before the promises from step 1 have resolved.

PR #1183, opened 2021-06-10
Fix layout animations interrupted by rerenders

### LINKED ISSUE #1085: [BUG] State updates still interrupting layout animations

rijk:
Supersedes #1181.

**Problem**
State updates that happen during a layout transition, trigger a new call to `animate()`. What then happens is this:

1. For the intial layout animation, the axis animations are added to `Promise.all()` and all is well
2. A state change happens, triggering a new call to `animate()`
3. No (new) animations; `Promise.all()` is called with no animations and resolves immediately
4. `visualElement.notifyLayoutAnimationComplete()` is called — before the promises from step 1 have resolved.

As a result the exiting element disappears mid transition, and a new transition is started for the following element. This can be seen in the Sandbox from the linked issue. 

**Solution**
This change prevents no-op rerenders (e.g. internal state changes that don’t result in a layout animation) from calling `Promise.all()` with `[undefined, undefined]` resulting in a preliminary call to `onComplete()` and `visualElement.notifyLayoutAnimationComplete()`.

Fixes #1085.

### DIFF
--- a/src/motion/features/layout/Animate.tsx
+++ b/src/motion/features/layout/Animate.tsx
@@ -289,11 +289,13 @@ class Animate extends React.Component<AnimateProps> {
          * kept in by the tree by AnimatePresence) then call safeToRemove when all axis animations
          * have successfully finished.
          */
-        return Promise.all(animations).then(() => {
-            this.isAnimatingTree = false
-            onComplete && onComplete()
-            visualElement.notifyLayoutAnimationComplete()
-        })
+        if (animations.filter(Boolean).length) {
+            return Promise.all(animations).then(() => {
+                this.isAnimatingTree = false
+                onComplete && onComplete()
+                visualElement.notifyLayoutAnimationComplete()
+            })
+        }
     }
 
     /**
--- a/src/motion/features/layout/Animate.tsx
+++ b/src/motion/features/layout/Animate.tsx
@@ -281,15 +281,15 @@ class Animate extends React.Component<AnimateProps> {
             }
         })
 
-        // Force a render to ensure there's no flash of uncorrected bounding box.
-        visualElement.syncRender()
-
         /**
          * If this visualElement isn't present (ie it's been removed from the tree by the user but
          * kept in by the tree by AnimatePresence) then call safeToRemove when all axis animations
          * have successfully finished.
          */
         if (animations.filter(Boolean).length) {
+            // Force a render to ensure there's no flash of uncorrected bounding box.
+            visualElement.syncRender()
+
             return Promise.all(animations).then(() => {
                 this.isAnimatingTree = false
                 onComplete && onComplete()
--- a/src/motion/features/layout/Animate.tsx
+++ b/src/motion/features/layout/Animate.tsx
@@ -295,6 +295,8 @@ class Animate extends React.Component<AnimateProps> {
                 onComplete && onComplete()
                 visualElement.notifyLayoutAnimationComplete()
             })
+        } else if (shouldStackAnimate) {
+            this.isAnimatingTree = false
         }
     }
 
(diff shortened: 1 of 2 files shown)


rijk:
Ok, it's failing the `drag-svg` and `layout-shared-lightbox-crossfade-repeated` e2e tests at the moment. Need to do some more research (or feel free to improve yourself) and will also add a test for this case.

rijk:
To be honest, I don't think the original one should be passing either:

https://user-images.githubusercontent.com/159500/121568482-09604380-ca20-11eb-8f63-f2eaf8d8b528.mp4

When I run the test (http://localhost:9990/?test=layout-shared-lightbox-crossfade&partial-ease=true) manually, I get the lightbox when clicking the green or blue box, but the red one is giving me this weird intermittent state as well:

image

This even happens on `main`, so might be an issue that has gone unnoticed.

rijk:
Ok, I think I got it now. Tests are passing.

mattgperry (approved):

mattgperry:
Yeah this is only possible with Cypress, though I'm not 100% how you'd test this. Some of the ways I've been doing it is writing demos that use a `transition` of `ease: () => 0.5` and ensuring at least the mid-point of the animation is correct but I'm not sure if that would help in this instance. The change looks good to me though.

src/motion/features/layout/Animate.tsx
```diff
@@ -281,19 +281,21 @@ class Animate extends React.Component<AnimateProps> {
...
-            this.isAnimatingTree = false
-            onComplete && onComplete()
-            visualElement.notifyLayoutAnimationComplete()
-        })
+        if (animations.filter(Boolean).length) {
+            // Force a render to ensure there's no flash of uncorrected bounding box.
+            visualElement.syncRender()
+
+            return Promise.all(animations).then(() => {
+                this.isAnimatingTree = false
```
mattgperry, line 294: If `animate` has been called and `this.isAnimatingTree` set to `true`, when will it get set to `false` if no animations are created?

src/motion/features/layout/Animate.tsx
```diff
@@ -281,19 +281,21 @@ class Animate extends React.Component<AnimateProps> {
...
+                onComplete && onComplete()
+                visualElement.notifyLayoutAnimationComplete()
+            })
+        }
```
mattgperry, line 298: Perhaps

```
 else if (shouldStackAnimate) { this.isAnimatingTree = false }
```

src/motion/features/layout/Animate.tsx
rijk: I tried searching the codebase but I don't see it being used anywhere:

image

rijk:
Hi @mattgperry , could you merge this into `fix/jump` and publish a beta update? Cheers

src/motion/features/layout/Animate.tsx
mattgperry: `shouldStackAnimate` is a Framer-internal thing so it isn't provided by anything in Motion

mattgperry:
Yeah can you add this https://github.com/framer/motion/pull/1183#discussion_r649772071

mattgperry (approved):

PR #1188, opened 2021-06-16
Make layoutAnimations exportable

RuudBurger:
Make layoutAnimations exportable so it can be used in LazyMotion without needing to load the `drag` feature


mattgperry:
Closing this out, I understand the desire here but in testing how the treeshaking works splitting these bundles up too finely can inadvertently increase overall bundlesize, often with some things ending up in the main bundle that you don't want.

I do want to take another run at fixing this problem though so hopefully this won't be an issue for too much longer

Issue #1178, opened 2021-06-02
[BUG] Proxy is not defined in older browsers


alirezavlz:
Hey, I'm using the framer-motion in my project, I'm using browserstack to test the application, it seems the framer-motion is using `Proxy` which is not polyfillable. Is there any other way to use framer motion that supports older browsers?

I saw the issue on Chrome v38

![image](https://user-images.githubusercontent.com/53314423/120521049-cb866e00-c3d4-11eb-9264-cde186b6551c.png)

mattgperry:
From `5.3.3` later today you'll be able to define components like this:

```javascript
const MotionDiv = motion("div")
```

If you need backwards compatibility with old browsers.

Issue #674, opened 2020-07-29
support ios 9 ?


littlee:
after importing `proxy-polyfill`,  `motion.div` is undefined on ios 9 device

```js
import 'proxy-polyfill';
import { motion } from 'framer-motion';
console.log(motion.div); // undefined
```

(I used to use `react-pose`, and it works fine on ios 9 with core-js polyfill)

```
import 'core-js/stable';
import 'regenerator-runtime/runtime';
```

InventingWithMonster:
Hey @littlee

Motion doesn't support runtimes without `Polyfill` support I'm afraid, and it isn't polyfillable. I am considering a seperate entry point for people who absolutely need it and don't mind the added bundle size though.

littlee:
found `motion.custom` here
https://inventingwithmonster.io/20200302-can-i-use-framer-motion-with-styled-components/
(missing docs for `motion.custom` ? 🌚 )

on iOS 9
though `motion.div` is undefined, `motion.custom` can be used instead.

After re-write my component to
```
import { motion } from 'framer-motion';
const Wrap
```

Issue #664, opened 2020-07-25
Not working on IE11 because in Framer Motion v2 "Proxy" is used, which is not supported [BUG]


asadhazara:
I've been using till v1.11.1 without any problem in IE11, but since the upgrade to v2 my website shows an unexpected error, which is "Proxy in undefined".

InventingWithMonster:
Hi @asadhazara 

Framer Motion hasn't ever supported IE11. Have you been using polyfills until now? Proxy isn't supported by polyfills.

asadhazara:
@InventingWithMonster Yes, i've been using polyfills. Thank you for the clarification. I guess I have to drop supporting IE11 aswell :)

PR #1194, opened 2021-06-24
feat: add patchMotionTags for legacy browsers which don't support Proxy

ambar:
Related: #364 #664 #674 #1178

## Usage
```js
import {patchMotionTags} from 'framer-motion'

patchMotionTags()
```

Test result:
| Tool | Bundle Diff | 
| --- | --- | 
| webpack | +1070B (gzip 464B) |
| esbuild | +1068B (gzip 469B) |

---
A less convenient solution is to rewrite export, but it will significantly increase the bundle size:

```js   
export let motion = motionProxy
// motion will always be bundled
export const patchMotion = () => {
    motion = createMotionFallback()
}
```


ppsirius:
What is the status of this?

mattgperry:
Sorry for the delay on this but closing this out as the point of using `Proxy` in the first place is to avoid having to maintain a list of valid tags that has to be then included in the bundle.

ambar:
The DOM tags in this patch would not be bundled unless you call `patchMotionTags()`.

Right now, all other workarounds are a little pain to use:

- `yarn patch` requires line number of source code (unstable)
- Aliasing requires different bundler config in different projects (and also, we may need to inject a fake Proxy polyfill and then revert it)
- Re-exporting m/motion from another package (we end up with this)

mattgperry:
They would be included though, because the bundler has no way of knowing whether `patchMotionTags` is going to be called by the client.

PR #1212, opened 2021-07-08
Fix: don't parse shadow string starts with 0

shuangq:
Shadow string like `0px 0px rgba(0,0,0,0)` is parsed to `0`, causing the code to crash when the shadow corrector checks `latest.includes`, see: https://github.com/framer/company/issues/22628

### DIFF
--- a/src/utils/is-zero-value-string.ts
+++ b/src/utils/is-zero-value-string.ts
@@ -1,4 +1,4 @@
 /**
  * Check if the value is a zero value string like "0px" or "0%"
  */
-export const isZeroValueString = (v: string) => /^0[\S]+$/.test(v)
+export const isZeroValueString = (v: string) => /^0[^.\s]+$/.test(v)
--- a/src/utils/is-zero-value-string.ts
+++ b/src/utils/is-zero-value-string.ts
@@ -1,0 +1,4 @@
+/**
+ * Check if the value is a zero value string like "0px" or "0%"
+ */
+export const isZeroValueString = (v: string) => /^0[\S]+$/.test(v)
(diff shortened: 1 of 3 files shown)


src/utils/is-zero-value-string.ts
```diff
@@ -0,0 +1,4 @@
+/**
+ * Check if the value is a zero value string like "0px" or "0%"
+ */
+export const isZeroValueString = (v: string) => /^0[\S]+$/.test(v)
```
nvh, line 4: What is the `\S` character in a regex?

nvh (approved):

nvh:
Shouldn’t the base of this commit be `main`, and that we cherry-pick it into `framer`?

shuangq:
> Shouldn’t the base of this commit be `main`, and that we cherry-pick it into `framer`?

Ah was wondering the same. Will rebase!

src/utils/is-zero-value-string.ts
shuangq: `\S` matches any non-whitespace characters. The regex here translates to "starts with 0 and doesn't contain any space until the end of the line".

nvh:
Hmm, @mattgperry You also applied the fix to the layout project branch, right? I remember us discussing this.
Thread here: https://framer-team.slack.com/archives/CR3CYA1D4/p1625152722009800?thread_ts=1625139055.481500&cid=CR3CYA1D4

nvh:
So I suggest we land it on the `framer` branch and then cherry pick it on the `feature/perfect-projection` branch

mattgperry (approved):

PR #1213, opened 2021-07-09
Update README.md

perilanglois:
updated readme with new image and link to Framer for Developers


mattgperry:
The images are broken here: https://github.com/framer/motion/tree/perilanglois-patch-1

I think it's dem curly quote marks instead of straight down quote marks (see diff)

perilanglois:
> The images are broken here: https://github.com/framer/motion/tree/perilanglois-patch-1
> 
> I think it's dem curly quote marks instead of straight down quote marks (see diff)

okay how bout now......

mattgperry:
Wonderful thanks

mattgperry (approved):

Issue #1214, opened 2021-07-12
Layout projection breaks after scrolling


rijk:
**CodeSandbox reproduction of the bug**
https://codesandbox.io/s/framer-motion-scroll-jump-issue-y97qn

**Steps to reproduce**

1. Click Menu button
2. Scroll colored area horizontally (make window smaller if there is no scroll)
3. Click one of the blue squares
3. Note white project container div shifting in the background while the modal is opening
4. Close modal
5. Note layout animation borked (lead/follow no longer synced)
6. Click Menu button 
7. Note project position is 'fixed' again by framer-motion when the tree is rerendered

**Expected behavior**

Scrolling should not affect layout transition functionality. When the transition starts it should calculate the correct coordinates taking the scroll position into account. Also, the parent div has no reason to shift horizontally as it does not even rerender?

**Video or screenshots**

https://user-images.githubusercontent.com/159500/125266186-dbfc0380-e305-11eb-9059-1d2dd8446e13.mov

mattgperry:
This should be fixed with the perfect-projection branch but let’s keep this open so we can track it

rijk:
[Not fixed](https://codesandbox.io/s/framer-motion-scroll-jump-issue-on-v500-pm75q?file=/src/Step.tsx) on `5.0.0-beta.22`:


https://user-images.githubusercontent.com/159500/131320089-411da375-8746-43ee-8f68-1fe134feb01f.mov

Here you can see initially the scrolling doesn't pose a problem, it starts when the sidebar is expanded:


https://user-images.githubusercontent.com/159500/131320284-75466404-3adc-43f4-b829-9195dab0878d.mov

Maybe it's because framer-motion assumes the scroll is directly relative to the viewport?

rijk:
Fixed in 5.0.0-beta.28. Great work!

rijk:
@mattgperry This is broken again in the latest version: https://codesandbox.io/s/framer-motion-scroll-jump-issue-2x0y4

https://user-images.githubusercontent.com/159500/134505447-a9649d7f-ce87-45b0-9348-fa68f96ff9a7.mov

rijk:
More specifically this was broken between beta.35 and beta.36.

mattgperry:
Scrollable elements need to be tagged with the `shouldMeasureScroll` prop (changing to `layoutScroll` in beta 45).

```jsx
<div className="projects">
// Becomes
<motion.div className="projects" shouldMeasureScroll>
```

Issue #1209, opened 2021-07-04
Existing layout projection breaks after resizing window


rijk:
**CodeSandbox reproduction of the bug**
https://codesandbox.io/s/framer-motion-border-radius-layout-issue-forked-to3ji

**Steps to reproduce**
This might be fixed with the projection rewrite, so just posting this issue so you can double check. 

1. Click one of the blue squares
2. Resize the window
3. Click again to close
4. Observe lead/follow no longer aligned, and they stay misaligned after this. Also observe the projects are distorted in width, and no longer aligned horizontally.

**Video or screenshots**

https://user-images.githubusercontent.com/159500/124381620-a92d8c00-dcc3-11eb-8bd6-3b8d8b077be5.mov

rijk:
[Fixed on `5.0.0-beta.22`!](https://codesandbox.io/s/framer-motion-window-resize-issue-on-v5-5-0-ww8lw?file=/src/Board.tsx)

Issue #1208, opened 2021-07-03
Border radius jumps to circle at the end of layout transition


rijk:
Might be related to #1205, however this happens _during_ the animation, as opposed to after it finishes.

**CodeSandbox reproduction of the bug**
https://codesandbox.io/s/framer-motion-border-radius-layout-issue-forked-x8t3w?file=/src/styles.css

**Steps to reproduce**

1. Click one of the blue squares
2. Click again to close the modal
3. At the end of the transition, the border radius of the modal suddenly is increased turning it into a circle

**Expected behavior**

The border radius should keep transitionining into the blue square's border radius.

**Video or screenshots**

https://user-images.githubusercontent.com/159500/124361713-c9603b00-dc30-11eb-9fcc-03852ed500ef.mov

rijk:
[Fixed](https://codesandbox.io/s/framer-motion-border-radius-layout-circle-issue-on-v500-ugyso) on `5.0.0-beta.22` (although the modal doesn't animate in for some reason, but I'm not sure this is my error).

pixelpandacreative:
> [Fixed](https://codesandbox.io/s/framer-motion-border-radius-layout-circle-issue-on-v500-ugyso) on `5.0.0-beta.22` (although the modal doesn't animate in for some reason, but I'm not sure this is my error).

It's still happening as of today.

Issue #1207, opened 2021-07-03
Slow spring animation is cut off after a while


rijk:
**CodeSandbox reproduction of the bug**
https://codesandbox.io/s/framer-motion-border-radius-layout-issue-forked-s4ymf?file=/src/Modal.tsx

**Steps to reproduce**
I am not 100% sure, but this does not seem like a smooth transition to me. Of course it could be because the config values are just too extreme, but it could also be a bug. This is the transition config:

```ts
const transition: Transition = {
  type: "spring",
  stiffness: 4,
  damping: 35,
  mass: 0.5
};
```

1. Click one of the blue squares
2. Midway during the animation, it suddenly accelerates to snap to the end

**Expected behavior**

The transition should keep an even pace until the end.

**Video or screenshots**

https://user-images.githubusercontent.com/159500/124361203-260e2680-dc2e-11eb-9cd1-05f5ef7fafea.mov

https://user-images.githubusercontent.com/159500/124361207-2ad2da80-dc2e-11eb-9511-2db4ff4f8523.mov

rijk:
[Still not working](https://codesandbox.io/s/framer-motion-slow-spring-animation-issue-on-v5-0-0-jv0b3) on `5.0.0-beta.22`:


https://user-images.githubusercontent.com/159500/131315839-2e904b80-827c-4b7d-88e3-7098f17cfda7.mov

eric-gitta-moore:
time based spring same here.

In the Spring Visualiser editor, one can also observe a strange jitter at the end of the animation.

<img width="996" height="1101" alt="Image" src="https://github.com/user-attachments/assets/22a7a586-8b74-4a91-829a-8f33d3416069" />

PR #1226, opened 2021-07-27
Perfect projection

mattgperry:
## TODO

- [ ] Add tests for `originX`/`originY`


rijk:
At which point is testing feedback welcome? (as this is marked as beta already)

mattgperry:
Yep it's welcome @rijk

rijk:
## Tickets

- [ ] #1187 — not fixed (updated screen recording in issue)
- [ ] #1190 — new issue when interrupting transition (details in issue)
- [ ] #1205 — modal open doesn't transition
- [ ] #1206 — modal open doesn't transition
- [ ] #1207 — still not working
- [x] #1208 — fixed, although animation is cut off at the end (see previous)
- [x] #1209 
- [ ] #1214 — not fixed, see details in issue
- [ ] #1085 — was fixed in `4.1.18-rc.26` (fix/jump branch), but broken again on `5.0.0-beta.22`

## API

### Removing `<AnimateSharedLayout>`
While it's cool that AnimateSharedLayout is no longer necessary, removing it in favour of a global `layoutId` system requires refactoring application code ([here's an example of how I use it right now](https://codesandbox.io/s/framer-motion-border-radius-layout-issue-bdneo?file=/src/Step.tsx)). Could it be possible to make this update more backwards compatible by reimplementing AnimateSharedLayout, and e.g. adding a auto-generated prefix to the childs' `layoutId`s?

### Drag to reorder example
Could you update the drag to reorder example again for v5? The implementation will probably be different, at least the one I used in v4 [stopped working](https://codesandbox.io/s/framer-motion-2-drag-to-reorder-on-v5-5-0-8x0ru?file=/src/App.js).

rijk:
Here's a focused Sandbox for the modal open animation that's not working: https://codesandbox.io/s/framer-motion-v5-5-0-modal-open-animation-not-working-2497l?file=/src/Board.tsx

It turns out the problem is caused by a wrapping `<motion.div layout>` (which I need for a sidebar animation), when you comment that out in the Sandbox you see the expected behaviour.


https://user-images.githubusercontent.com/159500/132219953-3e7fdced-ae31-4d5c-8a6a-1d6332605556.mov

dev/projection/element-scroll-remove.html
```diff
@@ -0,0 +1,137 @@
...
+
+            matchViewportBox(box1, {
+                top: 50,
+                bottom: 150,
+                left: 0,
+                right: 100,
+            })
+        </script>
+    </body>
+</html>
```
rijk, line 137: This test appears to pass even without the below changes, is that right?

rijk (commented):
_Disregard this comment please, sorry, I can't remove it._

rijk:
Can you release the new beta to NPM as well? I want to try it out in Codesandbox but it's still showing beta.28 as the latest version.

mattgperry:
@rijk It is on npm, sometimes CodeSandbox has been slow in updating versions. I've also pushed Git tags, maybe that's where they get them from?

rijk:
Still nothing..

mattgperry:
> Still nothing..

This isn't an us problem: 
Screenshot 2021-09-23 at 15 12 36
on

rijk:
Are you implying it's something on my end? Or Codesandbox's end? I see I am able to install it via package.json indeed, so I'll use that. It's just (still) not showing up in the dependency menu:

image

mattgperry:
@rijk Perhaps CodeSandbox. But the tags are pushed

PR #1228, opened 2021-07-28
Make onAnimateStart be called with animation definition

raon0211:
It seems that unlike `onAnimationComplete`, `onAnimationStart` is not called with the animation definition in its first argument. 

For consistency and ease of use, I made `onAnimationStart` to be called with the animation definition.


mattgperry:
Yes great call, thanks for the PR!

Issue #1143, opened 2021-05-10
[FEATURE] Passive event listeners


o-alexandrov:
**Is your feature request related to a problem? Please describe.**
Passive event listeners affect:
- performance
- SEO (decreasing Google’s PageSpeed score)

**Describe the solution you'd like**
Unfortunately, I am not so intimate with the library, so I cannot say whether you had special reasons to set event listeners as not passive.

**Describe alternatives you've considered**
Going w/ passive event listeners is a breaking change for some.
That’s why I request to modify the API, so developers are able to toggle whether they want to invoke listeners as passive.

**Additional context**
Reproducible since framer-motion v1.x and still the case in the latest version

o-alexandrov:
@mattgperry it would have, if `motion` used `preventDefault` and if there were users (developers) who depend on the scroll blocking nature of the `passive: false` and `preventDefault`.

I just pulled latest `main` branch and searched for use of `preventDefault`:
- there are no instances of invoking `preventDefault`

If you approve this change, I could create a PR to:
1. modify `addDomEvent` to mutate passed `options` applying `passive: true`, if `passive` property is not specified
2. remove all instances in existing code, where we specify `passive: true` (there will be no point in doing so, as we set it to be the default)

PR #1233, opened 2021-08-04
Improve performance of addEventListener usage

o-alexandrov:
Based on the issue/discussion here https://github.com/framer/motion/issues/1143#issuecomment-838398183
In short, since:
- there's no usage of `event.preventDefault()`
  - so there is no possibility of affecting any code within `motion` repo and users
- you bump the major version due to other improvements
  - so users who use `useDomEvent` method in their projects and depend on `event.preventDefault()` won't pull this PR changes if they have `^` for the version of the library


o-alexandrov:
Closes #1143

PR #1251, opened 2021-09-09
Fix: relative projection for instant transition

shuangq:
This PR fixes 2 issues:
- Relative projection for child components with a 0s transition. https://github.com/framer/company/issues/22848 
By starting the animation one frame later we ensure there's an animation frame with progress 0, so on the first frame the target should be the same as when the animation actually starts. This prevents elements with a 0s transition from shooting out of their parent boxes.
- A projection node might get an outdated layout on mount if the ancestors update their scroll after the prev lead unmounts. https://github.com/framer/company/issues/22872

### DIFF
--- a/dev/projection/animate-single-element-layout-change-with-child.html
+++ b/dev/projection/animate-single-element-layout-change-with-child.html
@@ -85,23 +85,25 @@
 
             boxProjection.root.didUpdate()
 
-            requestAnimationFrame(() => {
-                matchViewportBox(box, {
-                    top: 50,
-                    right: 270,
-                    bottom: 170,
-                    left: 100,
+            sync.postRender(() => {
+                sync.postRender(() => {
+                    matchViewportBox(box, {
+                        top: 50,
+                        right: 270,
+                        bottom: 170,
+                        left: 100,
+                    })
+                    matchViewportBox(child, {
+                        top: 60,
+                        right: 160,
+                        bottom: 110,
+                        left: 110,
+                    })
+                    matchOpacity(box, 1)
+                    matchOpacity(child, 1)
+                    matchBorderRadius(box, 0)
+                    matchBorderRadius(child, "40%")
                 })
-                matchViewportBox(child, {
-                    top: 60,
-                    right: 160,
-                    bottom: 110,
-                    left: 110,
-                })
-                matchOpacity(box, 1)
-                matchOpacity(child, 1)
-                matchBorderRadius(box, 0)
-                matchBorderRadius(child, "40%")
             })
         </script>
     </body>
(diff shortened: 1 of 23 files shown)


dev/projection/animate-single-element-layout-change-with-child.html
```diff
@@ -85,23 +85,25 @@
...
             boxProjection.root.didUpdate()
 
-            requestAnimationFrame(() => {
-                matchViewportBox(box, {
-                    top: 50,
-                    right: 270,
-                    bottom: 170,
-                    left: 100,
+            sync.postRender(() => {
+                sync.postRender(() => {
```
mattgperry, line 89: By nesting these two frames deep is there not a flash of the wrong bounding box?

src/projection/node/create-projection-node.ts
```diff
@@ -356,6 +356,20 @@ export function createProjectionNode<I>({
...
             }
 
+            if (
+                this !== this.root &&
+                this.options.shouldMeasureScroll === undefined &&
+                this.instance instanceof Element
+            ) {
+                const computedStyle = getComputedStyle(this.instance)
+                if (
+                    computedStyle.overflowX === "scroll" ||
```
mattgperry, line 366: This will trigger a synchronous style recalculation/layout https://gist.github.com/paulirish/5d52fb081b3570c81e3a#calling-getcomputedstyle

IMO it would be a reasonable API constraint to either 1) only check the React `style` prop and not support setting this via other CSS (preferable) or 2) make it an opt-in via a prop

shuangq: The first option is not enough to fix the issue as we do inject the CSS via a classname in FramerStudio. We can probably do both.

dev/projection/animate-single-element-layout-change-with-child.html
shuangq: The first frame would be still on progress 0 and the second to the mid box right?

src/projection/node/create-projection-node.ts
mattgperry: If it's possible to also add it via style this would be much more performant.

dev/projection/animate-single-element-layout-change-with-child.html
mattgperry: Oh yeah doh

src/projection/node/create-projection-node.ts
shuangq: Introduced a new prop `shouldMeasureScroll` instead!

mattgperry: Awesome!

mattgperry (approved):

PR #1255, opened 2021-09-16
Fix lingering snapshot

mattgperry:
At the end of a layout update we clear snapshots so they don't get reused in subsequent layout animations.

If we remove an element with a `layoutId`, it gets removed from the global list of nodes. Incoming elements can access its snapshot via the layoutId stack. At the end of a layout update we use the global list of nodes to clear snapshots, this PR ensures to go through stacks and clear snapshots for any removed nodes.

Additionally, if there's only one element being removed `willUpdate` is called but nothing is around to call `didUpdate`. This means snapshots aren't flushed at all. This PR adds a schedules a check, from `componentWillUnmount`, that fires at the end of the next animation frame, to see whether the layout update failed. This lifecycle run from `componentWillUnmount` through to layout effect is AFAIK synchronous, so the next animation frame will always occur afterwards.


package.json
```diff
@@ -37,7 +37,7 @@
         "test-server": "jest --config jest.config.ssr.json",
         "test-projection": "yarn run collect-projection-tests && start-server-and-test 'python -m SimpleHTTPServer' http://localhost:8000 'yarn run cypress run -s cypress/integration/projection.ts --config baseUrl=http://0.0.0.0:8000/'",
         "test-e2e": "start-server-and-test start-dev-server http://localhost:9990 'cypress run --config ignoreTestFiles=projection.ts'",
-        "test-e2e-file": "start-server-and-test start-dev-server http://localhost:9990 'cypress run --spec \"cypress/integration/drag-to-reorder.ts\"'",
+        "test-e2e-file": "start-server-and-test start-dev-server http://localhost:9990 'cypress run --spec \"cypress/integration/layout-shared.ts\"'",
```
nvh, line 40: Shouldn't this be just an argument instead of committed to git?

api/framer-motion.api.md
```diff
@@ -451,6 +451,8 @@ export interface IProjectionNode<I = unknown> {
     // (undocumented)
     root?: IProjectionNode;
     // (undocumented)
+    scheduleCheckUpdateFailed(): void;
```
nvh, line 454: What about `scheduleUpdateFailedCheck()`?

src/projection/shared/stack.ts
```diff
@@ -106,4 +107,14 @@ export class NodeStack {
             node.instance && node.scheduleRender(false)
         })
     }
+
+    /**
+     * Clear any leads that have been removed this render to prevent them from being
+     * used in future animations and to prevent memory leaks
+     */
+    clearLeadSnapshot() {
```
nvh, line 115: Not sure if "clear" is used in other places too, but something like `removeLeadSnapshot` makes a bit more sense to me.

nvh (approved):

PR #1256, opened 2021-09-17
Block layout animations on resize

mattgperry:


src/projection/use-instant-layout-transition.ts
```diff
@@ -3,10 +3,12 @@ import { rootProjectionNode } from "./node/HTMLProjectionNode"
 export function useInstantLayoutTransition(): (
     cb?: (() => void) | undefined
 ) => void {
+    console.log("render")
     return startTransition
 }
 
 function startTransition(cb?: () => void) {
+    console.log("starting transition, blocking update")
```
nvh, line 11: These logs can be removed.

package.json
```diff
@@ -37,7 +37,7 @@
         "test-server": "jest --config jest.config.ssr.json",
         "test-projection": "yarn run collect-projection-tests && start-server-and-test 'python -m SimpleHTTPServer' http://localhost:8000 'yarn run cypress run -s cypress/integration/projection.ts --config baseUrl=http://0.0.0.0:8000/'",
         "test-e2e": "start-server-and-test start-dev-server http://localhost:9990 'cypress run --config ignoreTestFiles=projection.ts'",
-        "test-e2e-file": "start-server-and-test start-dev-server http://localhost:9990 'cypress run --spec \"cypress/integration/layout-shared.ts\"'",
+        "test-e2e-file": "start-server-and-test start-dev-server http://localhost:9990 'cypress run --spec \"cypress/integration/layout-resize.ts\"'",
```
nvh, line 40: I think I mentioned this before, but should we remove this filename from the package.json?

src/projection/node/create-projection-node.ts
```diff
@@ -432,22 +446,26 @@ export function createProjectionNode<I>({
...
+        }
+
+        isUpdateBlocked() {
+            return this.updateManuallyBlocked || this.updateBlockedByResize
         }
 
         // Note: currently only running on root node
         startUpdate() {
-            if (this.updateBlocked) return
+            if (this.updateManuallyBlocked) return
```
nvh, line 462: Shouldn't we check `isUpdateBlocked()` here?

package.json
mattgperry: I don't think so, it makes it a lot easier to run just one test in this otherwise quite lengthy suite, and as it's in the interest of time prevents us having to rewrite the command each time

src/projection/node/create-projection-node.ts
```diff
@@ -351,12 +353,24 @@ export function createProjectionNode<I>({
...
-                if (globalProjectionState.hasAnimatedSinceResize) {
-                    globalProjectionState.hasAnimatedSinceResize = false
-                    this.nodes!.forEach(finishAnimation)
-                }
-            })
+            if (attachResizeListener) {
+                let unblockTimeout: number
+
+                const unblockUpdate = () =>
+                    (this.root.updateBlockedByResize = false)
```
nvh, line 360: There can be confusion between this and `this.unblockUpdate` can we rename this to something more specific? `resizeUblockUpdate` perhaps?

src/projection/node/create-projection-node.ts
```diff
@@ -432,22 +446,26 @@ export function createProjectionNode<I>({
 
         // only on the root
         blockUpdate() {
-            this.updateBlocked = true
+            this.updateManuallyBlocked = true
         }
 
         unblockUpdate() {
-            this.updateBlocked = false
+            this.updateManuallyBlocked = false
```
nvh, line 453: Not sure, but maybe we should use `manual` somewhere in these names too?

nvh (approved):

package.json
nvh: Yeah, but what I ment is that it should be possible to pas this as an argument to the command, right?

src/projection/node/create-projection-node.ts
mattgperry: Great catch thanks

nvh: Maybe also create a test around it?

package.json
mattgperry: Ah I see, I'll have to look into it, I think it's already a little difficult to do this through package.json scripts but additionally this is like a script within a script (see the embedded quotation marks) so I'm not sure how you'd pass it through to that

PR #1271, opened 2021-09-22
Fix layout animations for elements that only change layout relative to parent

mattgperry:
Previously we weren't properly animating elements that DON'T change layout relative to the page/viewport but DO change layout relative to the parent.


dev/projection/shared-relative-new-child.html
```diff
@@ -89,19 +89,18 @@
...
-                    top: 150,
-                })
-
-                matchViewportBox(child, {
-                    bottom: 210,
-                    left: 340,
-                    right: 390,
-                    top: 160,
-                })
+                // matchViewportBox(newBox, {
```
shuangq, line 92: Why are we skipping this test? Is it no longer the expected behaviour?

mattgperry: Ooooops thanks

mattgperry: Please still pass

mattgperry: NOOOO

dev/projection/script-animate.js
```diff
@@ -33,7 +33,7 @@ Animate.createNode = (
     element,
     parent,
     options = {},
-    transition = { duration: 200, ease: () => 0.5 }
+    transition = { duration: 10, ease: () => 0.5 }
```
shuangq, line 36: Wouldn't this break some of our projection tests? I vaguely remember there are some tests with a 50ms timeout, but maybe I was wrong 😅

src/projection/node/create-projection-node.ts
```diff
@@ -807,7 +814,11 @@ export function createProjectionNode<I>({
              * a relativeParent. This will allow a component to perform scale correction
              * even if no animation has started.
              */
-            if (!this.targetDelta && !this.relativeTarget) {
+            if (
+                this.safeToResolveRelativeTarget &&
```
shuangq, line 818: Why do we need this check? After the first time it should get a relative target right? 🤔

src/projection/node/create-projection-node.ts
```diff
@@ -893,7 +907,10 @@ export function createProjectionNode<I>({
             if (!this.parent || hasTransform(this.parent.latestValues))
                 return undefined
 
-            if (this.parent.target && this.parent.layout) {
+            if (
+                (this.parent.target || this.parent.targetDelta) &&
```
shuangq, line 911: Should we just check `targetDelta`? When do we expecting `target` to be there but not `targetDelta`?

dev/projection/script-animate.js
mattgperry: That's ten seconds

shuangq: OH 🙈

src/projection/node/create-projection-node.ts
mattgperry: I think maybe if we check for relativeTarget instead it'll catch both but earlier in the frame

src/projection/node/create-projection-node.ts
mattgperry: Ah yeah I've removed it, I think this is not needed any more

shuangq (approved):

PR #1272, opened 2021-09-22
Fix: sibling animations triggered by unmount

shuangq:
Previously we fixed the sibling animation triggered by an unmounting element by calling `didUpdate` in a React effect cleanup, but that no longer works because `checkUpdateFailed` cancel the update before the effect runs.
This PR moves it into `scheduleCheckUpdateFailed` (renamed to `scheduleCheckAfterUnmount`) so that we either run `didUpdate` or clear all the snapshots.

### DIFF
--- a/src/projection/node/create-projection-node.ts
+++ b/src/projection/node/create-projection-node.ts
@@ -571,8 +571,19 @@ export function createProjectionNode<I>({
             sync.preRender(this.updateProjection, false, true)
         }
 
-        scheduleUpdateFailedCheck() {
-            sync.postRender(this.checkUpdateFailed)
+        scheduleCheckAfterUnmount() {
+            /**
+             * If the unmounting node is in a layoutGroup and did trigger a willUpdate,
+             * we manually call didUpdate to give a chance to the siblings to animate.
+             * Otherwise, cleanup all snapshots to prevents future nodes from reusing them.
+             */
+            sync.postRender(() => {
+                if (this.isLayoutDirty) {
+                    this.root.didUpdate()
+                } else {
+                    this.root.checkUpdateFailed()
+                }
+            })
         }
 
         checkUpdateFailed = () => {
--- a/Makefile
+++ b/Makefile
@@ -68,10 +68,12 @@ publish: clean bootstrap
 	npm publish
 	git push
 
-test: yarn test
+test: bootstrap
+	yarn test
 	yarn tsc dist/framer-motion.d.ts --skipLibCheck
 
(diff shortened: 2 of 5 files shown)


src/projection/node/create-projection-node.ts
```diff
@@ -566,8 +564,19 @@ export function createProjectionNode<I>({
...
-        scheduleUpdateFailedCheck() {
-            sync.postRender(this.checkUpdateFailed)
+        scheduleCheckAfterUnmount(nodeInLayoutGroup?: IProjectionNode) {
+            /**
+             * If the unmounting node is in a layoutGroup and did trigger a willUpdate,
+             * we manually call didUpdate to give a chance to the siblings to animate.
+             * Otherwise, cleanup all snapshots to prevents future nodes from reusing them.
+             */
+            sync.postRender(() => {
+                if (nodeInLayoutGroup?.isLayoutDirty) {
```
mattgperry, line 574: Currently this fires on `root` but that was only so multiple elements could schedule just one callback that removed all snapshots.

If you're doing it this way I think you can just call `projection.schedueCheckAfterUnmount`

src/projection/node/create-projection-node.ts
```diff
@@ -399,9 +399,8 @@ export function createProjectionNode<I>({
                         hasLayoutChanged,
                         layout: newLayout,
                     }: LayoutUpdateData) => {
-                        const dragControls = elementDragControls.get(
-                            visualElement
-                        )
+                        const dragControls =
+                            elementDragControls.get(visualElement)
```
shuangq, line 403: @mattgperry hmm now the precommit hook starts to introduce these formatting changes, do you have any idea why?

mattgperry (approved):

src/projection/node/create-projection-node.ts
mattgperry: No its weird I wonder if Prettier has had a quiet update

mattgperry:
Does the full test suite pass locally?

src/projection/node/create-projection-node.ts
shuangq: I didn't quite understand, don't we want to same for `didUpdate` if multiple elements schedule it?

shuangq:
> Does the full test suite pass locally?

Yeah it does

src/projection/node/create-projection-node.ts
mattgperry: That's a good point but in this case didUpdate could be fired by an unlimited number of elements

Issue #1276, opened 2021-09-24
[BUG] Example link in readme is not properly linked to right URL.


dntandan:
Readme file has one section where one can check the examples under the docs section with a wrong URL.

![image](https://user-images.githubusercontent.com/32794634/134728646-54902aa7-2038-4409-9e42-84e3a1d05b44.png)

Current URL: https://framer.com/motion 
Right URL: https://www.framer.com/examples

dntandan:
Since Issue has been resolved, I am closing this PR now.

PR #1277, opened 2021-09-24
Corrected example URL in readme

dntandan:
Solves #1276


mattgperry:
That page is for Framer rather than Framer Motion but I will update the link to the Motion examples page, good call

PR #1280, opened 2021-09-29
Fix: check for relative target change even if the new update has the same layout

shuangq:
When an update happens during an existing animation, the new layout may be the same as the old one, but the element's position relative to its parent can change. This PR checks `hasRelativeTargetChanged` regardless of `hasLayoutChanged` to make sure `targetChanged` is correct.

This PR also changes the default cypress browser to chrome, as there's a rendering bug in the default one (electron 78)

Fixes: https://github.com/framer/company/issues/22986


cypress/integration/drag-to-reorder.ts
```diff
@@ -20,18 +20,18 @@ describe("Drag to reorder", () => {
             .get("#Tomato")
             .should(([$item]: any) => {
                 expectBbox($item, {
-                    height: 68,
+                    height: 67,
```
shuangq, line 23: This test has to be updated because of the default browser change.

mattgperry (approved):
If you look on line ~1500 of `create-projection-node` it should be that `hasRelativeLayoutChange` is already essentially `hasOnlyRelativeLayoutChanged`. This looks good but I'm not clear how the overall logic has changed. But the proof is in the tests!

PR #1281, opened 2021-09-29
Fix snapshot in scrolled & scaled parent

mattgperry:
Snapshots were being recorded incorrectly within one parent that was scrolled and another that was a scaled.

I suspect this was from an improperly computed `originPoint`. Switching the order we removed scroll & transform fixes it, I'm not 100% sure why but the measured viewport box is returned with the page scroll already removed so it makes sense to remove the element scroll next and then transforms.

### DIFF
--- a/src/projection/node/create-projection-node.ts
+++ b/src/projection/node/create-projection-node.ts
@@ -608,20 +608,7 @@ export function createProjectionNode<I>({
                 this.removeElementScroll(measured)
             )
             roundBox(layout)
-            // if (this.instance.id === "box")
-            //     console.log(
-            //         "box measured as",
-            //         measured.y,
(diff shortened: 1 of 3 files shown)


shuangq (approved):

src/projection/node/create-projection-node.ts
```diff
@@ -604,8 +604,8 @@ export function createProjectionNode<I>({
         updateSnapshot() {
             if (this.snapshot || !this.instance) return
             const measured = this.measure()!
-            const layout = this.removeElementScroll(
-                this.removeTransform(measured)
+            const layout = this.removeTransform(
```
shuangq, line 607: We could maybe put a comment here in case the order is changed by mistake later?

PR #1288, opened 2021-10-12
fix iframe events and measurements

### LINKED ISSUE #379: [BUG] Render into iframe
When a `motion.div` (or any `motion` component) is rendered into an iframe, the following invariant fails:

```
invariant(ref.current instanceof Element, "No `ref` found. Ensure components created with `motion.custom` forward refs using `React.forwardRef`");
```

Although `ref.current` is also an `Element`, it's an `Element` from another `window`. Thus it doesn't belong to the same prototype chain as the one referenced in above code`instanceof Element`.

A possible solution is to use `ref.current.constructor.name`? Or remove the invariance?

https://codesandbox.io/s/framer-motion-iframe-bug-x7ud9

souporserious:
This PR adds the ability to specify which `document` and `window` context to use by adding two new options to `MotionConfig`. This is most useful when needing to wrap motion components in an iframe and making sure events are attached and measured based on the iframe's `document` and `window`.

I've added a simple example that showcases how this works with `useViewportScroll`. Please let me know if there are any other areas that may be using `document` or `window` and I can get those updated 😃 thanks!

Fixes #379

### DIFF
--- a/src/context/MotionConfigContext.tsx
+++ b/src/context/MotionConfigContext.tsx
@@ -27,18 +27,11 @@ export interface MotionConfigContext {
     transition?: Transition
 
     /**
-     * Defines the document context to use for the entire tree.
-     *
-     * @public
-     */
-    documentContext: typeof document
-
-    /**
      * Defines the window context to use for the entire tree.
      *
      * @public
      */
-    windowContext: typeof window
+    windowContext: typeof window | null
 }
 
 /**
@@ -47,6 +40,5 @@ export interface MotionConfigContext {
 export const MotionConfigContext = createContext<MotionConfigContext>({
     transformPagePoint: (p) => p,
     isStatic: false,
-    documentContext: document,
-    windowContext: window,
+    windowContext: typeof window === "undefined" ? null : window,
 })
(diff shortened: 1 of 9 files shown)


src/context/MotionConfigContext.tsx
```diff
@@ -25,6 +25,20 @@ export interface MotionConfigContext {
      * @public
      */
     transition?: Transition
+
+    /**
+     * Defines the document context to use for the entire tree.
+     *
+     * @public
+     */
+    documentContext: typeof document
```
souporserious, line 34: Curious your thoughts on whether this should live in its own context provider or if it's ok in `MotionConfig`? Also, happy to change this if there's a better name or they should be in a single object rather than two separate properties.

src/context/MotionConfigContext.tsx
```diff
@@ -33,4 +47,6 @@ export interface MotionConfigContext {
 export const MotionConfigContext = createContext<MotionConfigContext>({
     transformPagePoint: (p) => p,
     isStatic: false,
+    documentContext: document,
+    windowContext: window,
```
mattgperry, line 51: I think this will fail server-side - you could add MotionConfig to ssr.test.ts to see if it renders ok

souporserious:
Closing this since I think using an iframe and communicating through URL params or message passing is the proper way to handle this rather than and creating and injecting the library code into an iframe requiring weird APIs like this PR was trying to add.

PR #1289, opened 2021-10-13
Fixes for animating between sibling scroll contexts

mattgperry:
For https://github.com/framer/company/issues/23054

### DIFF
--- a/src/projection/node/create-projection-node.ts
+++ b/src/projection/node/create-projection-node.ts
@@ -751,7 +751,7 @@ export function createProjectionNode<I>({
         applyTransform(box: Box, transformOnly = false): Box {
             const withTransforms = createBox()
             copyBoxInto(withTransforms, box)
-            for (let i = 1; i < this.path.length; i++) {
+            for (let i = 0; i < this.path.length; i++) {
                 const node = this.path[i]
 
                 if (


src/projection/node/create-projection-node.ts
```diff
@@ -748,16 +748,32 @@ export function createProjectionNode<I>({
...
+                        y: -node.scroll.y,
+                    })
+                }
+
                 if (!hasTransform(node.latestValues)) continue
                 transformBox(withTransforms, node.latestValues)
             }
 
-            if (hasTransform(this.latestValues)) {
+            if (!scrollOnly && hasTransform(this.latestValues)) {
```
shuangq, line 776: I feel it makes the function a bit too complicated to keep track of what we should/should not do when scroll only, should we create a new helper for it?

dev/projection/shared-scroll-b-a.html
```diff
@@ -0,0 +1,94 @@
...
+            boxProjection.root.didUpdate()
+
+            setTimeout(() => {
+                matchViewportBox(box, boxOrigin)
+                matchViewportBox(newBox, boxOrigin)
+                matchViewportBox(box, newBox)
+            }, 50)
+        </script>
+    </body>
+</html>
```
shuangq, line 94: should we also cover a test between two scroll screens?

src/projection/node/create-projection-node.ts
```diff
@@ -748,11 +748,24 @@ export function createProjectionNode<I>({
             return boxWithoutScroll
         }
 
-        applyTransform(box: Box): Box {
+        applyTransform(box: Box, transformOnly = false): Box {
             const withTransforms = createBox()
             copyBoxInto(withTransforms, box)
-            for (let i = 0; i < this.path.length; i++) {
+            for (let i = 1; i < this.path.length; i++) {
```
shuangq, line 754: Why do we skip the parent? Can we add a comment here to clarify?

mattgperry: I think this was the root node but I've put it back to 0 as it doesn't really make a difference and is clearer

PR #1293, opened 2021-10-18
Snap to cursor

mattgperry:
Reimplementing snap to cursor with added tests


src/gestures/drag/VisualElementDragControls.ts
```diff
@@ -83,19 +83,14 @@ export class VisualElementDragControls {
...
 
-            /**
-             * Save the initial point. We'll use this to calculate the pointer's position rather
-             * than the one we receive when the gesture actually starts. By then, the pointer will
-             * have already moved, and the perception will be of the pointer "slipping" across the element
-             */
-            initialPoint = getViewportPointFromEvent(event).point
+            if (snapToCursor) {
+                this.snapToCursor(extractEventInfo(event, "page").point)
+            }
```
shuangq, line 93: It probably doesn't make much of a difference, but should we snap only after the drag actually starts (in `onStart`)?

mattgperry: It was this way before but also the feel of it is quite different, a bit like how we stop existing animations on mouse down not mouse move

PR #1298, opened 2021-10-20
Disable `pointer-events` for follow components

mattgperry:
Otherwise if a follow component has a stacking context higher than the lead, it will block pointer events

### DIFF
--- a/src/projection/node/create-projection-node.ts
+++ b/src/projection/node/create-projection-node.ts
@@ -1437,6 +1437,11 @@ export function createProjectionNode<I>({
                 }
             }
 
+            /**
+             * Disable pointer events on follow components. This is to ensure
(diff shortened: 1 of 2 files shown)


src/projection/node/create-projection-node.ts
```diff
@@ -1435,6 +1436,12 @@ export function createProjectionNode<I>({
                     styles[key] = corrected
                 }
             }
+
+            if (this.options.layoutId) {
+                styles.pointerEvents =
+                    lead === this ? styleProp.pointerEvents || "" : "none"
```
shuangq, line 1442: I think we could add the description of this PR here as a comment to clarify.

shuangq (approved):

Issue #503, opened 2020-04-02
[FEATURE] Dynamic controls.start should accept variant labels


zeckdude:
**1. Read the [FAQs](#faqs) 👇**

**2. Describe the bug**

When passing custom property to control, you can't return a variant string. I am sending a value via the `custom` prop so I can use that to affect logic in the control as to what styles to apply. However, I want to use a variant string instead of an object with the styles. It doesn't like that.

```js
const ImperativeAnimationsButtonDemo = () => {
  const controls = useAnimation();

  const variants = {
    whenTrue: {
      backgroundColor: '#FFC0CB' // pink
    },
    whenFalse: {
      backgroundColor: '#0000FF' // blue
    },
  }

  useEffect(() => {
    controls.start(isTrue => isTrue ? 'whenTrue' : 'whenFalse');
  }, [])

  return (
    <motion.button
      variants={variants}
      custom={true}
      animate={controls}
    >
      Button
    </motion.button>
  );
}
```

This produces the following error: 
> You are trying to animate 0 from "0" to w. "0" is not an animatable value - to enable this animation set 0 to a value animatable to w via the `style` property.

> You are trying to animate 1 from "0" to h. "0" is not an animatable value - to enable this animation set 0 to a value animatable to h via the `style` property.

> You are trying to animate 2 from "0" to e. "0" is not an animatable value - to enable this animation set 0 to a value animatable to e via the `style` property.

> You are trying to animate 3 from "0" to n. "0" is not an animatable value - to enable this animation set 0 to a value animatable to n via the `style` property.

> You are trying to animate 4 from "0" to T. "0" is not an animatable value - to enable this animation set 0 to a value animatable to T via the `style` property.

> You are trying to animate 5 from "0" to r. "0" is not an animatable value - to enable this animation set 0 to a value animatable to r via the `style` property.

> You are trying to animate 6 from "0" to u. "0" is not an animatable value - to enable this animation set 0 to a value animatable to u via the `style` property.

> You are trying to animate 7 from "0" to e. "0" is not an animatable value - to enable this animation set 0 to a value animatable to e via the `style` property.

> Error: Failed to set an indexed property on 'CSSStyleDeclaration': Index property setter is not supported.

**3. IMPORTANT: Provide a CodeSandbox reproduction of the bug**

https://stackblitz.com/edit/framer-motion-imperative-return-variant

**4. Steps to reproduce**

1. Open the stackblitz link

**5. Expected behavior**

[...]

PR #1323, opened 2021-11-01
Allow `controls.start()` dynamic functions to return variant names

mattgperry:
Fixes https://github.com/framer/motion/issues/503

### DIFF
--- a/src/render/utils/variants.ts
+++ b/src/render/utils/variants.ts
@@ -58,6 +58,9 @@ export function resolveVariantFromProps(
     currentValues: ResolvedValues = {},
     currentVelocity: ResolvedValues = {}
 ) {
+    /**
+     * If the variant definition is a function, resolve.
+     */
     if (typeof definition === "function") {
         definition = definition(
             custom ?? props.custom,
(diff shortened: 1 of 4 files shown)


src/render/utils/variants.ts
```diff
@@ -58,13 +58,39 @@ export function resolveVariantFromProps(
...
+    /**
+     * If the variant definition is a function, resolve.
+     */
+    if (typeof definition === "function") {
+        definition = definition(
+            custom ?? props.custom,
+            currentValues,
+            currentVelocity
+        )
+    }
```
nvh, line 70: As this is now in there twice, should we pull it into a `resolve` function?

nvh (approved):

src/render/utils/variants.ts
mattgperry: I did think about reducing duplication but if anything we'd probably just be padding out the file given that the only real duplication is the typeof check

nvh: Ah, right, that makes sense.
Well, if you name the function right you wouldn't need the comment and there's the `??` but I see your point.

Issue #521, opened 2020-04-30
pathLength doens't work on non-scaling-stroke [BUG]


ShanonJackson:
**2. Describe the bug**
path length should animate even on non-scaling-strokes; My suspicion is you are calculating the line length using the DOM method `.getTotalLength()` however this returns the incorrect value for strokes with `non-scaling-stroke` this css value is extremely important to people creating graphs that are horizontally responsive without skew'ing the svg paths.

**3. IMPORTANT: Provide a CodeSandbox reproduction of the bug**
https://codesandbox.io/s/kind-knuth-b1ed2?file=/src/index.tsx
Let me know if this doesn't work don't use code sandbox often.

**5. Expected behavior**
Line shouldn't have gaps.

Refactoring out of react-spring love the library so this is a blocker for me atm but other than that had no problems with library.

EDIT: Possible solution outlined here:
https://stackoverflow.com/questions/44389668/svg-gettotallength-wont-return-the-right-size-for-non-scaling-stroke-get-sv

PR #1325, opened 2021-11-02
Enable pathLength animations for more SVG shapes

mattgperry:
Instead of measuring path length with `getTotalLength`, we now set the [`pathLength` attribute](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/pathLength) to `1`.

This change:
- Fixes path animations for elements with `vector-effect="non-scaling-stroke"` applied
- Extends path animations from `path` to `circle`, `ellipse`, `line`, `polygon`, `polyline` and `rect` components
- Adds SSR support for `pathLength`

Fixes https://github.com/framer/motion/issues/521

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,12 +2,17 @@
 
 Framer Motion adheres to [Semantic Versioning](http://semver.org/).
 
+## [5.1.0] 2021-11-02
+
+### Added
+
+-   Adding path drawing support for `circle`, `ellipse`, `line`, `path`, `polygon`, `polyline` and `rect` components.
+
 ### Fixed
 
+-   Fixed SSR for `pathLength`.
 -   Downgrading `whileFocus` to lowest gesture priority. [Issue](https://github.com/framer/motion/issues/1221)
-
-### Fixed
-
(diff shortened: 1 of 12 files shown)


willhindson:
I'm not sure this is the best way to get my issue fixed, but I think that a problem I am having aligns mostly to this previously flagged issue.

I'm using framer-motion to animate a simple path with a start point, end point and two control points from 0 to 1 in length.

the animation works smoothly but when I add `vectorEffect="non-scaling-stroke"` on `motion.path` (which is necessary for this stretched svg) you can clearly see the strokeDasharray — the line only gets half filled. As though the pathLength is not being correctly calculated

It seems from this issue that it should have been solved long ago, but maybe something was broken recently?

willhindson:
I am wondering if the issue is related to the fact that the path animates from `stroke-dasharray="0px 1px"` to `stroke-dasharray="1px 1px"`

Nickersoft:
Can confirm I'm also seeing this!

willhindson:
Any updates on a potential fix? Would be very appreciative 🙏

PR #1326, opened 2021-11-02
add dragSnapToOrigin to valid props

Evalon:
It clears react warning in dev console for dragSnapToOrigin (Reorder.Item)


mattgperry:
Ah great catch thanks!

PR #1330, opened 2021-11-03
fix: check the existence of  Proxy

ambar:
This is useful when importing the package in legacy browsers (especially `umd`).

Related: #1178
Closes: https://github.com/framer/motion/issues/1178


mattgperry:
Nice approach, this would still break in older browsers but would at least allow people who need this to do

```
const Component = motion("div")
```

ambar:
Also handy for calling `m()`. 

In current version we cannot construct `m() ` because `createDomMotionConfig` is not exported.

mattgperry (approved):

PR #1333, opened 2021-11-03
Adding check for mixing layout animation types

mattgperry:
This now checks for 

- Layout animations happening within layout-affecting animations
- Layout-affecting animations happening within layout animations

### DIFF
--- a/src/projection/node/create-projection-node.ts
+++ b/src/projection/node/create-projection-node.ts
@@ -1699,9 +1699,9 @@ function roundBox(box: Box): void {
  * layout-affecting animation.
  */
 function isTreeAnimatingLayoutAffectingStyle(
-    visualElement: VisualElement
+    visualElement: VisualElement | undefined
 ): boolean {
-    while (visualElement.parent) {
+    while (visualElement) {
         if (Boolean(visualElement.layoutAffectingAnimations)) return true
         visualElement = visualElement.parent
     }
--- a/src/projection/node/create-projection-node.ts
+++ b/src/projection/node/create-projection-node.ts
@@ -41,7 +41,7 @@ import { FlatTree } from "../../render/utils/flat-tree"
 import { Transition } from "../../types"
 import { resolveMotionValue } from "../../value/utils/resolve-motion-value"
 import { MotionStyle } from "../../motion/types"
-import { warnOnce } from "../../utils/warn-once"
+import { hasWarned, warnOnce } from "../../utils/warn-once"
 
 /**
  * We use 1000 as the animation target as 0-1000 maps better to pixels than 0-1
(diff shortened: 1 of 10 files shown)


nvh:
Works! https://codesandbox.io/s/framer-motion-simple-animation-forked-spq0l?file=/src/index.tsx

Is there any way we can be more specific of the element that is performing the bad behaviour? Especially the `animate` call?

src/projection/node/create-projection-node.ts
```diff
@@ -1161,6 +1162,27 @@ export function createProjectionNode<I>({
             this.pendingAnimation = sync.update(() => {
                 globalProjectionState.hasAnimatedSinceResize = true
 
+                /**
+                 * Check if this component or any of its ancestors are animating
+                 */
+                if (
+                    process.env.NODE_ENV !== "production" &&
```
nvh, line 1169: Smart! But wouldn't this mean that it does not work inside of production builds of FramerStudio?

src/projection/node/create-projection-node.ts
```diff
@@ -1677,3 +1699,18 @@ function roundBox(box: Box): void {
...
+ * Look up the tree and check whether any element is performing a
+ * layout-affecting animation.
+ */
+function isTreeAnimatingLayoutAffectingStyle(
+    visualElement: VisualElement | undefined
+): boolean {
+    while (visualElement) {
+        if (Boolean(visualElement.layoutAffectingAnimations)) return true
+        visualElement = visualElement.parent
+    }
```
nvh, line 1713: What's the performance impact of this?
Would it help to memoize it?

src/projection/node/create-projection-node.ts
```diff
@@ -1161,6 +1162,27 @@ export function createProjectionNode<I>({
...
+                    /**
+                     * We need to fire this in a setTimeout because layout animations will start
+                     * before any layout-affecting animations.
+                     */
+                    setTimeout(() => {
+                        warnOnce(
+                            !isTreeAnimatingLayoutAffectingStyle(
+                                this.options.visualElement
+                            ),
+                            `Attempting to animate layout within a component performing an animation on a layout-affecting properties (e.g. "width", "height", "top"). This is likely to break layout animations. Attempt to replace with the layout and style prop, e.g. <motion.div layout style={{ width: 100px }} />.`
```
nvh, line 1181: I think logging out the actual div that is at fault here would be helpful for debugging.

src/render/utils/animation.ts
```diff
@@ -167,6 +169,26 @@ function animateTarget(
             ...transition,
         })
 
+        /**
+         * Layout animations rely on synchronous rendering of layout, which
+         * is incompatible with animating layout-affecting properties.
+         * Here we check if this component or any above it are animating layout
+         * and warn if an animation is started on some popular layout-effecting properties.
```
nvh, line 176: Only the popular ones, or all of them?

src/render/utils/animation.ts
```diff
@@ -167,6 +169,26 @@ function animateTarget(
...
+         */
+        if (process.env.NODE_ENV !== "production") {
+            if (layoutFlags[key]) {
+                warnOnce(
+                    !visualElement.projection?.isTreeAnimating,
+                    `Attempting to animate layout-affecting style "${key}" to ${valueTarget} on a component that is performing a layout animation (e.g. <motion.div layout />). This is likely to break layout animations. Attempt to replace with <motion.div layout style={{ ${key}: ${valueTarget} }} />.`
```
nvh, line 182: Here I think we should also log the DOM element

nvh (commented):
Let's make sure this works inside of FramerStudio before merging

src/projection/node/create-projection-node.ts
mattgperry: Yes, to cut out the overhead. We don't throw any warnings in prod

src/projection/node/create-projection-node.ts
mattgperry: Actually I suppose because `warnOnce` is only going to fire once, this doesn't need to be checked a second time. So I can change it so the boolean passed to `warnOnce` is actually a resolver function that only gets run if we haven't already fired the warning

nvh: Well, depends if we want to include the DOM element for which this happens, right?

src/render/utils/animation.ts
mattgperry: Just the popular ones, I don't think we should add paddingLeft etc unless this comes up

nvh (approved):
Just (finally) verified that this actually warns when using overrides animating together with layout: https://feature-motion-warning.beta.development.framer.com/projects/Animate-layout-warning--4eAAz2zuQHBnYOl6IgpY-1hfZe?node=erAxaRbVM

mattgperry:
I'm closing this out for now as I don't think this has really come up much?

Issue #1336, opened 2021-11-05
[BUG] Using CSS variables doesn't trigger animation (version 5.0.2 and above)


xeho91:
I am using CSS variables to get the position `x` value **based on the element's width**.
In the version `5.0.1`, this was working fine, without any issues.

However, after I **upgraded `framer-motion`** to version `5.0.2` and eventually to `5.2.0`, I have noticed that the animations suddenly stopped working _(they're not triggered at all)_.

## Minimal reproduction Sandbox
https://codesandbox.io/s/brave-rhodes-ld5bs?file=/src/App.js

You can see that this pattern works fine in version `5.0.1`, however in more recent versions, it doesn't work.
Is it expected behaviour, or was there any breaking change that I missed?

mattgperry:
There's nothing jumping out at me in the changelogs, I'll take a look

PR #1337, opened 2021-11-05
Fixing unit conversion for translation

mattgperry:
Fixes https://github.com/framer/motion/issues/1336

### DIFF
--- a/src/render/dom/utils/unit-conversion.ts
+++ b/src/render/dom/utils/unit-conversion.ts
@@ -4,7 +4,7 @@ import { isKeyframesTarget } from "../../../animation/utils/is-keyframes-target"
 import { invariant } from "hey-listen"
 import { MotionValue } from "../../../value"
 import { transformProps } from "../../html/utils/transform"
-import { VisualElement } from "../../types"
+import { ResolvedValues, VisualElement } from "../../types"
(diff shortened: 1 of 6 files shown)


src/render/dom/utils/unit-conversion.ts
```diff
@@ -140,10 +148,8 @@ const convertChangedValueTypes = (
         // Restore styles to their **calculated computed style**, not their actual
         // originally set style. This allows us to animate between equivalent pixel units.
         const value = visualElement.getValue(key) as MotionValue
-        setAndResetVelocity(
-            value,
-            positionalValues[key](originBbox, elementComputedStyle)
-        )
+
+        setAndResetVelocity(value, origin[key])
```
nvh, line 152: Can you explain why this would fix it?

nvh (approved):

src/render/dom/utils/unit-conversion.ts
mattgperry: Yeah its in the comment in the previous line - computed style is live so we need to grab these before we trigger a render from it.

nvh: So how did this regress then?

PR #1344, opened 2021-11-09
whileInView

mattgperry:
This PR adds `whileInView`, `onViewportEnter` and `onViewportLeave` props.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "framer-motion",
-    "version": "5.3.0-rc.1",
+    "version": "5.2.1",
     "description": "A simple and powerful React animation library",
     "main": "dist/framer-motion.cjs.js",
     "module": "dist/es/index.mjs",
@@ -37,7 +37,7 @@
         "test-ci": "yarn test-client && yarn test-server",
         "test-client": "jest --coverage --config jest.config.json --max-workers=2",
         "test-server": "jest --config jest.config.ssr.json",
-        "test-projection": "yarn run collect-projection-tests && start-server-and-test 'python -m SimpleHTTPServer' http://localhost:8000 'yarn run cypress run -s cypress/integration/projection.chrome.ts --config baseUrl=http://localhost:8000/'",
+        "test-projection": "yarn run collect-projection-tests && start-server-and-test 'python -m SimpleHTTPServer' http://0.0.0.0:8000 'yarn run cypress run -s cypress/integration/projection.chrome.ts --config baseUrl=http://localhost:8000/'",
         "test-e2e-chrome": "start-server-and-test start-dev-server http://localhost:9990 'cypress run --headless --browser chrome  --spec \"cypress/integration/layout-relative.chrome.ts\"'",
         "test-e2e-electron": "start-server-and-test start-dev-server http://localhost:9990 'cypress run --headless --config ignoreTestFiles=*.chrome.ts'",
         "test-e2e": "yarn test-e2e-chrome && yarn test-e2e-electron",
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
         "test-e2e-chrome": "start-server-and-test start-dev-server http://localhost:9990 'cypress run --headless --browser chrome  --spec \"cypress/integration/layout-relative.chrome.ts\"'",
         "test-e2e-electron": "start-server-and-test start-dev-server http://localhost:9990 'cypress run --headless --config ignoreTestFiles=*.chrome.ts'",
         "test-e2e": "yarn test-e2e-chrome && yarn test-e2e-electron",
-        "test-e2e-file": "start-server-and-test start-dev-server http://localhost:9990 'cypress run --headless --spec \"cypress/integration/drag-framer-page.ts\"'",
+        "test-e2e-file": "start-server-and-test start-dev-server http://localhost:9990 'cypress run --headless --spec \"cypress/integration/while-in-view.ts\"'",
         "test-watch": "jest --watch --coverage --coverageReporters=lcov  --config jest.config.json",
         "projection-dev": "python -m SimpleHTTPServer",
         "collect-projection-tests": "node ./dev/projection/collect-projection-tests.js",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "framer-motion",
-    "version": "5.2.1",
+    "version": "5.3.0-rc.1",
     "description": "A simple and powerful React animation library",
     "main": "dist/framer-motion.cjs.js",
     "module": "dist/es/index.mjs",
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
         "test-ci": "yarn test-client && yarn test-server",
         "test-client": "jest --coverage --config jest.config.json --max-workers=2",
         "test-server": "jest --config jest.config.ssr.json",
-        "test-projection": "yarn run collect-projection-tests && start-server-and-test 'python -m SimpleHTTPServer' http://localhost:8000 'yarn run cypress run -s cypress/integration/projection.chrome.ts --config baseUrl=http://0.0.0.0:8000/'",
+        "test-projection": "yarn run collect-projection-tests && start-server-and-test 'python -m SimpleHTTPServer' http://localhost:8000 'yarn run cypress run -s cypress/integration/projection.chrome.ts --config baseUrl=http://localhost:8000/'",
         "test-e2e-chrome": "start-server-and-test start-dev-server http://localhost:9990 'cypress run --headless --browser chrome  --spec \"cypress/integration/layout-relative.chrome.ts\"'",
         "test-e2e-electron": "start-server-and-test start-dev-server http://localhost:9990 'cypress run --headless --config ignoreTestFiles=*.chrome.ts'",
         "test-e2e": "yarn test-e2e-chrome && yarn test-e2e-electron",
--- a/package.json
+++ b/package.json
@@ -136,31 +136,31 @@
     "bundlesize": [
         {
             "path": "./dist/framer-motion.js",
-            "maxSize": "34.1 kB"
+            "maxSize": "34.7 kB"
         },
         {
             "path": "./dist/size-rollup-m.js",
-            "maxSize": "6.5 kB"
+            "maxSize": "6.1 kB"
         },
         {
             "path": "./dist/size-rollup-dom-animation.js",
-            "maxSize": "17.2 kB"
+            "maxSize": "17.7 kB"
         },
         {
             "path": "./dist/size-rollup-dom-max.js",
-            "maxSize": "28.2 kB"
+            "maxSize": "28.7 kB"
         },
         {
             "path": "./dist/size-webpack-m.js",
-            "maxSize": "6.5 kB"
+            "maxSize": "6.3 kB"
         },
         {
             "path": "./dist/size-webpack-dom-animation.js",
-            "maxSize": "19 kB"
+            "maxSize": "19.3 kB"
         },
         {
             "path": "./dist/size-webpack-dom-max.js",
-            "maxSize": "30.5 kB"
+            "maxSize": "31 kB"
         }
     ]
 }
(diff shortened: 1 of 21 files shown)


src/motion/features/viewport/types.ts
```diff
@@ -0,0 +1,19 @@
...
+import { TargetAndTransition } from "../../../types"
+import { VariantLabels } from "../../types"
+
+export type ViewportEventHandler = () => void
+
+export interface ViewportOptions {
+    root?: RefObject<Element>
+    once?: boolean
+    margin?: string
+    amount?: "some" | "all"
```
nvh, line 11: In `useAppearEffect` we https://github.com/framer/FramerStudio/blob/d6077be8587602ac1a2e8eb901da57a91aaffc30/src/library/src/modules/useAppearEffect.ts#L82-L86, but only pass it 0, 0.5, and 1 as value.

Can we add a `half` or something similar to this?

src/motion/features/viewport/use-viewport.ts
```diff
@@ -0,0 +1,90 @@
...
+                const callback = isIntersecting
+                    ? props.onViewportEnter
+                    : props.onViewportLeave
+                callback?.()
+            }
+        )
+    }, [shouldObserve, root, rootMargin, amount])
+}
+
+function useMissingIntersectionObserver(
```
shuangq, line 75: should we throw a warning if it's not supported?

src/motion/features/viewport/use-viewport.ts
```diff
@@ -0,0 +1,90 @@
...
+            visualElement.getInstance(),
+            {
+                root: root?.current,
+                rootMargin,
+                threshold: amount === "some" ? 0 : 1,
+            },
+            (entry) => {
+                const { isIntersecting } = entry
+
+                if (once && !isIntersecting && hasEnteredView.current) {
```
shuangq, line 50: shouldn't this trigger `onViewportLeave`?

src/motion/features/viewport/use-viewport.ts
mattgperry: Yeah good idea!

src/motion/features/viewport/use-viewport.ts
mattgperry: I'm not sure, but if `once` affects the callbacks then people get that consistent behaviour baked in. It's quite nice for when you wanna trigger other stuff on the first viewport enter

src/motion/features/viewport/types.ts
mattgperry: Sure, I wanted to avoid the fractional values for the reasons you were telling me about the other day, but perhaps we could also accept a number as an undocumented type

src/motion/features/viewport/observers.ts
```diff
@@ -0,0 +1,71 @@
...
+    root,
+    ...options
+}: IntersectionObserverInit): IntersectionObserver {
+    /**
+     * If we don't have an observer lookup map for this root, create one.
+     */
+    if (!observers.has(root || document)) {
+        observers.set(root || document, {})
+    }
+    const rootObservers = observers.get(root || document)!
```
nvh, line 42: Why don't we assign `root || document` to a constant here?
And shouldn't it be `root ?? document`?

nvh: And this should be called `intersectionObservers`, right?

src/motion/features/viewport/observers.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    const key = JSON.stringify(options)
+
+    /**
+     * If we don't have an observer for this combination of root and settings,
+     * create one.
+     */
+    if (!rootObservers[key]) {
```
nvh, line 50: Because the root is not part of the key, we need to check that the root is actually the same as the existing intersectionObserver, like we do [here](https://github.com/framer/FramerStudio/blob/d6077be8587602ac1a2e8eb901da57a91aaffc30/src/library/src/modules/useAppearEffect.ts#L63)

src/motion/features/viewport/observers.ts
```diff
@@ -0,0 +1,71 @@
...
+    element: Element,
+    options: IntersectionObserverInit,
+    callback: IntersectionHandler
+) {
+    const rootInteresectionObserver = initIntersectionObserver(options)
+
+    observerCallbacks.set(element, callback)
+    rootInteresectionObserver.observe(element)
+
+    return () => rootInteresectionObserver.unobserve(element)
```
nvh, line 70: I would say that this shouldn't be part of creating the observer, and it leads to a bit weird code where we return the result from this in useEffect. I don't feel strongly, but shouldn't we pull this out?

nvh: We wrapped the IntersectionObserver in a separate class for this [here](https://github.com/framer/FramerStudio/blob/25d3c81abbc4638e84e371cc9c62883b6c72142c/src/library/src/modules/withVariantAppearEffect.tsx#L5)

package.json
```diff
@@ -37,11 +37,11 @@
...
+        "test-projection": "yarn run collect-projection-tests && start-server-and-test 'python -m SimpleHTTPServer' http://localhost:8000 'yarn run cypress run -s cypress/integration/projection.chrome.ts --config baseUrl=http://localhost:8000/'",
         "test-e2e-chrome": "start-server-and-test start-dev-server http://localhost:9990 'cypress run --headless --browser chrome  --spec \"cypress/integration/layout-relative.chrome.ts\"'",
         "test-e2e-electron": "start-server-and-test start-dev-server http://localhost:9990 'cypress run --headless --config ignoreTestFiles=*.chrome.ts'",
         "test-e2e": "yarn test-e2e-chrome && yarn test-e2e-electron",
-        "test-e2e-file": "start-server-and-test start-dev-server http://localhost:9990 'cypress run --headless --spec \"cypress/integration/drag-framer-page.ts\"'",
+        "test-e2e-file": "start-server-and-test start-dev-server http://localhost:9990 'cypress run --headless --spec \"cypress/integration/while-in-view.ts\"'",
```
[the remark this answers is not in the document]
mattgperry: " \"cypress/integration/drag-framer-page.ts\"'"," isn't a parameter someone should have to type in imo

src/motion/features/viewport/use-viewport.ts
```diff
@@ -0,0 +1,90 @@
...
+
+function useMissingIntersectionObserver(
+    shouldObserve: boolean,
+    hasEnteredView: MutableRefObject<boolean>,
+    visualElement: VisualElement
+) {
+    useEffect(() => {
+        if (!shouldObserve) return
+
+        requestAnimationFrame(() => {
```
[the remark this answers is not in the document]
mattgperry: I'll add a comment

src/motion/features/viewport/use-viewport.ts
```diff
@@ -0,0 +1,90 @@
...
+            }
+        )
+    }, [shouldObserve, root, rootMargin, amount])
+}
+
+function useMissingIntersectionObserver(
+    shouldObserve: boolean,
+    hasEnteredView: MutableRefObject<boolean>,
+    visualElement: VisualElement
+) {
```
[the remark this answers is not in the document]
mattgperry: Animates to "in view" on mount for people  without `IntersectionObserver`

src/motion/features/viewport/observers.ts
mattgperry: I don't see why it'd be `??` but I can assign to a var

src/motion/features/viewport/observers.ts
nvh: What about:

```typescript
export function createIntersectionObserver(
    options: IntersectionObserverInit,
    callback: IntersectionHandler
) {
    const rootInteresectionObserver = initIntersectionObserver(options)

    const observe = (element: Element) => {
        observerCallbacks.set(element, callback)
        rootInteresectionObserver.observe(element)
    }
    const unobserve = (element: Element) => {
        rootInteresectionObserver.unobserve(element)
    }

    return { observe, unobserve }
}
```

and then in `useIntersectionObserver` we can do:

```typescript
useEffect(() => {
    if (!shouldObserve) return

    const {observe, unobserve} = createIntersectionObserver(...)
    const element = visualElement.getInstance()
    observe(element)
    return () => unobserve(element)
})
```

src/motion/features/viewport/observers.ts
mattgperry: We've already selected from a subset of intersection observers for this root

src/motion/features/viewport/observers.ts
mattgperry: The name was chosen as it's specifically the subset of intersection observers we're storing for this particular root

nvh: Ah, I somehow completely missed that we're using a object keyed by root here 🤦🏻

src/motion/features/viewport/observers.ts
mattgperry: I'm a big believer in subscribers returning unsubscriptors and it's quite a common pattern in Motion. For instance `value.onChange`, `SubscriptionManager` and so on. I don't see the need for a wrapper class?

package.json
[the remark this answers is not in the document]
nvh: No, but, the parameter would be more like `drag-framer-page`, but this is also not really important. I'll shut up about it 😉

src/motion/features/viewport/observers.ts
nvh: Yeah I can see that and agree. Maybe it's more about the naming, I wouldn't expect a function that is called `create...Observer` to also start to observe something.
Ok, last try:

useEffect(() => {
    if (!shouldObserve) return
    const observe = createIntersectionObserver(...)
    return observe(visualElement.getInstance())
})

As I said, I don't feel strongly, so do what you think is best.

nvh: And the wrapper class is not needed indeed, but I just wanted to point out that we had a similar problem in in the useAppearEffect hook

src/motion/features/viewport/observers.ts
nvh: 💯

src/motion/features/viewport/observers.ts
mattgperry: Oh I see it's more about the observation. I think just renaming to `startObservingIntersection`, `observeIntersection` or similar is fine then, creating the intersection observer itself is in some ways an implementation detail.

We can delete the observer too although my suspicion was it's unneeded because by disconnecting the observer we remove it from `entries` and, if the element itself has been removed, so too will the reference in the WeakMap. But there's no problem being explicit here just in case.

nvh: Ah, yeah, that rename sounds great!

src/motion/features/viewport/use-viewport.ts
```diff
@@ -0,0 +1,135 @@
...
+                ? props.onViewportEnter
+                : props.onViewportLeave
+            callback?.()
+        }
+
+        return observeIntersection(
+            visualElement.getInstance(),
+            options,
+            intersectionCallback
+        )
```
nvh, line 96: Much better! 👏🏻

src/motion/features/viewport/types.ts
```diff
@@ -0,0 +1,24 @@
...
+    root?: RefObject<Element>
+    once?: boolean
+    margin?: string
+    amount?: "some" | "all" | number
+}
+
+export interface ViewportProps {
+    whileInView?: VariantLabels | TargetAndTransition
+    onViewportEnter?: ViewportEventHandler
+    onViewportLeave?: ViewportEventHandler
```
nvh, line 17: While looking at our implementation, I noticed we provide a single callback that receives a boolean if the element is in viewport. I was wondering if that actually is a bit more flexible/smaller API than having two separate callbacks?

nvh (approved):

src/motion/features/viewport/types.ts
mattgperry: `onViewportChange`?

mattgperry: `onViewChange`

mattgperry: I do like the idea of a single listener but I'm leaning towards keeping them enter/leave because it mirrors everything else, `onHoverStart`/`End`, `onPressStart`/`End` etc

PR #1351, opened 2021-11-12
Add layoutDependency to valid props

rbourdon:
Resolves unrecognized prop warning from react.


mattgperry:
Great contribution, thanks!

Issue #1346, opened 2021-11-09
Keyframes don't replay when variant label changes


ShahriarKh:
Using the same keyframes sequence for opacity in different variants only animates on the first render. Also, the opacity stays at the last value in the next renders.

https://codesandbox.io/s/angry-raman-bc9sf?file=/src/App.js

**Steps to reproduce the behavior:**
1. Go to codesandbox link above.
2. Click on "X" Button to toggle state (& fire animation)
3. See problem

**Expected behavior**
The box should go from `opacity: 0` to `opacity: 1` on every click.

Changing the other keyframe to another value with a little difference works correctly:
`[0, 1]` for both variants => animaties on initial render and stays on 1
`[0, 1]` for one and `[0, 1.01`] for another => works correctly

Question on Stackoverflow:
https://stackoverflow.com/questions/69887735/framer-motion-opacity-keyframe-only-animates-on-initial-render

PR #1355, opened 2021-11-17
Keyframe fixes

mattgperry:
This PR makes a few keyframe fixes:

1. If keyframes in two variants are the same, but the variant has changed, re-run the keyframes animation.

```jsx
<motion.div
  animate={is ? "a" : "b"}
  variants={{ a: { x: [0,100] }, b: {x: [0,100] }  }}
/>
```

Fixes https://github.com/framer/motion/issues/1346

2. If `animate` contains keyframes, set the initial value/style to the initial keyframe

```jsx
<motion.div animate={{ opacity: [0, 1] }} />
// <div style="opacity: 0"></div>
```

3. If `initial={false}` and `animate` contains keyframes, set the initial value/style to the final keyframe

```jsx
<motion.div initial={false} animate={{ opacity: [0, 1] }} />
// <div style="opacity: 1"></div>
```

### DIFF
--- a/src/motion/utils/use-visual-state.ts
+++ b/src/motion/utils/use-visual-state.ts
@@ -63,19 +63,18 @@ function makeState<I, RS>(
     return state
 }
 
-export const makeUseVisualState = <I, RS>(
-    config: UseVisualStateConfig<I, RS>
-): UseVisualState<I, RS> => (
-    props: MotionProps,
-    isStatic: boolean
-): VisualState<I, RS> => {
-    const context = useContext(MotionContext)
-    const presenceContext = useContext(PresenceContext)
+export const makeUseVisualState =
+    <I, RS>(config: UseVisualStateConfig<I, RS>): UseVisualState<I, RS> =>
+    (props: MotionProps, isStatic: boolean): VisualState<I, RS> => {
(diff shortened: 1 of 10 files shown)


src/render/utils/animation-state.ts
```diff
@@ -201,8 +201,12 @@ export function createAnimationState(
              * a changed value or a value that was removed in a higher priority, we set
              * this to true and add this prop to the animation list.
              */
+            const isVariantChanged = variantsHaveChanged(
```
nvh, line 204: Maybe `variantDidChange`?

src/motion/utils/use-visual-state.ts
```diff
@@ -120,7 +119,20 @@ function makeLatestValues(
...
             const { transitionEnd, transition, ...target } = resolved
 
-            for (const key in target) values[key] = target[key]
+            for (const key in target) {
+                let valueTarget = target[key]
+
+                if (Array.isArray(valueTarget)) {
+                    const index = initialAnimationIsBlocked
+                        ? valueTarget.length - 1
+                        : 0
```
nvh, line 128: Can you explain why we take the last element here?

nvh (approved):

Issue #1358, opened 2021-11-18
[BUG] A motion element with the layout flag with an AnimatePresence sibling does not transition smoothly


rickyhopkins:
**1. Read the [FAQs](#faqs) 👇**
👍 

**2. Describe the bug**

A motion element with the layout flag with an `AnimatePresence` sibling does not transition smoothly when the AnimatePresence element is removed.
It will animate nicely on enter but not exit.

I may be doing something wrong here, but I would expect this to work.

If I move the layout div inside the AnimatePresence then it does work as expected, but this is not always possible.

**3. IMPORTANT: Provide a CodeSandbox reproduction of the bug**

https://codesandbox.io/s/framer-motion-simple-animation-forked-6m9ej?file=/src/index.tsx

**4. Steps to reproduce**

```tsx
    <div className="container">
      <Refresh onClick={toggle} />
      <AnimatePresence>
        {show && (
          <motion.div
            key="1"
            className="example-container"
            animate={{ opacity: 1 }}
            exit={{ opacity: 0 }}
            onClick={toggle}
          ></motion.div>
        )}
      </AnimatePresence>
      <motion.div
        onClick={toggle}
        key="2"
        className="example-container"
        layout
      ></motion.div>
    </div>
```

**5. Expected behavior**

The bottom motion div should animate smoothly when the top one is removed from the dom.

**6. Video or screenshots**

I think the codesandbox is easy enough to understand

rbourdon:
You can fix this by wrapping everything in a `<LayoutGroup>` 

example](https://codesandbox.io/s/framer-motion-simple-animation-forked-1q14u?file=/src/index.tsx)

rickyhopkins:
Thanks for the tip @rbourdon I had no idea that was even a thing, I still cannot find any reference to it in the docs.

I have tested this out in my app and I am now running into an issue where the item is not being removed from the dom by `AnimatePresence`.

I have a helper component which manages a fadeout for its children:

```tsx
interface Props {
	condition: boolean;
}

export const IfFade: FC<Props> = ({ children, condition }) => {
	return (
		<AnimatePresence>
			{condition && (
				<motion.div layout animate={{ opacity: 1 }} exit={{ opacity: 0 }} initial={{ opacity: 0 }}>
					{children}
				</motion.div>
			)}
		</AnimatePresence>
	);
};
```

[...]

PR #1361, opened 2021-11-22
Ensure sibling AnimatePresence components remove exiting children

mattgperry:
Fixes https://github.com/framer/motion/issues/1358

### DIFF
--- a/src/utils/use-force-update.ts
+++ b/src/utils/use-force-update.ts
@@ -3,16 +3,16 @@ import { useState, useCallback, useRef } from "react"
 import { useUnmountEffect } from "./use-unmount-effect"
 
 export function useForceUpdate(): [VoidFunction, number] {
-    const unloadingRef = useRef(false)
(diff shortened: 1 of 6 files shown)


src/utils/use-force-update.ts
```diff
@@ -6,10 +7,18 @@ export function useForceUpdate(): [VoidFunction, number] {
...
+    /**
+     * Defer this to  the end of the next animation frame in case there are multiple
+     * synchronous calls.
+     */
+    const deferredForceRender = useCallback(
+        () => sync.postRender(forceRender),
+        [forceRender]
+    )
+
+    return [deferredForceRender, forcedRenderCount]
```
nvh, line 23: What is the forcedRenderCount used for?

mattgperry: It could, for instance, be used in cache breaking

nvh (approved):

rickyhopkins:
Love your work!

Issue #1334, opened 2021-11-04
SVG properties do not animate when target value is a CSS Custom Property


tinleym:
**1. Describe the bug**
SVG properties do not animate when target is a CSS Custom Property; animation seems to work fine when the value is hardcoded.

**2. IMPORTANT: Provide a CodeSandbox reproduction of the bug**
In the example below, hardcoded values work on SVGs where variables do not...
https://codesandbox.io/s/keen-resonance-5rovv?file=/src/App.js

**3. Steps to reproduce**
1. Hover the text -- the animation sets color to a CSS Custom Property: ok
2. Hover the smile -- the animation sets the wrapper color to a CSS Custom Property (the target's stroke is currentColor): ok
3. Hover the paperclip -- the animation sets stroke to a CSS Custom Property: _does not work_
4. Hover the shield -- the animation sets stroke to a hardcoded value: ok
5. Hover the eye -- the animation sets stroke-width the a CSS Custom Property: _does not work_
6. Hover the wrench -- the animation sets stroke-width to a hardcoded value: ok

**4. Expected behavior**
The SVG properties should animate to CSS Custom Properties like they do for a hardcoded values.

**5. Environment details**
framer-motion 5.2.0 / Chrome 95.0.4638.69 / MacOs 11.4

tinleym:
OK, so this turned out to be me misunderstanding some things... here's an example of everything working:
https://codesandbox.io/s/keen-resonance-5rovv?file=/src/App.js

PR #1366, opened 2021-11-23
Allow SVG attributes to animate to CSS variables

mattgperry:
Fixes https://github.com/framer/motion/issues/1334

### DIFF
--- a/src/render/dom/utils/css-variables-conversion.ts
+++ b/src/render/dom/utils/css-variables-conversion.ts
@@ -15,7 +15,8 @@ function isCSSVariable(value: any): value is string {
  *
  * @param current
  */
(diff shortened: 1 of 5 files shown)


src/render/dom/utils/css-variables-conversion.ts
```diff
@@ -15,7 +15,8 @@ function isCSSVariable(value: any): value is string {
  *
  * @param current
  */
-export const cssVariableRegex = /var\((--[a-zA-Z0-9-_]+),? ?([a-zA-Z0-9 ()%#.,-]+)?\)/
+export const cssVariableRegex =
```
shuangq, line 18: Should we also cover some unit tests for the regex?

shuangq (approved):

src/render/dom/utils/css-variables-conversion.ts
mattgperry: The function that uses it, `parseCSSVariable`, is already covered by unit tests.

Issue #1313, opened 2021-10-28
Reorder creates a stacking context which breaks children with a z-index


PhilGarb:
# Description

Hi everyone! 
Huge fan of the library and the changes in v5. I stumbled upon an issue when trying out the new Reorder Components. If an element that is part of the Reorder.Item needs to have a z-index to overlap other items (like a dropdown or combobox) it is not working with the current approach. 

![image](https://user-images.githubusercontent.com/38015558/139300645-2262b6d2-6b46-46e1-8817-bfa838514a8c.png)
> **Please see the codesandbox for a reproduction. It is the Reorder example codesanbox forked.**
> [**Reproduction**](https://codesandbox.io/s/framer-motion-5-drag-to-reorder-lists-forked-7zv5x?file=/src/Item.tsx)

## Debug
I think the Reorder.Item has a stacking context, because it has `transform: none` applied. This might be coming from here, but I am absolutely not sure about that. https://github.com/framer/motion/blob/4f9e23b5e2bcbaf5aa8dd826191b60bbcab9f7f4/src/gestures/drag/VisualElementDragControls.ts#L472-L481

### Result
This results in children of Reorder.Item not being able to overlap other elements. 

# Expected Behavior

A child with a z-index should correctly overlap other items.

I see two possible ways to achieve this. 
1. Change the applied css so no stacking context is created. This is probably quite difficult if this is not just how the Reorder components work.
2. Allow the consumer to overwrite the zIndex of an item. It seems like here a z-index is created: https://github.com/framer/motion/blob/4f9e23b5e2bcbaf5aa8dd826191b60bbcab9f7f4/src/components/Reorder/Item.tsx#L60-L61 It is applied here and overwrites user styles making it impossible to pass a z-index to the element: https://github.com/framer/motion/blob/4f9e23b5e2bcbaf5aa8dd826191b60bbcab9f7f4/src/components/Reorder/Item.tsx#L79 is If instead of always using 0 here the user could provide a z-index it could be up to the consumer when to "raise" an item above the others.

Let me know if you need additional details or if I am missing a reason why it is working like this.

PR #1367, opened 2021-11-23
Changing at-rest z-index for Reorder.Item to "unset"

mattgperry:
Fixes https://github.com/framer/motion/issues/1313


dfelber:
FYI: This introduced a rendering bug (Dragged item not always on top) when reordering items.
You can see it even on the official page: https://www.framer.com/motion/reorder/ in the first example. (uses 6.2.8)
https://codesandbox.io/p/sandbox/framer-motion-5-drag-to-reorder-lists-uonye

In the example using drag triggers it behaves correctly (uses 5.0.0-rc.4)
https://codesandbox.io/p/sandbox/framer-motion-5-drag-to-reorder-lists-with-drag-handle-j9enw

PR #1385, opened 2021-12-07
fix: package.json exports paths

edoardolincetto:
After the release of the v5.4.2, importing Framer Motion will cause an error: `Invalid "exports" main target "dist/cjs/index.js" defined in the package config`. 
That's because all the paths defined in "exports" must be relative file URLs starting with `./`.


mattgperry:
Thank you!

Issue #1398, opened 2021-12-22
[BUG] MotionConfig does not propagate changes to transition


wurstbonbon:
**1. Describe the bug**

`MotionConfig` uses `toString` to memoize the transition that is passed down via context.  
`toString` returns `[object Object]` for regular objects, thus the context isn't updated when the transition is changed.

**3. IMPORTANT: Provide a CodeSandbox reproduction of the bug**
https://codesandbox.io/s/framer-motion-motion-config-memo-rohc3

**4. Steps to reproduce**

1. Click "Disabled Animation"
2. Click the button below.
-> Bug: The animation is still happing with a non-zero duration.

**5. Expected behavior**

Changes to the transition should be propagated. 

**6. Environment details**

Should be broken in all browser.

power-f-GOD:
https://github.com/framer/motion/pull/1399#issuecomment-1005276236

PR #1399, opened 2021-12-22
Fix MotionConfig does not propagate changes to transition

wurstbonbon:
Fixes https://github.com/framer/motion/issues/1398

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 Framer Motion adheres to [Semantic Versioning](http://semver.org/).
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


power-f-GOD:
Could this be a similar reason to why `motion.div` does not propagate changes to `variants` as well?
Whenever I change `variants` or (set|change)State of `variants`, changes don't reflect and the `motion` component still uses the previous value of `variants`. This is a bug. Will it be fixed? 

@wurstbonbon  Can you fix it too?

mattgperry (approved):
Excellent contribution, thanks

Issue #1406, opened 2022-01-04
[BUG] Prevent props from passing through to DOM elements


Simba14:
It seems the issue detailed in [Issue 272](https://github.com/framer/motion/issues/272) has re-occurred when using the latest version `5.5.5` in combination with styled components.  I have tried earlier versions but same error is thrown: 
`React does not recognize the `customProp` prop on a DOM element.`

It seems all props are passed down by motion to the DOM element. They should be stripped out to prevent this error.
I have seen that an earlier fix was to utilise ` @emotion/is-prop-valid` as an optional dependency.
Perhaps this dependency is not correctly being used in cases when it should?

**3. IMPORTANT: Provide a CodeSandbox reproduction of the bug**
https://github.com/mattbrannon/motion-console-warning
Unfortunately unable to reproduce via sandbox, but @mattbrannon made this public repo specifically to demonstrate the issue. 

**4. Steps to reproduce**
`customProp` is passed on to the DOM element when using Styled Components
```
const Custom = styled(motion.div)`
  opacity: ${({ customProp }) => customProp};
`

...
return (
  <Custom customProp={1}></CustomProp>
)
```
**5. Expected behavior**

For custom props to filtered out and not passed on to the DOM element created by motion.

GR34SE:
I have encountered the same issue right after upgrading from v3.10.6 to the latest v5.5.5. It seems like it appeared since v5.0.0 as it works well in 4.1.17

Simba14:
A workaround is using Styled Components [transient props](https://styled-components.com/docs/api#transient-props) for anyone blocked by this

mattgperry:
I wonder if the move to `.mjs` es files somehow breaks the ability to use optional dependencies. I'll take a look

mattgperry:
Yeah, in that sandbox by moving the optional import outside the `try` I get `require is not defined`

nataliarossini:
I had the same issue and fixed by only adding a $ to my props like so `$customProp`
Apparently this release solves the issue with props filtering
https://github.com/styled-components/styled-components/releases/tag/v5.1.0

mattgperry:
I've just published `5.6.0` which will let you pass a custom function:

```jsx
import { MotionConfig } from "framer-motion"
import isValidProp from "@emotion/is-prop-valid"

function App() {
  return (
    <MotionConfig isValidProp={isValidProp}>

   </MotionConfig>
  )
}
```

PR #1410, opened 2022-01-07
Adding isValidProp to MotionConfig

mattgperry:
This PR adds an `isValidProp` prop to `MotionConfig` that allows the external injection of an `isValidProp` function.

Fixes https://github.com/framer/motion/issues/1406

### DIFF
--- a/src/components/MotionConfig/__tests__/index.test.tsx
+++ b/src/components/MotionConfig/__tests__/index.test.tsx
@@ -8,7 +8,7 @@ describe("custom properties", () => {
         const Component = () => {
             return (
                 <MotionConfig isValidProp={(key) => key !== "data-foo"}>
-                    <motion.div data-foo="bar" />
+                    <motion.div data-foo="bar" data-bar="foo" />
                 </MotionConfig>
             )
         }
@@ -16,5 +16,6 @@ describe("custom properties", () => {
         const { container } = render(<Component />)
 
         expect(container.firstChild).not.toHaveAttribute("data-foo")
+        expect(container.firstChild).toHaveAttribute("data-bar")
     })
 })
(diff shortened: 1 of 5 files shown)


src/components/MotionConfig/__tests__/index.test.tsx
```diff
@@ -0,0 +1,20 @@
...
+            return (
+                <MotionConfig isValidProp={(key) => key !== "data-foo"}>
+                    <motion.div data-foo="bar" />
+                </MotionConfig>
+            )
+        }
+
+        const { container } = render(<Component />)
+
+        expect(container.firstChild).not.toHaveAttribute("data-foo")
```
shuangq, line 18: Should we also test that a valid attribute is forwarded to the DOM?

src/render/dom/utils/filter-props.ts
```diff
@@ -3,6 +3,16 @@ import { isValidMotionProp } from "../../../motion/utils/valid-prop"
 
 let shouldForward = (key: string) => !isValidMotionProp(key)
 
+export type IsValidProp = (key: string) => boolean
+
+export function loadExternalIsValidProp(emotionIsPropValid?: IsValidProp) {
```
shuangq, line 8: Why do we still call it `emotionIsPropValid` if it could be a user-provided function?

src/render/dom/utils/filter-props.ts
```diff
@@ -3,6 +3,16 @@ import { isValidMotionProp } from "../../../motion/utils/valid-prop"
...
+    if (!emotionIsPropValid) return
+
+    // Handle events explicitly as Emotion validates them all as true
+    shouldForward = (key: string) =>
+        key.startsWith("on") ? !isValidMotionProp(key) : emotionIsPropValid(key)
```
shuangq, line 13: If there's an injected `isValidProp` provided by `MotionConfig`, wouldn't it be hijacked by `isValidMotionProp`? For example, `isValidProp={key => key !== "onStart"}`

mattgperry: I think that might have to be a limitation of the API that our events are always consumed

mattgperry:
@shuangq Great comments, I've just pushed some changes

shuangq (approved):
LGTM 👏

PR #1415, opened 2022-01-11
Handle removal of transformTemplate

mattgperry:


src/motion/__tests__/transformTemplate.test.tsx
```diff
@@ -0,0 +1,114 @@
...
+    })
+
+    it("removes transformTemplate if prop is removed and transform is not changed", async () => {
+        const { container, rerender } = render(
+            <motion.div
+                transformTemplate={() => `translateY(20px)`}
+                style={{ x: 10 }}
+            />
+        )
+        expect(container.firstChild).toHaveStyle("transform: translateY(20px)")
```
hemlok, line 93: I'm probably being thick, but why doesn't this have `transform: translateY(20px) translateX(10px)`?

hemlok: Oh I am, because the template doesn't pass through the transforms 🤦

hemlok (approved):

Issue #1413, opened 2022-01-09
must borderRadius be unitless for crossfade effect ?


ArianHamdi:
**1. Read the [FAQs](#faqs) 👇**

**2. Describe the bug**

when I want to change `borderRadius` during the crossfade effect if I add a unit e.g. `borderRadius : "20px"` or `borderRadius : "20%"` , it doesn't work properly as I expected. Thus I was forced to omit the unit e.g. `borderRadius : 20`.

Is it a bug or it's normal for you?

**3. IMPORTANT: Provide a CodeSandbox reproduction of the bug**

https://codesandbox.io/s/objective-fermat-t4hq0?file=/src/App.js

**4. Steps to reproduce**

1. Open codesandbox link.
2. Click on an element to see the crossfade effect and smooth changes in borderRadius.
3. Add unit to borderRadius e.g. "20px".
4. Now click on an element and see borderRadius changes. you can see that the changes are not smooth.

ShahriarKh:
🤔 Seems to be a bug with pixel units.
I checked the browser console. Unitless `border-radius` is converted to percentages.
![image](https://user-images.githubusercontent.com/31452340/148684198-11b58571-9181-49fe-a591-7e07ddfefc00.png)

mattgperry:
Yeah I think if it’s a string it’s handled as a percentage, number for pixels. We should probably check for the px

mattgperry:
My mistake its that strings aren't supported at all for crossfading

PR #1416, opened 2022-01-11
Crossfade border-radius of "%" and "px" units

mattgperry:
Fixes https://github.com/framer/motion/issues/1413

### DIFF
--- a/src/projection/animation/mix-values.ts
+++ b/src/projection/animation/mix-values.ts
@@ -1,9 +1,13 @@
 import { circOut, linear, mix, progress as calcProgress } from "popmotion"
+import { percent } from "style-value-types"
 import { ResolvedValues } from "../../render/types"
 import { EasingFunction } from "../../types"
 
 const borders = ["TopLeft", "TopRight", "BottomLeft", "BottomRight"]
 const numBorders = borders.length
+
+const asNumber = (value: string | number) =>
+    typeof value === "string" ? parseFloat(value) : value
 
 export function mixValues(
     target: ResolvedValues,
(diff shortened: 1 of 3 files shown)


src/projection/styles/scale-border-radius.ts
```diff
@@ -17,6 +17,7 @@ export function pixelsToPercent(pixels: number, axis: Axis): number {
 export const correctBorderRadius: ScaleCorrectorDefinition = {
     correct: (latest, node) => {
         if (!node.target) return latest
+
```
nvh, line 20: This seems unnecessary

src/projection/animation/mix-values.ts
```diff
@@ -47,17 +51,13 @@ export function mixValues(
...
-        ) {
-            const radius = Math.max(mix(followRadius, leadRadius, progress), 0)
-            target[borderLabel] = radius
+        target[borderLabel] = Math.max(
+            mix(asNumber(followRadius), asNumber(leadRadius), progress),
+            0
+        )
+
+        if (percent.test(leadRadius) || percent.test(followRadius)) {
+            target[borderLabel] += "%"
```
nvh, line 60: What happens if one is pixels and the other is percent? Then we would treat them both as percent, right?

nvh: (Also don't see that in a test case)

mattgperry: Yeah currently the crossfade breaks completely, in this case they'd be treated as a percentage. I'm not sure either is ideal though we don't generally support animating between value types anywhere else either.

nvh: Wouldn't it be better to not animate then do something weird?

mattgperry: Ah yeah I've just pushed this fix

src/projection/animation/mix-values.ts
```diff
@@ -51,13 +54,22 @@ export function mixValues(
...
         leadRadius ||= 0
 
-        target[borderLabel] = Math.max(
-            mix(asNumber(followRadius), asNumber(leadRadius), progress),
-            0
-        )
+        const canMix =
+            followRadius === 0 ||
+            leadRadius === 0 ||
+            isPx(followRadius) === isPx(leadRadius)
```
nvh, line 60: I'm confused why we're not adding `percent.test(leadRadius) === percent.test(followRadius)` to this?

nvh: But the test case seems to be still passing...

nvh (approved):

src/projection/animation/mix-values.ts
mattgperry: This is essentially testing if they're different

nvh: Ah right! So in the purely hypothetical case that there would be a third format, we would be fucked.
Let's add a comment about that then, because I think it's pretty confusing.

PR #1426, opened 2022-01-20
Add `reducedMotion` prop to `MotionConfig`

mattgperry:
This PR adds a new option to `MotionConfig` that lets people automatically `respectReducedMotion`

```jsx
<MotionConfig respectReducedMotion>
```

Setting this option to `true` will disable all movement animations, for now just transforms and layout animations. While leaving things like opacity and backgroundColor in-tact.


packages/framer-motion/src/context/MotionConfigContext.tsx
```diff
@@ -25,6 +25,14 @@ export interface MotionConfigContext {
...
      */
     transition?: Transition
+
+    /**
+     * If true, will respect the device prefersReducedMotion setting by switching
+     * transform animations off.
+     *
+     * @public
+     */
+    respectReducedMotion?: boolean
```
nvh, line 35: It seems to me that it would be better to have three options here: 
* always reduce motion
* never reduce motion
* listen to prefersReduceMotion
Maybe we can make the API reflect those?


Something like:

```suggestion
    reducedMotion?: "always" | "never" | "userPreference"`
```

packages/framer-motion/src/context/MotionConfigContext.tsx
```diff
@@ -33,4 +41,5 @@ export interface MotionConfigContext {
 export const MotionConfigContext = createContext<MotionConfigContext>({
     transformPagePoint: (p) => p,
     isStatic: false,
+    respectReducedMotion: false,
```
nvh, line 44: Why not enable this by default? Seems like a nice thing to have, right?

packages/framer-motion/src/projection/node/create-projection-node.ts
```diff
@@ -451,13 +451,21 @@ export function createProjectionNode<I>({
...
                                 ...getValueTransition(
                                     layoutTransition,
                                     "layout"
                                 ),
                                 onComplete: onLayoutAnimationComplete,
-                            })
+                            }
+
+                            if (visualElement.shouldReduceMotion) {
+                                animationOptions.delay = 0
```
nvh, line 464: I don't understand why we set the delay to 0

packages/framer-motion/src/render/types.ts
```diff
@@ -35,6 +35,7 @@ export interface VisualElement<Instance = any, RenderState = any>
     measureViewportBox(withTransform?: boolean): Box
     addVariantChild(child: VisualElement): undefined | (() => void)
     getClosestVariantNode(): VisualElement | undefined
+    shouldReduceMotion?: boolean | null
```
nvh, line 38: What is the difference between `undefined` and `null` here?

nvh (commented):
I'm not sure about the naming of `respectReducedMotion`, and thinking we can come up with a more powerful API

packages/framer-motion/src/context/MotionConfigContext.tsx
mattgperry: What's the use-case for `"always"`?

packages/framer-motion/src/projection/node/create-projection-node.ts
mattgperry: Ah because usually `delay` is set (or derived from `staggerChildren`/`delayChildren`) to work in tandem with `duration`, which is now essentially `0`

packages/framer-motion/src/render/types.ts
mattgperry: `null` is provided server-side, basically in situations where `window` doesn't exist. The API here is pre-existing in `useReducedMotion`

mattgperry:
@nvh I'm not sure what more we'd want to do with this other than hone the types of motion we keep enabled or disable. Ideally this is quite simple to implement - I'm thinking a checkbox in the site settings page.

packages/framer-motion/src/context/MotionConfigContext.tsx
nvh: Maybe it doesn't make a lot of sense to use an animation framework and disable the animations 😆, but I was thinking about giving the developer control over this feature, like being able to offer a toggle to the user to enable/disable reduced motion.

nvh: I was mostly trying to future-proof this API, because there is a third option that might be usefull at some point.

mattgperry: Ah I see, yeah that's a good idea. How about to keep the cleanliness of 

```jsx
<MotionConfig respectReducedMotion>
```

And to keep the potential Framer settings page toggle in mind, we make this a boolean and then in a later run extend `true` and `false` with

```jsx
<MotionConfig respectReducedMotion="enforce">
```

mattgperry: Actually on further thought if we made it `"user"` I think this is a good change, we can still surface it as a boolean in Framer

nvh: I lost track, can you write out the API you have in mind now?

nvh: Ah, I see you already implemented it, yes, this looks good!

nvh (approved):

PR #1438, opened 2022-02-01
Upgrade to React 18

mattgperry:
This PR makes Framer Motion compatible with React 18.

I've downgraded it back to 17 in resolutions as it wasn't compatible with all test suites.

The biggest change is that components can be re-mounted, so they can't consider an unmount the same as the end of their lifecycle.

Closes https://github.com/framer/motion/pull/826


packages/framer-motion/src/animation/use-animated-state.ts
```diff
@@ -59,14 +58,16 @@ export function useAnimatedState(initialState: any) {
 
     useEffect(() => {
         element.mount({})
-        return element.unmount()
-    }, [])
+        return element.unmount
```
mattgperry, line 61: This isn't React 18 specific, but something that must have been a quiet bug before

vpontis:
Awesome thank you @mattgperry 👍

PR #1443, opened 2022-02-03
Fixing layout animations with gap Safari fix

mattgperry:


packages/framer-motion/src/projection/node/create-projection-node.ts
```diff
@@ -1081,11 +1090,11 @@ export function createProjectionNode<I>({
         /**
          * Animation
          */
-        animationProgress = 0
```
nvh, line 1084: What was this used for before?

mattgperry: Nothing, must have been left around from long ago

packages/framer-motion/src/projection/node/create-projection-node.ts
```diff
@@ -467,6 +467,15 @@ export function createProjectionNode<I>({
 
                             this.startAnimation(animationOptions)
                         } else {
+                            /**
+                             * If the layout hasn't changed and we have an animation that hasn't started yet,
+                             * finish it immediately. Otherwise it will be animating from a location
+                             * that was probably never commited to screen and look like a jumpy box.
+                             */
+                            if (!hasLayoutChanged && this.progress === 0) {
```
shuangq, line 475: If the layout hasn't changed, why did it start the animation? 🤔

mattgperry: In a previous render where the layout did change. This is a new render that may have happened so soon that the old animation didn't even start to play.

shuangq (approved):
Code looks good, any chance we can cover a test? 😬

packages/framer-motion/src/projection/node/create-projection-node.ts
```diff
@@ -1081,11 +1090,11 @@ export function createProjectionNode<I>({
         /**
          * Animation
          */
-        animationProgress = 0
         animationValues?: ResolvedValues
         pendingAnimation?: Process
         currentAnimation?: AnimationPlaybackControls
         mixTargetDelta: (progress: number) => void
+        progress = 0
```
nvh, line 1097: I do feel that `animationProgress` is more descriptive, why did you go for `progress` here?

nvh (approved):

packages/framer-motion/src/projection/node/create-projection-node.ts
mattgperry: Yeah animationProgress makes sense at a class level, I'll change it

PR #1452, opened 2022-02-13
Use isomorphic layout effect in useIsMounted

thebuilder:
The React 18 upgrade in #1438 introduced a new `useIsMounted` hook, that that didn't use the `useIsomorphicLayoutEffect`. This resulted in a warning rendered serverside. Fix by replacing `useLayoutEffect` with `useIsomprohicLayoutEffect`.

Fix: #578


mattgperry:
Brilliant thanks for the contribution! I’m OOO until a week on Monday but will publish then

PR #1466, opened 2022-02-23
Fix blocked animations

mattgperry:
Closes https://github.com/framer/motion/pull/1451

This clears protectedKeys after animations triggered by `setActive`, ie gestures like `whileHover` etc. This removes the race condition of trying to animate a tree via state immediately after doing so with a gesture.

### DIFF
--- a/packages/framer-motion/src/render/utils/animation-state.ts
+++ b/packages/framer-motion/src/render/utils/animation-state.ts
@@ -384,7 +384,14 @@ export function createAnimationState(
 
         state[type].isActive = isActive
 
-        return animateChanges(options, type)
+        const animations = animateChanges(options, type)
+
+        // Remove this and create failing test
+        for (const key in state) {
+            state[key].protectedKeys = {}
+        }
+
+        return animations
     }
 
     return {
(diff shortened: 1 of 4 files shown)


packages/framer-motion/src/render/utils/animation-state.ts
```diff
@@ -384,7 +384,13 @@ export function createAnimationState(
 
         state[type].isActive = isActive
 
-        return animateChanges(options, type)
+        const animations = animateChanges(options, type)
+
+        for (const key in state) {
+            state[key].protectedKeys = {}
```
onnlucky, line 390: 😄

onnlucky (approved):
LGTM, good to see tests :)

packages/framer-motion/src/render/utils/animation-state.ts
shuangq: Does this mean animations triggered by setActive will never be protected if the value is unchanged? Would this cause other issues?

mattgperry: None that have been caught by tests 😅

This shouldnt be a problem because setActive will start the animation from the instigating component down through the tree whereas animations fired from useEffect will go from components bottom-up, and this is when we need to protect animations from being overridden by parents

shuangq: ah nice context, thanks for confirming!

Issue #1048, opened 2021-03-20
[BUG] Lingering `requestAnimationFrame` when removing animated component


DerNamenlose:
When removing an animated component with an infinite loop, there is a lingering `requestAnimationFrame` in the background, that keeps on using the CPU. 

**CodeSandbox**
https://mo41z.csb.app/

**4. Steps to reproduce**

1. Conditionally render a component with an animation with `loop: Infinity`.
2. Remove that component from the DOM by setting the condition to false.
3. Create a Performance Profile in DevTools -> there is a `requestAnimationFrame` running every 16 ms (for my 60 Hz monitor anyway), even though there is no longer any animated component in the DOM.
4. (Unsure) The `requestAnimationFrame`-calls seem to add up, causing 100% CPU in a long running process (Electron App in my case). However, I'm not 100% positive on how to actually reproduce this reliably.

**5. Expected behavior**

When the animated component is removed, the calls to `requestAnimationFrame` should stop, relinquishing any control of the CPU.

**7. Environment details**

Electron 12.0.1 (Chromium 89.0.4389.82) on Kubuntu 20.10.

fedekau:
@DerNamenlose did you find out something that helped you resolve the problem?

We have something very similar, just by doing this we are able to reproduce it 🤯 

```
<AnimatePresence>
   <div>Hi</div>
</AnimatePresence>
```

fedekau:
@mattgperry I am pinging you since I saw you active on a thread about `requestAnimationFrame` too. Does this have an explanation? Our Framer version is `4.1.17`. Let me know if you need more details

DerNamenlose:
@fedekau Sorry, I replaced the animation with a CSS animation due to a code overhaul anyway.

fedekau:
Thanks @DerNamenlose!

We needed an animation to run infinitely as long as the component was mounted so we ended up forcing the animation by changing the key `onAnimationCompleted`. Basically infinity loop, but manually. 

```jsx
const [key, setKey] = useState(0);

return <motion.div key={key} onAnimationCompleted={() => setKey(key+1)} ...>...</motion.div


```

fatso83:
Any idea on how to fix this in code? I made a full reproduction case after profiling this for a while.
https://github.com/fatso83/issue-reproductions/tree/master/framer-motion-cpu-issue

[...]

Issue #886, opened 2020-12-11
[BUG] AnimatePresence doesn't unmount when animation is finished when using a custom bounce transition value


linusklingzell:
**1. Read the [FAQs](#faqs) 👇**
yes

**2. Describe the bug**

AnimatePresence dosen't unmount its children after exit animation is finished when using a custom bounce value

**3. IMPORTANT: Provide a CodeSandbox reproduction of the bug**

https://codesandbox.io/s/sharp-water-8rh86?file=/src/App.js

**4. Steps to reproduce**

Add a custom bounce value in transition prop in a motion.div component when using AnimatePresence
Click "Toggle me" in sandbox twice to show and then hide the component

**5. Expected behavior**

The children in AnmiatePresence should be unmounted when animation is finished

Lukavyi:
I have a similar issue, had to disable animations because of it.

KirdesMF:
it works if you set the duration of the transition

williambout:
I can repro as well

DovahBrownies:
Can confirm that it will keep bouncing infinitely.
Adding a `duration` does unmount the animation.

fatso83:
Possibly related to #1048.

mattgperry:
Since fixed

PR #1479, opened 2022-03-04
Fix for  #1048: stop all animations on unmount

fatso83:
Issue #1048 (and possibly #886?) is a performance issue that basically comes down to Framer Motion not cleaning up the popmotion animations. This PR simply adds an event listener on the VisualElement that will cause it to call `stop()` on all `MotionValue`s. Test verification has been added.

There were a few hiccups along the way with regards to testing (some reported), which prevented me from being able to run the entire test suite, but I made sure to test the changes I did by only running those files:
```
 npx jest --testPathPattern src/render/__tests__/index.test.ts
```

P.S. I was not totally sure what to put in the changelog, seeing that it requires a date for the version and such, which seems like a job for some automation tool.


fatso83:
I am not quite able to decipher the TS issue:
```
TS2322: Type 'VisualElement<any, any>' is not assignable to type 'VisualElement<unknown, any>'.   The types of 'parent.sortNodePosition' are incompatible between these types.     Type '(element: import("/Users/carlerik/dev/motion/packages/framer-motion/types/render/types").VisualElement<any, any>) => number' is not assignable to type '(element: import("/Users/carlerik/dev/motion/packages/framer-motion/src/render/types").VisualElement<any, any>) => number'.       Types of parameters 'element' and 'element' are incompatible.         Type 'import("/Users/carlerik/dev/motion/packages/framer-motion/src/render/types").VisualElement<any, any>' is not assignable to type 'import("/Users/carlerik/dev/motion/packages/framer-motion/types/render/types").VisualElement<any, any>'.           Types of property 'animateMotionValue' are incompatible.             Type '(key: string, value: import("/Users/carlerik/dev/motion/packages/framer-motion/src/value/index").MotionValue<any>, target: import("/Users/carlerik/dev/motion/packages/framer-motion/src/types").ResolvedValueTarget, transition?: import("/Users/carlerik/dev/motion/packages/framer-motion/src/types").Transition) => Promi...' is not assignable to type '(key: string, value: import("/Users/carlerik/dev/motion/packages/framer-motion/types/value/index").MotionValue<any>, target: import("/Users/carlerik/dev/motion/packages/framer-motion/types/types").ResolvedValueTarget, transition?: import("/Users/carlerik/dev/motion/packages/framer-motion/types/types").Transition) =>...'.               Types of parameters 'value' and 'value' are incompatible.                 Type 'import("/Users/carlerik/dev/motion/packages/framer-motion/types/value/index").MotionValue<any>' is not assignable to type 'import("/Users/carlerik/dev/motion/packages/framer-motion/src/value/index").MotionValue<any>'.                   Types have separate declarations of a private property 'current'.  types.ts(153, 5): The expected type comes from property 'parent' which is declared here on type 'VisualElementOptions<ResolvedValues, any>'
```

I could throw a bunch of `any`s on the problem, but I would rather do it right.

packages/framer-motion/src/render/index.ts
```diff
@@ -515,6 +515,13 @@ export const visualElement =
             },
         }
 
+        element.onUnmount(() => {
+            element.forEachValue((value) => {
+                //console.log(value)
```
fedekau, line 520: This can be removed 😄

fatso83: Thanks! Any idea about the TypeScript issues? Not sure what that is about.

fatso83 (commented):
In hindsight, I could probably have stuffed the changed directly in [the `unmount` method](https://github.com/framer/motion/pull/1479/files#diff-0d13ecc0e463d5f0388f01d12b30363d33f01819480822de9c04604d39eda21aL250). I thought I had tested that already (unsuccessfully) instead of as a callback on the `unmount` event., but after [patching just that method in 2.9.5](https://github.com/framer/motion/issues/1048#issuecomment-1059137184) and seeing that work just fine I am unsure what I saw :shrug: 

Will test that as an alternative once TS tests are passing.

packages/framer-motion/src/render/__tests__/create-element-util.ts
```diff
@@ -0,0 +1,78 @@
...
+        const origin = getOrigin(target as any, transition || {}, element)
+        checkTargetForNewValues(element, target, origin as any)
+        return { transition, transitionEnd, ...target }
+    },
+})
+
+export function createVisualElementAndAnimationState(
+    props: MotionProps = {},
+    parent: VisualElement<any> | undefined = undefined
+): { element: VisualElement; state: any } {
```
fedekau, line 43: I think it expects `VisualElement | undefined` Right?

mattgperry:
The fundamental problem with this approach is a motion value can be shared across multiple VisualElements. Perhaps it could be that if all `onChange` listeners are removed, `stop` is called?

fatso83:
Ah, good catch. I'll update. My basic assumption is that unmount = destroy (release resources), seeing that the [`unmount()`](https://github.com/framer/motion/blob/7526357419fd75aed18d82252aa843e046f2ca21/packages/framer-motion/src/render/index.ts#L254) releases all its listeners on the MotionValue. With that in mind, I should be able to something like  this straight after the event listener release in Line 254 of the VisualElement `unmount()` method
```javascript    
            element.forEachValue((value) => {
                if(value.hasNoListeners()) { value.stop(); }
            });
```
followed by this addition to [MotionValue](https://github.com/framer/motion/blob/7526357419fd75aed18d82252aa843e046f2ca21/packages/framer-motion/src/value/index.ts#L175):

```javascript
hasNoListeners() {
  return this.updateSubscribers.getSize() === 0;
}
```

Right?

fatso83:
[A patch for 2.9.5 using your approach](https://github.com/framer/motion/issues/1048#issuecomment-1059137184) worked fine (i.e. CPU usage was down to 0 when idle), so will update the PR for the `main` branch with something similar.

fatso83:
While the approach of stopping MotionValue's when the number of listeners reached zero seemed fine at first, as it solved the issue of loaders with `repeat=Infinite`, I had to revert the patch when we found that the animated messages in a message list never transitioned from their `opacity=0` to `opacity=1` state, staying hidden. I have not had time to do a deep dive into what about the "fix" is causing this, but it means the approach needs some adjustments in some way or another.

The code that never transitioned (because `stop()` was being called on the backing MotionValues) looks quite inconspicuous to me:
```javascript
const StyledMessageItem = styled(motion.li)<StyledMessageItemProps>`
  display: flex;
  flex-direction: column;
  margin-bottom: 0.75rem;
`

export function MessageItem(props: MessageItemProps): JSX.Element {
  const { message } = props

  return (
    <StyledMessageItem
      initial={{ opacity: 0, scale: 0.8 }}
      animate={{ opacity: 1, scale: 1 }}
    >{message}
    </StyledMessageItem>
  }
}
```

this again is being used quite vanilla like this:
```javascript
            <StyledMessageList>
              {messages.map((message, index) => (
                <MessageItem key={index} message={message} />
              ))}
            </StyledMessageList>
```

Not sure what makes this so special, tbh, as we don't apply anything special with regards to MVs, but when I set some breakpoints, it seemed as the check for zero listeners on onChange happened very early.

So basically this PR is not ready and going back to DRAFT state.

mattgperry:
Closing this out for now, feel free to reopen if there's a solution found though

PR #1497, opened 2022-03-31
Make sure useCycle's `runCycle` doesn't change on every render

### LINKED ISSUE #1496: [BUG]

**2. Describe the bug**

You cannot use the results of useCycle inside a useEffect, because useCycle's 2nd return value changes on every render.

e.g.

```ts
const [v, cycleV] = useCycle(1,2,3);
useEffect(() => {
  setMyThing(myThing, v);
  cycleV();
}, [myThing, cycleV]);
```

The problem is that `cycleV` changes on every render, so this useEffect will just keep getting called infinitely. 

A CodeSandbox minimal reproduction will allow us to quickly follow the reproduction steps. **Without one, this bug report won't be accepted.**

I'm pushing a PR, I'm hoping with PR will make this self-explanatory

**4. Steps to reproduce**

1. see code example above

**5. Expected behavior**

`useEffect` should only fire once, when `myThing` changes, above

**6. Video or screenshots**

If applicable, add a video or screenshots to help explain the bug.

**7. Environment details**

If applicable, let us know which OS, browser, browser version etc you're using.

alecf:
This allows the runCycle method to be called from useEffect, without triggering useEffect itself.

Fixes #1496

### DIFF
--- a/packages/framer-motion/src/utils/use-cycle.ts
+++ b/packages/framer-motion/src/utils/use-cycle.ts
@@ -1,5 +1,5 @@
-import { useState, useRef } from "react"
 import { wrap } from "popmotion"
+import { useCallback, useRef, useState } from "react"
 
 type Cycle = (i?: number) => void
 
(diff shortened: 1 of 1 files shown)


mattgperry:
Thanks for the PR!

PR #1500, opened 2022-04-05
Use `process` library instead of global process

huntercaron:
- Adds `process` dependency 
- Imports and uses `process` in the 3 files that we currently expect `process` to be a global

This is a pattern we already use in `FramerStudio` and will fix usages of framer-motion in no-build browser environments.


vovacodes (approved):
😍

mattgperry:
If you run `yarn build & yarn measure` `process` is adding 0.5kb to the generated bundles but I don't think it should be included at all?

PR #1503, opened 2022-04-06
[FEATURE] ability to pass layout to Reorder.Item

gmathieu:
This is primarily used to disable `layout="size"` without breaking reorder functionality. The default is still `layout=true` for backwards compatibility.


mattgperry:
Thanks for the PR!

PR #1508, opened 2022-04-07
Change useSpring memo dependency to json string

kcirtaptrick:
useSpring will emit the error below if the number of config properties changes dynamically
```diff
- Warning: The final argument passed to useMemo changed size between renders. The order and size of this array must remain constant.
```
This also causes the config to not update properly

This PR fixes the issue


mattgperry:
Great catch, thanks!

PR #1522, opened 2022-04-21
Fix SSR detection in Deno

heypiotr:
Deno provides a `window` object, which makes Framer Motion believe it's running in a browser when using Deno for SSR. They suggest checking the `document` instead:

https://github.com/denoland/deno/issues/13367

There's a few more `typeof window` checks in `dist`, centered around `window.requestAnimationFrame`. I suspect these are coming from some dependency. In my testing (SSRing www.framer.com/sites in Deno), they haven't caused any trouble, despite `window.requestAnimationFrame` being `undefined` (which means if there is some code path which triggers them, it would crash).

image

### DIFF
--- a/packages/framer-motion/src/value/scroll/use-viewport-scroll.ts
+++ b/packages/framer-motion/src/value/scroll/use-viewport-scroll.ts
@@ -21,7 +21,6 @@ let hasListeners = false
 
 function addEventListeners() {
     hasListeners = true
-    if (typeof window === "undefined") return
 
     const updateScrollValues = createScrollUpdater(
         viewportScrollValues,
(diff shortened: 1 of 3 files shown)


packages/framer-motion/src/value/scroll/use-viewport-scroll.ts
```diff
@@ -21,7 +21,6 @@ let hasListeners = false
 
 function addEventListeners() {
     hasListeners = true
-    if (typeof window === "undefined") return
```
heypiotr, line 24: `addEventListeners` is called in an effect, which makes this check redudant. But let me know if you feel it's better to leave it in.

mattgperry (approved):

PR #1542, opened 2022-05-23
Update MotionConfig context value when reducedMotion is updated

### LINKED ISSUE #1541: [BUG] MotionConfig updating reducedMotion prop does not update context value
Updating `reducedMotion` prop to `MotionConfig` does not update context value, so inner `useReducedMotionConfig` cant being updated.

[codesandbox](https://codesandbox.io/s/inspiring-monad-18ipee?file=/src/App.js)

clicking on the button the phrase above should change from "should reduce motion" to "should reduce motion" but nothing occurs

kalimantos:
add reducedMotion to useMemo dependencies so that context value can update when reducedMotion is updated

fixes #1541

### DIFF
--- a/packages/framer-motion/src/components/MotionConfig/index.tsx
+++ b/packages/framer-motion/src/components/MotionConfig/index.tsx
@@ -53,7 +53,7 @@ export function MotionConfig({
      */
     const context = useMemo(
         () => config,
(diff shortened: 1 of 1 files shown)


mattgperry:
Thanks for the PR!

PR #1306, opened 2021-10-26
Improve performance of addEventListener usage

o-alexandrov:
Description copied from #1233

Closes #1143
Please scroll below or open [Files changed](https://github.com/framer/motion/pull/1306/files) to see the explanation and a comment to the actual change and what exact methods will be improved by that change.

In short, since:
- there's no usage of `event.preventDefault()`
  - so there is no possibility of affecting any code within the `motion` repo
- you bump the major version in #1226 
  - so users who use `useDomEvent` method in their projects and depend on `event.preventDefault()` won't pull this PR changes if they have `^` for the version of the library

PR #1547, opened 2022-05-25
Passive events by default

mattgperry:
This PR sets event listeners to passive by default, so things like `whileTap` will be non-blocking. We disable this if a user has add an explicit event listener, in case they use it to call `event.preventDefault()`, as a backwards-compatible way of introducing this. 

In the future it might be possible to add a prop that can enable/disable this behaviour so a user can say "this event doesnt block events" but might be worth checking into seeing how React is handling this or waiting for them to do so https://github.com/facebook/react/issues/6436.

Closes https://github.com/framer/motion/pull/1306
Fixes https://github.com/framer/motion/issues/1143

### DIFF
--- a/packages/framer-motion/src/events/use-dom-event.ts
+++ b/packages/framer-motion/src/events/use-dom-event.ts
@@ -4,7 +4,7 @@ export function addDomEvent(
     target: EventTarget,
     eventName: string,
     handler: EventListener,
-    options?: AddEventListenerOptions
+    options: AddEventListenerOptions = { passive: true }
 ) {
     target.addEventListener(eventName, handler, options)
 
--- a/packages/framer-motion/src/events/use-dom-event.ts
+++ b/packages/framer-motion/src/events/use-dom-event.ts
@@ -8,7 +8,7 @@ export function addDomEvent(
 ) {
     target.addEventListener(eventName, handler, options)
 
-    return () => target.removeEventListener(eventName, handler, options)
+    return () => target.removeEventListener(eventName, handler)
 }
 
 /**
--- a/packages/framer-motion/src/gestures/use-hover-gesture.ts
+++ b/packages/framer-motion/src/gestures/use-hover-gesture.ts
@@ -28,6 +28,9 @@ export function useHoverGesture({
     whileHover,
     visualElement,
 }: FeatureProps) {
+    /**
+     * TODO: Make passive: false if onHoverStart doesnt exist
+     */
     usePointerEvent(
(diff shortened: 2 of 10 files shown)


o-alexandrov:
Thank you

packages/framer-motion/src/events/use-dom-event.ts
```diff
@@ -4,7 +4,7 @@ export function addDomEvent(
     target: EventTarget,
     eventName: string,
     handler: EventListener,
-    options?: AddEventListenerOptions
+    options: AddEventListenerOptions = { passive: true }
```
shuangq, line 7: This is not new, but I'm surprised the interface matches `removeEventListener`  (line:11), I expect add & removeEventListener to have different option types 🤔

packages/framer-motion/src/gestures/use-hover-gesture.ts
```diff
@@ -28,19 +28,27 @@ export function useHoverGesture({
     whileHover,
     visualElement,
 }: FeatureProps) {
+    /**
+     * TODO: Make passive: false if onHoverStart doesnt exist
```
shuangq, line 32: What is this TODO for? Shouldn't it be `passive: true` if onHoverStart doesn't exist?

packages/framer-motion/src/gestures/use-tap-gesture.ts
```diff
@@ -59,6 +59,9 @@ export function useTapGesture({
         if (isPressing.current) return
         isPressing.current = true
 
+        /**
+         * TODO: Make passive if tap handlers dont exist?
```
shuangq, line 63: Also here, why aren't we doing it already?

shuangq (approved):
Code looks good 👍  Left a couple of questions:

packages/framer-motion/src/gestures/use-hover-gesture.ts
mattgperry: Ah we are - I forgot to get rid of the TODOs!

packages/framer-motion/src/gestures/use-tap-gesture.ts
mattgperry: And here I forgot to implement too

packages/framer-motion/src/events/use-dom-event.ts
mattgperry: Good point, I think they do, not sure why that wasn't flagging

Issue #1544, opened 2022-05-24
[BUG] Importing framer-motion in a component and testing that component with Cypress and @cypress/vite-dev-server breaks


reintroducing:
**1. Read the [FAQs](#faqs) 👇**

**2. Describe the bug**

Simply importing a framer-motion component in another component that is being component tested using Cypress and @cypress/vite-dev-server completely breaks the component with the following error: 
<img width="360" alt="Screen Shot 2022-05-24 at 10 05 21 AM" src="https://user-images.githubusercontent.com/259901/170092500-ad28f279-0dbe-4586-913c-e62fd70f299d.png">

Commenting out the import of framer-motion causes the tests to pass as expected.

**3. IMPORTANT: Provide a CodeSandbox reproduction of the bug**

I can't provide a code sandbox since Cypress is involved but I've put together a minimal example in this zip: 
[vite-project.zip](https://github.com/framer/motion/files/8764847/vite-project.zip)

**4. Steps to reproduce**

1. Unzip file and run `npm i`.
2. Run `npm test`.
3. Click `Foo.spec.jsx` from the Cypress UI to run the test.
4. Observe the error.
5. Open `src/Foo.jsx` and comment out line 1 (framer-motion import).
6. Go back to the Cypress browser that was loaded and observe the test passing.

What is interesting is that a prior version of @cypress/vite-dev-server (2.2.2) works fine with this (I'm not sure why that is, I've commented on their release of 2.2.3 with this same bug information). That being said, [Vite doesn't use process.env](https://vitejs.dev/guide/env-and-mode.html#env-variables) for environment variables so I imagine the use of process.env.NODE_ENV in framer-motion is the most likely cause of this issue. 

**5. Expected behavior**

The tests pass as expected, even when framer-motion is imported into the component.

**6. Video or screenshots**

N/A

**7. Environment details**

Mac OS 12.3.1, node 16.13.0, npm 8.1.0, Cypress' default Chrome version from their UI.

### Type error with `AnimateSharedLayout`

`AnimateSharedLayout` was deprecated in 5.0. Refer to the [upgrade guide](https://www.framer.com/docs/guide-upgrade/##shared-layout-animations) for instructions on how to remove.

### Preact isn't working

Framer Motion isn't compatible with Preact. 

### `AnimatePresence` isn't working

Have all of its immediate children got a unique `key` prop that **remains the same for that component every render**?

```jsx
// Bad: The index could be given to a different component if the order of items changes
<AnimatePresence>
[...]

PR #1558, opened 2022-06-01
Making safer process.env check

mattgperry:
`process` might be defined, but `env` might not be, which throws in certain environments.

Fixes https://github.com/framer/motion/issues/1544


reintroducing:
@mattgperry i just tried the new version and can confirm it fixes the issue. thank you very much.

Issue #1540, opened 2022-05-23
[BUG] Weird scroll into view behavior with height: 'auto' motion


mattp0123:
### CodeSandbox
This is a forked accordion example from official one. The difference between the original one is that more accordions are added.

https://codesandbox.io/s/framer-motion-accordion-forked-uq6e9o

### Steps to reproduce
Expand the first accordion and scroll to the bottom of the page, then expand the last accordion.
After these steps, you will notice that the page scroll to the top (which the first accordion is at).

### Expected behavior
The page should not scroll to the top.

youngkyo0504:
That problem is associated with only CSS. It looks like scrolling. but actually just container height changed.
You can solve the problem simply by setting the height on the elment. 

`styles.css`
.example-container {
  width: 320px;
  height: 200vh;
  padding: 20px;
}

```css

```

mattp0123:
@youngkyo0504 What if I want the container's height to be auto (depends on children's height)

mattgperry:
Playing around with it, the snapping only happens if the viewport is larger than the height of all the items when they're all closed and smaller when one is open. This is because to animate to auto we do some measurements that might have all elements in the collapsed state for long enough that the scroll resets. It might be possible to store and restore the scroll position during this operation, I'll take a look

mattp0123:
This is my workaround to auto height (using the custom prop of motion element):
```typescript=
function AnimatedBox(props: { children: ReactNode }) {
  const ref = useRef<HTMLDivElement>(null);
  const [height, setHeight] = useState(0);
  useEffect(() => setHeight(ref.current!.scrollHeight), []);
  return (
    <Box
      ref={ref}
      initial={BoxVariant.Collapse}
      animate={BoxVariant.Open}
      exit={BoxVariant.Collapse}
      custom={height}
      variants={{
        [DetailBoxVariant.Open]: (height: number) => ({ height }),
        [DetailBoxVariant.Collapse]: { height: 0 },
      }}
    >
      {props.children}
    </Box>
  );
}

function App() {
  const [isBoxOpen, setIsBoxOpen] = useState(false);
  return (
    <AnimatePresence initial={false}>
      {isBoxOpen && (
        <AnimatedBox>
          <div>Text</div>
        </AnimatedBox>
      )}
    </AnimatePresence>
  );
}
```

PR #1561, opened 2022-06-01
Restore scroll position after measuring height

mattgperry:
When animating between different `height` units like `0` and `auto` we do some DOM measurements that might change the page height in such a way that scroll position resets to `0`. This PR measures and restores the scroll position.

We could do the same with `width` but this is the first time this has come up even with y scroll so it's probably not neccessary.

Fixes https://github.com/framer/motion/issues/1540

### DIFF
--- a/packages/framer-motion/src/render/dom/utils/unit-conversion.ts
+++ b/packages/framer-motion/src/render/dom/utils/unit-conversion.ts
@@ -261,6 +261,11 @@ const checkAndConvertChangedValueTypes = (
     })
 
     if (changedValueTypeKeys.length) {
+        const scrollY =
+            changedValueTypeKeys.indexOf("height") >= 0
+                ? window.pageYOffset
+                : null
+
         const convertedTarget = convertChangedValueTypes(
             target,
             visualElement,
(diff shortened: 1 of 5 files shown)


packages/framer-motion/src/render/dom/utils/unit-conversion.ts
```diff
@@ -277,6 +282,9 @@ const checkAndConvertChangedValueTypes = (
         // Reapply original values
         visualElement.syncRender()
 
+        // Restore scroll position
+        if (scrollY !== null) window.scrollTo({ top: scrollY })
```
shuangq, line 286: Is it possible that the scroll position also changed meanwhile, that we don't wanna restore the prev value?

shuangq (approved):
Code looks good, could we cover a test if it is not too hairy?

packages/framer-motion/src/render/dom/utils/unit-conversion.ts
mattgperry: This is all synchronous, it's not batched like layout animations, much older code, but even so there couldn't be a change in between.

PR #1570, opened 2022-06-03
Fix rotation during layout projection

mattgperry:
Fixes https://github.com/framer/company/issues/24876

The order of scale and rotation was previously incorrect, meaning the applied scale correction would be applied on the rotated axis rather than the screen axis.


shuangq (approved):
LGTM 👍

PR #1571, opened 2022-06-03
Missing parens in `useTransform` example

fturmel:
The arrow function is malformed here. Even though there is technically only one argument, the array destructuring requires parentheses.

Screen Shot 2022-06-03 at 11 40 56 AM

I also want to point out that this example doesn't play well with TypeScript and requires to either cast the arguments passed to the arrow function or use (undocumented?) generics with the `useTransform` function.

Screen Shot 2022-06-03 at 11 45 06 AM

Screen Shot 2022-06-03 at 11 45 43 AM

Is this a type bug? The argument type is determined correctly when a single value is passed.

Screen Shot 2022-06-03 at 11 44 30 AM

### DIFF
--- a/packages/framer-motion/src/value/use-transform.ts
+++ b/packages/framer-motion/src/value/use-transform.ts
@@ -99,7 +99,7 @@ export function useTransform<I, O>(
  * export const MyComponent = () => {
  *   const x = useMotionValue(0)
  *   const y = useMotionValue(0)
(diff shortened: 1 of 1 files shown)


mattgperry (approved):

mattgperry:
As for the example, the examples should be in JS but yeah I've had trouble getting the argument types inferred correctly from the input vars

PR #1573, opened 2022-06-06
Explicitly declare 'children' props for React.FunctionComponent

stkao05:
Version 18 of `@types/react` introduces a breaking change - removal of implicit children props ([ref](https://solverfox.dev/writing/no-implicit-children/)). This introduces type error for some of the components

```sh
error TS2559: Type '{ children: Element; }' has no properties in common with type 'IntrinsicAttributes'.

       <AnimateSharedLayout>
```

The PR update the type definition to reflect this change.

---

Reference
- [[RFC] React 18 and types-only breaking changes](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/46691)
- [Removal Of Implicit Children](https://solverfox.dev/writing/no-implicit-children/)


mattgperry:
Can you reopen with a Codesandbox reproduction of the error? These components all have props arguments already typed with `React.PropsWithChildren`.

stkao05:
>  These components all have props arguments already typed with React.PropsWithChildren.

Because the component was typed as 

```
export const AnimateSharedLayout: React.FunctionComponent = ...
```

The type definition [AnimateSharedLayout.d.ts](https://unpkg.com/framer-motion@6.3.10/types/components/AnimateSharedLayout.d.ts) is outputted as 

```tsx
import * as React from "react";
export declare const AnimateSharedLayout: React.FunctionComponent;
```

Here is also the [Codesandbox](https://codesandbox.io/s/framer-motion-type-issue-nl27u8?file=/src/App.tsx) that demonstrates the type issue. Note that the type error only shows with the latest typing enabled `@type/react: 18.0.12`

mattgperry:
Thanks for the repro!

mattgperry (approved):

PR #1574, opened 2022-06-07
Split scale correction and element resize scales

mattgperry:
The last PR had a problem where it induced a skew to elements as the tree scale correction and element resize scales need to happen on pre and post rotated axes. This PR splits them out


shuangq (approved):
Nice fix 👏

PR #1586, opened 2022-06-20
Rollup .d.ts files

eelco:
Rolls up the .d.ts files into a single entry.


packages/framer-motion-3d/src/utils/use-time.ts
```diff
@@ -4,7 +4,7 @@ import { useContext } from "react"
 
 export function useTime() {
     const time = useMotionValue(0)
-    const { isStatic } = useContext(MotionConfigContext)
+    const isStatic = useContext(MotionConfigContext)["isStatic"] // Internal API
```
eelco, line 7: I think this is the only real hack I had to do to dip into the internals without too much trickery.

mattgperry (approved):

Issue #1589, opened 2022-06-22
[BUG] Type error: Property 'set' does not exist on type 'AnimationControls' in v6.3.12


eli-s-r:
**1. Read the [FAQs](#faqs) 👇**

**2. Describe the bug**

In version [6.3.12](https://github.com/framer/motion/releases/tag/v6.3.12), the [`set`](https://www.framer.com/docs/utilities/###set) method on `AnimationControls` seems to have been removed from the typedefs. As far as I can tell, there is no documentation of this method intentionally being deprecated or removed.

**3. IMPORTANT: Provide a CodeSandbox reproduction of the bug**

https://codesandbox.io/s/objective-wiles-3w2w7e?file=/src/App.tsx

**4. Steps to reproduce**

1. Import the `useAnimation` hook and attempt to call the `set` method on the return value (or on any object of type `AnimationControls`).

**5. Expected behavior**

No compilation error.

PR #1590, opened 2022-06-22
Changes TSDoc tags for internal remarks to prevent the declaration to be considered internal

bdriguesdev:
Using the [stripInternal](https://www.typescriptlang.org/tsconfig#stripInternal) to remove internal declarations means that declarations with the tag `@internal` will be removed, but this also means that if a tag starts with `internal` it will also be removed from the declarations. Therefore, when using the `@internalRemarks` tag in a declaration, it will be considered an internal declaration and will be removed.

To avoid this, I used the tag [@privateRemarks](https://tsdoc.org/pages/tags/privateremarks/) instead of `@internalRemarks`. This new tag doesn't prevent the remark from being published in the declaration, but with the previous tag(`@internalRemarks`), we weren't preventing it either.

Fixes https://github.com/framer/motion/issues/1589

### DIFF
--- a/packages/framer-motion/src/animation/types.ts
+++ b/packages/framer-motion/src/animation/types.ts
@@ -67,7 +67,7 @@ export interface AnimationControls {
      * controls.set("hidden")
      * ```
      *
-     * @internalremarks
+     * @privateRemarks
      * We could perform a similar trick to `.start` where this can be called before mount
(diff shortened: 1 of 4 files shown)


mattgperry:
Great to know, thanks for this!

mattgperry (approved):

Issue #1535, opened 2022-05-20
[BUG] Layout animations don't work properly in sticky containers


two-legs:
**1. Describe the bug**
I have Reorder layout in sticky container (also it can be reproduced in fixed container). When I mix items reordering and main page scroll reorder animation is broking sometimes. 

Seems like layout projection for stuck elements uses window scroll in calculations but it shouldn't be done for fixed or stuck elements. 

**3. IMPORTANT: Provide a CodeSandbox reproduction of the bug**

https://codesandbox.io/s/framer-motion-5-drag-to-reorder-lists-forked-q9cixq

**4. Steps to reproduce**

1. Reorder items
2. Scroll page 
3. Repeat

**5. Expected behavior**

There shouldn't be animation glitches

**6. Video or screenshots**

https://user-images.githubusercontent.com/9340251/169492501-b38a84e2-c5b6-4c1f-a8a6-928d19d8079b.mov

**7. Environment details**
macOS 12.3 
Chrome 101.0.4951.64 (Official Build) (arm64)

mattgperry:
In the next couple weeks I want to introduce a prop which will allow us to tag elements that are position sticky/fixed and account for this in the layout animations, so hopefully we can find a fix for this soon

equinusocio:
Same issue here, but more visible. Here a reproduction site:

0. The header is sticky via CSS
1. [Go here](https://staging.design.wonderflow.ai/documentation/release-notes) and scroll down
2. Change to playground page from the top nav
3. You see the highlight element is animated from the bottom

This happens only if there page (body) scrolled.

pie6k:
@mattgperry is any workaround possible in the meanwhile?

isc30:
I believe its possible to go up the DOM using `getComputedStyle` and the first css `position` that is set is the actual one. This way we wouldn't need a specific prop to mark this case.

For example if I have a reusable component with a layout animation inside, the application developer can use it and add `position: sticky`. How would they pass the new prop to the nested `motion.div`?

If we can detect this case automatically, its a massive win

mattgperry:
@isc30 But then you'd have to, for every node involved in the layout animation, traverse up the tree and hit `getComputedStyle`. The will be performance implications there that have to be looked at as without some consideration to mitigate/optimise this it wouldn't scale well.

isc30:
[...]

PR #1594, opened 2022-06-23
Fix layout animations in position: fixed

mattgperry:
As per https://github.com/framer/motion/issues/1535, layout animations break in `position: fixed` and `scroll` elements.

This PR makes it possible to tag `position: fixed` elements with the `layoutScroll` prop. Elements with the `layoutScroll` prop are now checked to see if they are also `position: fixed`. If so, all parent scrolls will be disregarded.

I haven't applied to `sticky` for now as more research is needed to see if we can determine when the element is in sticky mode vs inline.

### DIFF
--- a/dev/projection/single-element-page-scroll-overlay.html
+++ b/dev/projection/single-element-page-scroll-overlay.html
@@ -1,0 +1,68 @@
+<html>
+    <head>
+        <style>
+            body {
+                padding: 0;
+                margin: 0;
+            }
+
+            #box {
+                width: 100px;
+                height: 100px;
+                background-color: #00cc88;
+            }
+
+            #overlay {
+                position: fixed;
+                inset: 0;
+            }
+
+            #trigger-overflow {
+                width: 1px;
+                height: 1px;
+                position: absolute;
+                top: 2000px;
+                left: 2000px;
+            }
+
+            [data-layout-correct="false"] {
+                background: #dd1144 !important;
+                opacity: 0.5;
+            }
+        </style>
+    </head>
+    <body>
+        <div id="overlay">
+            <div id="box"></div>
+        </div>
+        <div id="trigger-overflow"></div>
+
+        <script src="../../packages/framer-motion/dist/projection.dev.js"></script>
+        <script src="./script-assert.js"></script>
+        <script src="./script-undo.js"></script>
+        <script>
+            const { createNode } = window.Undo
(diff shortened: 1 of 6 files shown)


packages/framer-motion/src/projection/node/HTMLProjectionNode.ts
```diff
@@ -23,4 +23,6 @@ export const HTMLProjectionNode = createProjectionNode<HTMLElement>({
     resetTransform: (instance, value) => {
         instance.style.transform = value ?? "none"
     },
+    checkIsScrollRoot: (instance) =>
+        Boolean(window.getComputedStyle(instance).position === "fixed"),
```
isc30, line 27:
```suggestion
{
        const position = window.getComputedStyle(instance).position
        return Boolean(position  === "fixed" || position  === "sticky")
}
```

isc30: By adding this change, it will also fix `sticky` positioned elements that use the `body` scroll.
As forthe other sticky elements, they will be broken either way like before so its worth having this change here

isc30 (changes requested):

packages/framer-motion/src/projection/node/HTMLProjectionNode.ts
isc30: ah true, the scroll position will not help here.
I usually check if the sticky is stickying using intersectionobserver but that will kill the performance of a generic implementation

equinusocio:
@mattgperry i saw this long time ago, maybe is helpful to detect when element become sticky
https://developer.chrome.com/blog/sticky-headers/

equinusocio:
Any ETA for a release with this fix?

dev/projection/single-element-page-scroll-overlay.html
```diff
@@ -0,0 +1,68 @@
...
+            const boxOrigin = box.getBoundingClientRect()
+
+            boxProjection.willUpdate()
+
+            const scrollOffset = [50, 100]
+            window.scrollTo(...scrollOffset)
+
+            boxProjection.root.didUpdate()
+
+            matchViewportBox(box, { top: 0, left: 0, bottom: 100, right: 100 })
```
shuangq, line 65: Might be helpful to also check if the position fixed element (overlay here) changes its position?

PR #1617, opened 2022-07-14
Updating useTransform to subscribe on layout effect

mattgperry:


packages/framer-motion/src/value/use-on-change.ts
```diff
@@ -1,18 +1,18 @@
-import { useEffect } from "react"
+import { useIsomorphicLayoutEffect } from "../three-entry"
```
huntercaron, line 1: @mattgperry importing this from `../three-entry` seems like it would create a circular import or something wouldn't it?

PR #1627, opened 2022-07-21
Adding deprecated tag to use-viewport-scroll and use-element-scroll

alexandredev3:
Adding the deprecated tag will make the IDE fire a warning not to use these hooks anymore.
scroll-hooks-deprecated


mattgperry:
Great addition, thanks!

mattgperry (approved):

PR #1640, opened 2022-08-02
docs: add installation with `yarn` and `pnpm`

dammy001:


mattgperry:
Hey thanks for the contribution - for simplicity though I just keep this to npm. If you're using another package manager you know how to translate that into your own installation instruction. I'll add a little sentence so it's a little less terse though.

dammy001:
Okay 👍 great then

PR #1641, opened 2022-08-02
Upgrading to React 18

mattgperry:


packages/framer-motion-3d/src/components/use-layout-camera.ts
```diff
@@ -11,6 +11,8 @@ import { clamp } from "popmotion"
 const calcBoxSize = ({ x, y }: Box) => ({
     width: calcLength(x),
     height: calcLength(y),
+    top: 0,
+    left: 0,
```
nvh, line 15: Why were these omitted before?

packages/framer-motion-3d/src/render/utils/set-value.ts
```diff
@@ -57,6 +57,10 @@ export function setThreeValue(
     if (setters[key]) {
         setters[key](instance, values[key])
     } else {
+        if (key === "opacity" && !instance.transparent) {
+            instance.transparent = true
+        }
```
nvh, line 62: Is this a driveby fix?

packages/framer-motion/src/utils/__tests__/use-cycle.test.tsx
```diff
@@ -23,12 +23,6 @@ describe("useCycle", () => {
         click(container.firstChild as Element)
         click(container.firstChild as Element)
 
-        /**
-         * 1 is doubled at the start of the array because in StrictMode
-         * useEffect is double-fired on mount.
-         *
-         * When resolution is changed to 18, add another 1 to start of array
-         */
         expect(results).toEqual([1, 2, 3, 4, 1])
```
nvh, line 26: You did not implement this comment? 🤔

dev/examples/ThreeBox.tsx
```diff
@@ -43,11 +61,9 @@ function Box(props) {
 
 export const App = () => {
     const [isHovered, setHover] = useState(false)
+    console.log({ isHovered })
```
nvh, line 64: Stray console.log

dev/examples/ThreeBox.tsx
```diff
@@ -43,11 +61,9 @@ function Box(props) {
...
     return (
-        <MotionConfig
-            isStatic
-            transition={{ type: "spring", bounce: 0, duration: 0.7 }}
-        >
+        <MotionConfig transition={{ type: "spring", bounce: 0, duration: 0.7 }}>
```
nvh, line 66: Why is the `isStatic` removed?

dev/examples/ThreeBox.tsx
```diff
@@ -11,15 +29,15 @@ function Box(props) {
...
+    // const scale = useTransform(time, (t) => Math.sin(t) * 0.5 + 1)
 
     // Subscribe this component to the render-loop, rotate the mesh every frame
     // useFrame((state, delta) => (ref.current.rotation.x += 0.01))
     // Return the view, these are regular Threejs elements expressed in JSX
     return (
         <motion.mesh
             {...props}
-            scale={scale}
+            // scale={scale}
```
nvh, line 40: Are these comments intentional?

nvh (commented):
There's a ton of tiny remarks, but I think it's good overall

packages/framer-motion-3d/src/components/use-layout-camera.ts
mattgperry: The type has changed in R3F

packages/framer-motion-3d/src/render/utils/set-value.ts
mattgperry: No - either the latest version of Three.js or R3F doesn't set `transparent` to `true` by default.

packages/framer-motion/src/utils/__tests__/use-cycle.test.tsx
mattgperry: I changed the test instead

dev/examples/ThreeBox.tsx
mattgperry: Thanks - fixed

dev/examples/ThreeBox.tsx
mattgperry: I'm not sure why it was there tbh, probably mucking about with passing context via `MotionCanvas`. But it was disabling the animations in the sandbox.

dev/examples/ThreeBox.tsx
mattgperry: Yeah - it doesn't make sense to mix the `useTime` powered scale with animations, I could remove but it's quite interesting to play with

nvh: Maybe add a comment why it's commented out, so other people know what to do with it too

nvh:
Github won't let me approve, but here's a ✅

Issue #1518, opened 2022-04-18
[BUG] ReorderGroup not work correctly on React 18


debugksir:
codesandbox link:

https://codesandbox.io/s/framermotionbugreport-12vxmn?file=/src/index.tsx

mattgperry:
I had a brief look into this and so far I think this is happening specifically when using `useDragControls`.

c-ehrlich:
> I had a brief look into this and so far I think this is happening specifically when using `useDragControls`.

I'm also experiencing this bug in a project since upgrading to React 18, and agree with your conclusion - the issue appears only when using `useDragControls`.

19ozman:
CKrook have you seen this?

ckrook:
@19ozman  Just saw this during our daily standup. Are there any solutions in sight?

19ozman:
@ckrook did you also try the Magnum?

Systemcluster:
Note that this only appears to be the case when StrictMode is used and in development mode. This does not seem to affect production builds.

19ozman:
Cool! It's working now!

maltesa:
Also happens without dragControls (only in strict-mode during dev):
https://codesandbox.io/s/framermotionbugreport-forked-7fwnrj?file=/src/Item.tsx

mtysor:
> Also happens without dragControls (only in strict-mode during dev):
> https://codesandbox.io/s/framermotionbugreport-forked-7fwnrj?file=/src/Item.tsx
This looks to be related to [strict effects](https://github.com/reactwg/react-18/discussions/19) which double-invokes  useEffect. Unfortunately I suspect this affects several packages. Good news is this will still work on prod, and will work locally by disabling StrictMode

heymartinadams:
An easy way to not use `useDragControls` and therefore sidestep this bug for now is to do the following:

```js
const [items, setItems] = useState([])
const [draggable, setDraggable] = useState(false)

<Reorder.Group axis='y' onReorder={setItems} values={items}>
  <Reorder.Item
    dragListener={draggable}
    value={item}
  >
    <div
      onMouseEnter={() => setDraggable(true)}
      onMouseLeave={() => setDraggable(false)}
      onTouchStart={() => setDraggable(true)}
      onTouchEnd={() => setDraggable(false)}
    >
      <ReorderIcon />
    </div>

    {item}
  </Reorder.Item>
</Reorder.Group>
```

minlare:
[...]

PR #1643, opened 2022-08-03
Fixing drag to reorder

mattgperry:
Bundling this into the `7.0.0` release as e2e tests don't actually pass without this fix. The solution here was to ensure that if the didUpdate event listener was double-applied (as it is in StrictMode) then it unsubscribes correctly first.

Closes https://github.com/framer/motion/issues/1518

### DIFF
--- a/packages/framer-motion/src/components/Reorder/Group.tsx
+++ b/packages/framer-motion/src/components/Reorder/Group.tsx
@@ -71,7 +71,9 @@ export function ReorderGroup<V>(
     }: Props<V> & HTMLMotionProps<any> & React.PropsWithChildren<{}>,
     externalRef?: React.Ref<any>
 ) {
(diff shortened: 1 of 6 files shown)


packages/framer-motion/src/components/Reorder/Group.tsx
```diff
@@ -71,7 +71,9 @@ export function ReorderGroup<V>(
     }: Props<V> & HTMLMotionProps<any> & React.PropsWithChildren<{}>,
     externalRef?: React.Ref<any>
 ) {
-    const Component = useConstant(() => motion(as)) as FunctionComponent<React.PropsWithChildren<HTMLMotionProps<any> & { ref?: React.Ref<any> }>>
+    const Component = useConstant(() => motion(as)) as FunctionComponent<
```
mattgperry, line 74: This is probably from the prettier bump in the previous PR

PR #1648, opened 2022-08-08
Adding `useWillChange` to apply `will-change` to animating values

mattgperry:


dev/examples/Prop-style.tsx
```diff
@@ -15,7 +15,7 @@ const style = {
 
 export const App = () => {
     return (
-        <motion.div static>
+        <motion.div>
```
mattgperry, line 18: `static` no longer exists as a prop - change unrelated to PR

packages/framer-motion/src/render/utils/animation-state.ts
```diff
@@ -82,10 +80,6 @@ export function createAnimationState(
         return acc
     }
 
-    function isAnimated(key: string) {
```
mattgperry, line 85: Unused API

packages/framer-motion/src/value/use-will-change/__tests__/index.test.tsx
```diff
@@ -0,0 +1,87 @@
...
+                    expect(container.firstChild).toHaveStyle(
+                        "will-change: transform, background-color;"
+                    )
+                    resolve()
+                })
+            })
+        })
+    })
+
+    test("Removes `transform` when all transforms finish animating", async () => {
```
shuangq, line 63: We could maybe cover one more test to verify that when all the motion values (not just transforms) finish animating, will change is revert to "auto"?

shuangq (approved):
Left some questions, but code looks good in general 👍

packages/framer-motion/src/value/use-will-change/index.ts
```diff
@@ -0,0 +1,56 @@
...
+        } else if (
+            !isTransformOriginProp(name) &&
+            !isCSSVariable(name) &&
+            name !== "willChange"
+        ) {
+            memberName = camelToDash(name)
+        }
+
+        if (memberName) {
+            addUniqueItem(this.members, memberName)
```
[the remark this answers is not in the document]
mattgperry: It's more just so I can use `join` more conveniently when we make the `will-change` string

packages/framer-motion/src/utils/array.ts
```diff
@@ -1,10 +1,16 @@
 export function addUniqueItem<T>(arr: T[], item: T) {
-    arr.indexOf(item) === -1 && arr.push(item)
+    if (arr.indexOf(item) === -1) {
+        arr.push(item)
+        return true
```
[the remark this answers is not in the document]
mattgperry: This is actually unnecessary now. I was using to to return from the function whether the operation was successful. But not using this functionality any more so I can take it out

Issue #296, opened 2019-08-27
[FEATURE] AnimatePresence popFromFlow


mattgperry:
In `PoseGroup`, we had the `flipMove` functionality where exiting components would be popped from the document flow. To do this we'd measure the element, apply `position: absolute` and then attempt to visually reposition the element. 

This allowed us to do stuff like simultaneously animate in and out statically positioned divs without them taking up each other's space.

However due to the nature of measuring and reconstructing the visual position of these elements, potentially something to do with not account for margins of internal elements, this often resulted in unexpected behaviour. If we reintroduce this behaviour I'd like to ensure all of these are accounted for.

PR #1657, opened 2022-08-11
Pop exiting elements from the document flow

mattgperry:
This PR adds a `mode` prop to `AnimatePresence`.

When `mode="popLayout"` is enabled, exiting elements will be "popped" from the layout. This allows surrounding `layout` elements to immediately animate into the space previously occupied by the exiting element, rather than waiting for it to animate out completely.

`exitBeforeEnter` becomes deprecated and replaced with `mode="wait"`

Fixes https://github.com/framer/motion/issues/296

### DIFF
--- a/packages/framer-motion/src/components/AnimatePresence/PopChild.tsx
+++ b/packages/framer-motion/src/components/AnimatePresence/PopChild.tsx
@@ -4,6 +4,8 @@ import { useRef, useInsertionEffect, useId } from "react"
 interface Size {
     width: number
     height: number
+    top: `${number}px` | "auto"
+    left: `${number}px` | "auto"
 }
 
 interface Props {
@@ -22,11 +24,29 @@ interface MeasureProps extends Props {
  */
 class PopChildMeasure extends React.Component<MeasureProps> {
     getSnapshotBeforeUpdate(prevProps: MeasureProps) {
-        if (prevProps.isPresent && !this.props.isPresent) {
-            const element = this.props.childRef.current
-            const size = this.props.sizeRef.current
-            size!.height = element?.offsetHeight || 0
-            size!.width = element?.offsetWidth || 0
+        const element = this.props.childRef.current
+        if (element && prevProps.isPresent && !this.props.isPresent) {
+            const { position } = getComputedStyle(element)
+            const size = this.props.sizeRef.current!
+            size.height = element.offsetHeight || 0
+            size.width = element.offsetWidth || 0
+            size.top = size.left = "auto"
+
+            /**
+             * If this element is position: relative and the parent has
+             * padding, we need to explicitly set a top and/or left.
+             */
+            if (position === "relative" && element.offsetParent) {
+                const { paddingTop, paddingLeft } = getComputedStyle(
+                    element.offsetParent
+                )
+                if (parseFloat(paddingTop)) {
+                    size.top = `${element.offsetTop}px`
+                }
+                if (parseFloat(paddingLeft)) {
+                    size.left = `${element.offsetLeft}px`
+                }
+            }
         }
 
         return null
@@ -45,7 +65,12 @@ class PopChildMeasure extends React.Component<MeasureProps> {
 export function PopChild({ children, isPresent }: Props) {
     const id = useId()
     const ref = useRef<HTMLElement>(null)
-    const size = useRef({ width: 0, height: 0 })
+    const size = useRef<Size>({
+        width: 0,
+        height: 0,
+        top: "auto",
+        left: "auto",
+    })
 
     /**
      * We create and inject a style block so we can apply this explicit
@@ -57,7 +82,7 @@ export function PopChild({ children, isPresent }: Props) {
      * styles set via the style prop.
      */
     useInsertionEffect(() => {
-        const { width, height } = size.current
+        const { width, height, top, left } = size.current
         if (isPresent || !ref.current || !width || !height) return
 
         ref.current.dataset.motionPopId = id
@@ -70,6 +95,8 @@ export function PopChild({ children, isPresent }: Props) {
             position: absolute !important;
             width: ${width}px !important;
             height: ${height}px !important;
+            top: ${top} !important;
+            left: ${left} !important;
           }
         `)
         )
--- a/packages/framer-motion/src/components/AnimatePresence/PopChild.tsx
+++ b/packages/framer-motion/src/components/AnimatePresence/PopChild.tsx
@@ -16,6 +16,10 @@ interface MeasureProps extends Props {
     sizeRef: React.RefObject<Size>
 }
 
+/**
+ * Measurement functionality has to be within a separate component
+ * to leverage snapshot lifecycle.
+ */
 class PopChildMeasure extends React.Component<MeasureProps> {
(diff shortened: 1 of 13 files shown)


packages/framer-motion/src/components/AnimatePresence/index.tsx
```diff
@@ -75,13 +75,16 @@ function onlyElements(children: ReactNode): ReactElement<any>[] {
...
+export const AnimatePresence: React.FunctionComponent<
+    React.PropsWithChildren<AnimatePresenceProps>
+> = ({
     children,
     custom,
     initial = true,
     onExitComplete,
     exitBeforeEnter,
     presenceAffectsLayout = true,
+    popLayout = false,
```
nvh, line 87: I don't see the `custom` prop being used anywhere, but I was thinking, couldn't we make something like:
`transition={"pop"}`, so we allow for flexibility into the future?

packages/framer-motion/src/components/AnimatePresence/PopChild.tsx
```diff
@@ -0,0 +1,87 @@
...
+        const style = document.createElement("style")
+        document.head.appendChild(style)
+        style.appendChild(
+            document.createTextNode(`
+          [data-motion-pop-id="${id}"] {
+            position: absolute !important;
+            width: ${width}px !important;
+            height: ${height}px !important;
+          }
+        `)
```
nvh, line 74: If I remember correctly, using the DOM api's to insert CSS rules was much faster than setting the text of a style node:
```suggestion
        style.insertRule(`[data-motion-pop-id="${id}"] {
            position: absolute !important;
            width: ${width}px !important;
            height: ${height}px !important;`)
```

Like we do here: https://github.com/framer/FramerStudio/blob/fff2bbf1ab45fb823caf98c59bfdc0f5e3c4604a/src/modules/withCSS.tsx/withCSS.tsx#L8-L34

nvh: Also good practice to set the type on the style rule:
```typescript
        styleElement.setAttribute("type", "text/css")
```

nvh (commented):
I think we need to think about the API a bit. Can we make a more generic API on `AnimatePresence` that will allow for adding different animations in the future?

packages/framer-motion/src/components/AnimatePresence/PopChild.tsx
mattgperry: Ah amazing thanks, I'll replace with this - although I think `text/css` is deprecated now?

packages/framer-motion/src/components/AnimatePresence/index.tsx
mattgperry: In theory I would be open to this but it seems to me like a binary behaviour - either elements are popped from layout while exiting or they're not. I can't imagine the "category" of behaviour this lives within, ie its not a transition or animation, nor how it would be incompatible with those other behaviours as a boolean.

nvh: I misunderstood what the prop was doing, it seems fine like a boolean, but is there a better name than `pop` for it? I can't really come up with it, but might be worth trying a bit of other options.

packages/framer-motion/src/components/AnimatePresence/types.ts
```diff
@@ -51,10 +51,18 @@ export interface AnimatePresenceProps {
...
      */
     exitBeforeEnter?: boolean
 
+    /**
+     * If set to `true`, top-level `exit` components will be "popped" out of the DOM layout,
+     * so surrounding elements can animate to their new layout immediately.
+     *
+     * @public
+     */
+    popLayout?: boolean
```
nvh, line 64: How does this interact with `exitBeforeEnter` and `presenceAffectsLayout`?
Just trying a bit:
- `popOnExit`
- `exitImmediately`
- `removeOnExit`
- `exitPops`

packages/framer-motion/src/components/AnimatePresence/PopChild.tsx
nvh: Ah, did not know that, would trust other sources more than me about it 😉

packages/framer-motion/src/components/AnimatePresence/types.ts
nvh: Is there maybe a way to combine into `exitBehaviour: "pop" | "beforeEnter"`?

nvh: `appearanceBehaviour: "popExit" | "deferredEnter"`

nvh: ~`method`~
~`action`~
`mode: "sync" | "popLayout" | "deferEnter"`
~`switch: "default" | "popLayoutExit" | "deferEnter"`~
~`swap`~

nvh: Default is `sync`

packages/framer-motion/src/components/AnimatePresence/index.tsx
```diff
@@ -75,14 +76,23 @@ function onlyElements(children: ReactNode): ReactElement<any>[] {
...
     initial = true,
     onExitComplete,
     exitBeforeEnter,
     presenceAffectsLayout = true,
+    mode = "sync",
 }) => {
+    // Support deprecated exitBeforeEnter prop
+    if (exitBeforeEnter) {
+        mode = "wait"
+        warnOnce(false, "Replace exitBeforeEnter with mode='wait'")
```
nvh, line 93: 👏🏻

packages/framer-motion/src/components/AnimatePresence/types.ts
```diff
@@ -51,10 +51,25 @@ export interface AnimatePresenceProps {
      * )
      * ```
      *
-     * @beta
+     * @deprecated
+     *
+     * Replace with `mode="popLayout"`
```
nvh, line 56: This should be `mode="wait"`, right?

nvh, line 56:
```suggestion
     * Replace with `mode="wait"`
```

packages/framer-motion/src/components/AnimatePresence/types.ts
```diff
@@ -51,10 +51,25 @@ export interface AnimatePresenceProps {
...
     exitBeforeEnter?: boolean
 
+    /**
+     * Determines how to handle entering and exiting elements.
+     *
+     * - `"sync"`: Default. Elements animate in and out as soon as they're added/removed.
+     * - `"popLayout"`: Exiting elements are "popped" from the page layout, allowing sibling
+     *      elements to immediately occupy their new layouts.
+     * - `"wait"`: Only renders one component at a time. Wait for the exiting component to animate out
+     *      before animating the next component in.
```
nvh, line 67: It would be great to add a link to a code sandbox or similar example, to show the three different types.

mattgperry: I'll be putting sandboxes in the documentation but would prefer to reduce the places where I put these links

nvh (approved):

PR #1660, opened 2022-08-12
Add missing constraint

RyanCavanaugh:
Adds a missing constraint caught by a bugfix in TS4.8; see https://github.com/microsoft/TypeScript/issues/49489


mattgperry:
Awesome, thanks!

mattgperry (approved):

PR #1670, opened 2022-08-18
File size reductions

mattgperry:
- Adds bundlesize visualisation
- Changes Typescript target to es2018, this removes transpilation of rest parameters, which enjoy similar browser support to the previous target change.
- Refactor in `transform.ts` to further reduce filesize


packages/framer-motion/src/render/html/utils/build-transform.ts
```diff
@@ -31,20 +37,12 @@ export function buildTransform(
...
-    for (let i = 0; i < numTransformKeys; i++) {
-        const key = transformKeys[i]
+    for (const key of transformKeys) {
         transformString += `${translateAlias[key] || key}(${transform[key]}) `
-
-        if (key === "z") transformHasZ = true
     }
 
-    if (!transformHasZ && enableHardwareAcceleration) {
+    if (enableHardwareAcceleration && !transform.z) {
```
shuangq, line 45: Should this be `transform.z !== undefined`? It can be 0 right 🤔

packages/framer-motion/src/render/html/utils/build-transform.ts
```diff
@@ -63,15 +61,3 @@ export function buildTransform(
 
     return transformString
 }
-
-/**
- * Build a transformOrigin style. Uses the same defaults as the browser for
- * undefined origins.
- */
```
shuangq, line 70: Let's also move the comment over

packages/framer-motion/src/render/html/utils/build-styles.ts
```diff
@@ -16,9 +16,6 @@ export function buildHTMLStyles(
 ) {
     const { style, vars, transform, transformKeys, transformOrigin } = state
 
-    // Empty the transformKeys array. As we're throwing out refs to its items
-    // this might not be as cheap as suspected. Maybe using the array as a buffer
-    // with a manual incrementation would be better.
```
shuangq, line 21: Why did we remove the comment here? Is it no longer the case?

packages/framer-motion/src/render/html/utils/build-styles.ts
```diff
@@ -59,31 +56,38 @@ export function buildHTMLStyles(
...
             hasTransformOrigin = true
+
+            transformOrigin[key] = valueAsType
         } else {
             style[key] = valueAsType
         }
     }
 
-    if (hasTransform) {
+    if (hasTransform || transformTemplate) {
```
shuangq, line 70: Does `buildTransform` output the same result as `transformTemplate({}, "")`? Also why do we need the `transformTemplate` argument? It doesn't seem to be used at all in `buildTransform` 👀

shuangq (approved):
Left some minor comments, but code looks good in general

packages/framer-motion/src/render/html/utils/build-transform.ts
mattgperry: Ah yeah but in this case if it were `0` it wouldn't have been included in the list and this conditional is adding `translateZ(0)` anyway

packages/framer-motion/src/render/html/utils/build-styles.ts
mattgperry: Yeah I re-read it and thought wow that was a stupid idea

packages/framer-motion/src/render/html/utils/build-styles.ts
mattgperry: It is, near the bottom (line 53)

But yeah this would basically be the same thing as passing `buildTransform` an empty object

packages/framer-motion/src/render/html/utils/build-transform.ts
shuangq: But in case it was 0, we would be adding double `translateZ(0)`, right? (on line:42 `transformString += `${translateAlias[key] || key}(${transform[key]})`)

packages/framer-motion/src/render/html/utils/build-styles.ts
shuangq: If the state we pass in here does have a non-empty `transfrom`, it wouldn't be the same no? 🤔 
```
if (transformTemplate) {
        transformString = transformTemplate(
            transform,
            transformIsDefault ? "" : transformString
        )
    } 
```

packages/framer-motion/src/render/html/utils/build-transform.ts
mattgperry: Yeah exactly - if its `0` it's already been added and 

```
    if (enableHardwareAcceleration && !transform.z) {
        transformString += "translateZ(0)"
    }
```

Doesn't actually run (because `!transform.z`), so we don't add a double `translateZ(0)`

packages/framer-motion/src/render/html/utils/build-styles.ts
mattgperry: Yeah but the previous situation was

- If we have a transform, `buildTransform`
  - If we have a `transformTemplate`, use that
- If we don't have a transform, but we do have a `transformTemplate`, use that

Whereas now it's
- `buildTransform`
  - If we have a `transformTemplate`, use that
  
Which are the same.

shuangq: Ah I see, I thought the arguments we passed in were a bit different, but I missed that if we don't have a transform, `transformIsDefault` will be true, so we will be passing in an empty string anyway. Thanks for explaining!

PR #1679, opened 2022-08-23
`mSVG` component

mattgperry:
This PR splits SVG functionality out of the `m` component into its own `msvg` component, bringing the bundlesize of `m` down to 3.7kb.


mattgperry:
Closing this out as the bundlesize saving only gets shifted to the feature bundles, which is good in some ways (can be lazy-loaded) but maybe not worth it in exchange for the added internal and external complexity.

Issue #1400, opened 2021-12-23
[FEATURE] Multi-axis support in `Reorder.Group`


tommoor:
**Is your feature request related to a problem? Please describe.**

The new `Reorder` components are great for horizontal and vertical lists, it would be fantastic to have first class support in the same api for re-orderable grids. I also raised this on [twitter](https://twitter.com/tommoor/status/1474050314693136385) as the wording in the docs can be read as if this feature already exists.

**Describe the solution you'd like**

Ideally the `axis` prop on `Reorder.Group` would take an additional option `axis="xy"`, or similar. It could be that the reality of dual axis makes the interface more complex, but this would be the _ideal_.

**Describe alternatives you've considered**

The alternative would be custom building using the drag and drop primitives as pre-5.

nanxiaobei:
Really need this!

AndrejGajdos:
would be great having this feature

chocobuckle:
That "To allow dragging on both axes, pass the drag prop to child Reorder.Item components." line in the Reorder.Group props section completely threw me for the past 2 days. I too was under the impression that xy reordering was supported after reading that, only to now stumble upon the Twitter thead linked to above. 

Would be fantastic if this could be added.

cjoecker:
Any update on this?

cjoecker:
Just submitted a Pull Request. This is how it is going to look:
![Reorder](https://user-images.githubusercontent.com/46030630/188180862-a95218de-afe3-43a2-99a8-0bac8b00c800.gif)

flackeryy:
Updates here ?

flackeryy:
Yeah I see... however I have almost working example on codesandbox with custom drag dandlers, but it has some problems that i'm trying to resolve... So wanted to check if there is something ready from the box. 

So i'll try to resolve bugs of my version :(

cjoecker:
You could also copy-paste what I made and use [patch-package](https://www.npmjs.com/package/patch-package) to save the changes in your `node-modules`

flackeryy:
@cjoecker 

just in case, maybe you know a solution..

If you could look at this [very old variant](https://codesandbox.io/s/framer-motion-grid-cc859) that I found, it looks like the old `onDrag` handler on each box, returns `point` parameter that looks **_like a dragged distance from the initial position of a component_**. 

[...]

PR #1685, opened 2022-09-01
add multidimensional reorder

cjoecker:
As asked in #1400 , here is a PR to allow multidimensional reorder

For now, the user just needs to give the number of items per row. 
In another PR, we can calculate it using the children's refs.

For the docs, do I need to create another PR to adapt it?


mattgperry:
Hey @cjoecker, thanks for your submission! However I think there's a problem here with the `itemsPerAxis` prop. This doesn't mesh well with dynamic/responsive layouts. Rather than adding a prop the next step for this API would ideally be to remove the `axis` prop. We know the positions of all the children and the order in which they're rendered. So from that we should be able to work out how the list should be sortable.

cjoecker:
Hey @mattgperry , I removed the `axis` and `itemsPerAxis` prop 👏🏻  We should write in the documentation that the multidimensional reordering is only working for items of the same size. 
This can be improved in the future but I keep it simple for now since the logic got pretty complex already. 

Now it is fully responsive. I would just need your help for making it possible to use external refs. I couldn't make it possible to read data from the `reorder` `ref` but also allow an external `ref`.

Can you also tell me how to fix the pipeline? It is asking for a `config` but I couldn't see anything in the documentation about it.

cjoecker:
The PR is ready. I'm just waiting for the maintainers to review it :)

TheTimeWalker:
I'm currently testing this PR out and I have noticed that there seems to be an issue when there's only one item then an uncaught error is being thrown: At least two children components are necessary. This also goes further that when you try to have something like this e.g.

```
        <Reorder.Group
          values={sort}
          onReorder={(newOrder: string[]) => {
            console.log(newOrder);
          }}
        >
          <Reorder.Item key={'value'} value={'value'}>
            <span>test</span>
          </Reorder.Item>
        </Reorder.Group>
```
Then Typescript complains that only a single child has been provided.

cjoecker:
@TheTimeWalker I made that on purpose [here](https://github.com/cjoecker/motion/blob/fdf86df95bb24e4bb035692ae452d0659b423f63/packages/framer-motion/src/components/Reorder/utils/useOrderingParameters.ts#L15-L18).
My idea was to throw an error when `Reorder.Group` has only one child. Like that, the developers will know what is wrong.

TheTimeWalker:
Ah, I see. But wouldn't that break anything when the list only has one item or when <AnimatePresence> is used? This would essentially break examples like this one: https://www.framer.com/docs/reorder/##layout-animations

cjoecker:
@TheTimeWalker good point! Is just fixed it.

shandysulen:
just wanted to a voice a +1, I'm excited for this feature and looking forward to using it for a client project :)

mattgperry:
Hey @cjoecker thanks for your work on this. I'm going to merge it into a branch now and have a play, check tests etc. Hopefully we'll finally have this out soon!

joi-lightyears:
Hi, have you guys merged this?
Because in newer framermotion version I still not use reorder on both axis

Theghost256 (commented):
Thank you for your help ☺️🙏

Theghost256 (commented):
Thank you for letting me help

Theghost256 (commented):
Thank you for letting me help

Theghost256 (commented):
Thank you for letting me help 🙂

PR #1689, opened 2022-09-07
Add isMotionComponent and unwrapMotionComponent

k15a:
This will add two new utility functions to framer-motion:

**1. isMotionComponent**

Checks if a React component is a motion component.

```jsx
isMotionComponent(motion.div) → true
isMotionComponent(motion(CustomComp)) → true
isMotionComponent("div") → false
isMotionComponent(CustomComp) → false
```

**2. unwrapMotionComponent**

Returns the wrapped component from a motion component or undefined if it's not a motion component.

```jsx
unwrapMotionComponent(motion.div) → "div"
unwrapMotionComponent(motion(CustomComp)) → CustomComp
unwrapMotionComponent("div") → undefined
unwrapMotionComponent(CustomComp) → undefined
```

### DIFF
--- a/packages/framer-motion/src/motion/index.tsx
+++ b/packages/framer-motion/src/motion/index.tsx
@@ -18,6 +18,8 @@ import { VisualElementHandler } from "./utils/VisualElementHandler"
 import { LazyContext } from "../context/LazyContext"
 import { SwitchLayoutGroupContext } from "../context/SwitchLayoutGroupContext"
 
+const wrappedComponentSymbol = Symbol.for("wrappedMotionComponent")
+
 export interface MotionComponentConfig<Instance, RenderState> {
     preloadedFeatures?: FeatureBundle
     createVisualElement?: CreateVisualElement<Instance>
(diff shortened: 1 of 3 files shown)


packages/framer-motion/src/motion/index.tsx
```diff
@@ -143,7 +145,43 @@ export function createMotionComponent<Props extends {}, Instance, RenderState>({
         )
     }
 
-    return forwardRef(MotionComponent)
+    const ForwardRefComponent = forwardRef(MotionComponent)
+
+    Object.defineProperty(ForwardRefComponent, wrappedComponentSymbol, {
```
k15a, line 150: Not sure if we should do this instead:
```
ForwardRefComponent[wrappedComponentSymbol] = Component
```

k15a: Looks like `enumerable: false` is not doing anything for symbols. Not sure if we need to make this readonly. Setting the property directly would make this code more readable.

mattgperry: Yeah I'd just set it directly

mattgperry (approved):

mattgperry:
LGTM, if you just switch to direct assignment I'll merge and publish

k15a:
@mattgperry I've switched it.

PR #1706, opened 2022-09-18
[fix] declared variables at the top of their scope

nurcinozer:
It's good practice to declare variables at the top of their scope.


mattgperry:
Thanks for the PR but this is a stylistic choice that I disagree with. Hoisting serves no purpose with `let/const` declarations and removing variable initialisation from its context makes code harder to read.

PR #1712, opened 2022-09-22
Adding `animationValues` prop

mattgperry:
`animationValues` allows motion values to be passed into a component. These will be used for all animations but won't be rendered. This allows animations to be composed outside the component, then passed back in via `style` for rendering.

```jsx
            const { scrollY } = useScroll()
            const parallaxY = useTransform(scrollY, transform)
            const animatedY = useMotionValue(0)
            const y = useTransform([parallaxY, animatedY], ([a, b]: [number, number]) => a + b)
            
            return (
              <motion.div animationValues={{ y: animatedY }} style={{ y }} />
            )
```

Some other use-cases ideas:

Here we could derive `x` and `y` from `distance` and `angle` motion values which dont get rendered, only animated.

```
<motion.div animate={{ distance: 100, angle: 360 }} animationValues={{ distance, angle }} style={{ x, y }} />
```

Here we could derive `transform` from `x` via `useMotionTemplate`, replacing `transformTemplate` prop

```
<motion.div animate={{ x: 100 }} animationValues={{ x }} style={{ transform }} />
```

TODO: 

- [ ] Name
- [ ] Ensure that when a value is defined only in `animationValues` it doesn't get rendered on the element unless a custom `style` is defined


packages/framer-motion/src/motion/types.ts
```diff
@@ -265,6 +265,10 @@ export interface MotionAdvancedProps {
     inherit?: boolean
 }
 
+type ExternalMotionValues = {
+    [K in keyof MotionStyle]: MotionValue<number> | MotionValue<string>
```
nvh, line 269: Can we be a bit smarter about the types here and exclude the non-MotionValue type from `MotionStyle[K]`?

mattgperry: What do you mean by this? Regardless I also had to change this to `[key: string]` as more of the interesting use-cases involve values not included in `MotionStyle`

nvh: Ah right, so then I guess it doesn't really matter anymore.

PR #1719, opened 2022-09-30
Cancel scheduled frame when unmounting useTransform

mattgperry:
This PR ensures that when` useTransform`'s subscription effect is cleaned up it cancels any scheduled animation frame.

 I haven't been able to capture this in a test but have verified it fixes the corresponding bug in our Slideshow component.

### DIFF
--- a/dev/package.json
+++ b/dev/package.json
@@ -1,6 +1,6 @@
 {
     "name": "framer-motion--dev",
-    "version": "7.5.0",
+    "version": "7.5.1-beta.0",
     "private": true,
     "scripts": {
         "dev": "webpack serve --config ./webpack/config.js --hot"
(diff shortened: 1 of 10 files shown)


dev/package.json
```diff
@@ -1,15 +1,15 @@
...
     "scripts": {
         "dev": "webpack serve --config ./webpack/config.js --hot"
     },
     "dependencies": {
         "@react-three/drei": "^7.27.3",
         "@react-three/fiber": "^8.2.2",
-        "framer-motion": "^7.5.0",
-        "framer-motion-3d": "^7.5.0",
+        "framer-motion": "^7.5.1-beta.0",
+        "framer-motion-3d": "^7.5.1-beta.0",
```
nvh, line 12: should we include beta versions here?

packages/framer-motion/src/value/use-on-change.ts
```diff
@@ -11,9 +11,17 @@ export function useOnChange<T>(
     }, [callback])
 }
 
-export function useMultiOnChange(values: MotionValue[], handler: () => void) {
+export function useMultiOnChange(
+    values: MotionValue[],
+    handler: () => void,
+    clean: () => void
```
nvh, line 17:
```suggestion
    cleanup: () => void
```

nvh (approved):
There's an empty test file added, we should remove that.

dev/package.json
mattgperry: this is just cruft really, it'll get updated to 7.5.1 when I publish from main

packages/framer-motion/src/value/use-on-change.ts
nvh: Shouldn't we also make this optional?

PR #1729, opened 2022-10-06
Fix behaviour for values removed from `animate`

mattgperry:
This PR fixes the behaviour of animating values when removed from `animate`.

Previously, if a value was originally animated from `initial`:
- If the value was removed from `animate`, it would animate back to the **originally read value**.
- If the value was removed from both `initial` and `animate`, it would animate back to the **originally read value**.

Now, if a value is originally animated from `initial`:
- If the value is removed from `animate`, it animates to the **current value** in `initial`.
- If the value is removed from both `initial` and `animate`, it **doesn't animate**.

### DIFF
--- a/packages/framer-motion/src/render/index.ts
+++ b/packages/framer-motion/src/render/index.ts
@@ -103,7 +103,7 @@ export const visualElement =
         }
 
         /**
-         * Create an object of the values we initially from (if initial prop present).
+         * Create an object of the values we initially animated from (if initial prop present).
          */
         const initialValues = props.initial ? { ...latestValues } : {}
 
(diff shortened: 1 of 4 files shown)


packages/framer-motion/src/render/index.ts
```diff
@@ -101,6 +102,11 @@ export const visualElement =
             ...latestValues,
         }
 
+        /**
+         * Create an object of the values we initially from (if initial prop present).
```
hemlok, line 106:
```suggestion
         * Create an object of the values we initially animated from (if initial prop present).
```
?

hemlok (approved):
Nice!

PR #1738, opened 2022-10-17
Adding frame-locked delay function

mattgperry:
This is a framerate-locked `delay` function that can avoid the use of `setTimeout`.

### DIFF
--- a/packages/framer-motion/src/animation/utils/transitions.ts
+++ b/packages/framer-motion/src/animation/utils/transitions.ts
@@ -13,6 +13,7 @@ import { warning } from "hey-listen"
 import { getAnimatableNone } from "../../render/dom/value-types/animatable-none"
 import { instantAnimationState } from "../../utils/use-instant-transition-state"
 import { resolveFinalValueInKeyframes } from "../../utils/resolve-value"
+import { delay } from "../../utils/delay"
 
 type StopAnimation = { stop: () => void }
 
@@ -23,7 +24,7 @@ type StopAnimation = { stop: () => void }
  */
 export function isTransitionDefined({
     when,
-    delay,
+    delay: _delay,
     delayChildren,
     staggerChildren,
     staggerDirection,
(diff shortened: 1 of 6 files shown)


packages/framer-motion/src/animation/utils/transitions.ts
```diff
@@ -13,6 +13,7 @@ import { warning } from "hey-listen"
 import { getAnimatableNone } from "../../render/dom/value-types/animatable-none"
 import { instantAnimationState } from "../../utils/use-instant-transition-state"
 import { resolveFinalValueInKeyframes } from "../../utils/resolve-value"
+import { delay } from "../../utils/delay"
```
nvh, line 16: It seems this import is unused?

packages/framer-motion/src/animation/utils/transitions.ts
```diff
@@ -277,18 +277,19 @@ export function startAnimation(
...
+        let cancelDelay: VoidFunction
+        if (delayBy) {
+            cancelDelay = delay(start, secondsToMilliseconds(delayBy))
         } else {
             start()
         }
 
         return () => {
-            clearTimeout(delayTimer)
+            cancelDelay && cancelDelay()
```
nvh, line 292: Would it be safe to set `cancelDelay` to undefined after calling it?

nvh (changes requested):
I think there's still some bugs, left comments

nvh:
I really like this idea, btw! Think its smart!

packages/framer-motion/src/animation/utils/transitions.ts
mattgperry: I don't think it's necessary, same as cancelTimeout etc nothing will happen if you keep calling it

packages/framer-motion/src/animation/utils/transitions.ts
mattgperry: On 286

nvh: 💯

nvh (approved):

PR #1748, opened 2022-11-01
Converting `VisualElement` to class

mattgperry:
Changing `VisualElement` from a factory function to a class reduces memory usage on framer.com by 25%, from 44mb to 33mb.


dev/projection/script-undo.js
```diff
@@ -83,7 +83,7 @@ Undo.createNode = (element, parent, options = {}, overrideId) => {
         scheduleRender()
     }
 
-    node.render = () => visualElement.syncRender()
+    node.render = () => visualElement.render()
```
nvh, line 86: What's the difference between `syncRender` and `render`?

mattgperry: I just renamed the method as there wasn't a public `render` method anyway and so the `sync` seemed a bit redundant, perhaps it was legacy

packages/framer-motion/src/gestures/drag/VisualElementDragControls.ts
```diff
@@ -78,7 +78,7 @@ export class VisualElementDragControls {
     private elastic = createBox()
 
     constructor(visualElement: VisualElement) {
-        this.visualElement = visualElement
+        this.visualElement = visualElement as any
```
nvh, line 81: Oof, any way we can avoid this cast?

nvh: maybe?
```suggestion
    constructor(visualElement: VisualElement<HTMLElement>) {
        this.visualElement = visualElement
```

packages/framer-motion/src/gestures/drag/VisualElementDragControls.ts
```diff
@@ -490,7 +490,7 @@ export class VisualElementDragControls {
          * Update the layout of this element and resolve the latest drag constraints
          */
         const { transformTemplate } = this.visualElement.getProps()
-        this.visualElement.getInstance().style.transform = transformTemplate
+        this.visualElement.current!.style.transform = transformTemplate
```
nvh, line 493: This also doesn't look good. There surely must be something we can do with the types here.

packages/framer-motion/src/gestures/drag/VisualElementDragControls.ts
mattgperry: Ah yeah, fixed this now

packages/framer-motion/src/render/types.ts
```diff
@@ -170,14 +37,109 @@ export type VisualElementOptions<Instance, RenderState = any> = {
...
+     */
+    onUnmount?(): void
+}
+
+export type EventProps = LayoutLifecycles & AnimationLifecycles
+
+export type CreateVisualElement<Instance> = (
+    Component: string | React.ComponentType<React.PropsWithChildren<unknown>>,
+    options: VisualElementOptions<Instance>
+) => VisualElement<Instance>
```
nvh, line 145: Why is this type needed? We shoud be able to use the constructor form VisualElement, right?

packages/framer-motion/src/render/types.ts
```diff
@@ -170,14 +37,109 @@ export type VisualElementOptions<Instance, RenderState = any> = {
...
+     */
+    onLayoutAnimationStart?(): void
+
+    /**
+     * @internal
+     */
+    onLayoutAnimationComplete?(): void
+}
+
+export interface AnimationLifecycles {
```
nvh, line 77: Not, sure but is lifecycle one word or two?
```suggestion
export interface AnimationLifeCycles {
```

packages/framer-motion/src/gestures/drag/VisualElementDragControls.ts
mattgperry: Yeah we can do a check here for `current` but we're making explicit here something that was actually theorhetically not true (that `getInstance()` would always return an element) even when it effectively was true

mattgperry: Ok updated with checks

packages/framer-motion/src/render/types.ts
mattgperry: This is a function that returns **a** visual element. For our DOM motion components this can be either `HTMLVisualElement` or `SVGVisualElement`

packages/framer-motion/src/gestures/drag/VisualElementDragControls.ts
nvh: I think if we think it should crash, we should make that explicit (with an assert or throwing an error) if it shouldn't crash, we should just do a check.

packages/framer-motion/src/render/types.ts
mattgperry: TIL it is two words

packages/framer-motion/src/motion/features/viewport/use-viewport.ts
```diff
@@ -93,7 +93,7 @@ function useIntersectionObserver(
         }
 
         return observeIntersection(
-            visualElement.getInstance(),
+            visualElement.current!,
```
nvh, line 96: We should add a check here too.

packages/framer-motion/src/render/types.ts
mattgperry: Ah but in terms of programming it's normally one word https://reactjs.org/docs/state-and-lifecycle.html

packages/framer-motion/src/motion/index.tsx
```diff
@@ -135,7 +135,7 @@ export function createMotionComponent<Props extends {}, Instance, RenderState>({
                             visualState,
                             context.visualElement,
                             externalRef
-                        ),
+                        ) as React.Ref<Instance>,
```
nvh, line 138: Can we do without this cast? Maybe by making `useMotionRef` generic?

packages/framer-motion/src/motion/utils/use-visual-element.ts
```diff
@@ -21,8 +22,9 @@ export function useVisualElement<Instance, RenderState>(
     const presenceContext = useContext(PresenceContext)
     const reducedMotionConfig = useContext(MotionConfigContext).reducedMotion
 
-    const visualElementRef: MutableRefObject<VisualElement | undefined> =
-        useRef(undefined)
+    const visualElementRef: MutableRefObject<
+        VisualElement<Instance> | undefined
+    > = useRef(undefined)
```
nvh, line 27: I think you can infer the type by doing 
```suggestion
    const visualElementRef = useRef<VisualElement<Instance>>(undefined)
```

packages/framer-motion/src/motion/features/viewport/use-viewport.ts
mattgperry: Added check

packages/framer-motion/src/render/VisualElement.ts
```diff
@@ -0,0 +1,810 @@
...
+const propEventHandlers = [
+    "AnimationStart",
+    "AnimationComplete",
+    "Update",
+    "Unmount",
+    "BeforeLayoutMeasure",
+    "LayoutMeasure",
+    "LayoutAnimationStart",
+    "LayoutAnimationComplete",
+]
```
nvh, line 52: Ah, we are. I think we can do
```suggestion
const propEventHandlers = [
    "AnimationStart",
    "AnimationComplete",
    "Update",
    "Unmount",
    "BeforeLayoutMeasure",
    "LayoutMeasure",
    "LayoutAnimationStart",
    "LayoutAnimationComplete",
] as const
```
here to improve the type

packages/framer-motion/src/motion/utils/use-visual-element.ts
mattgperry: Thanks, changed this

packages/framer-motion/src/render/VisualElement.ts
nvh: Or, event better:
```suggestion
const propEventHandlers: (keyof VisualElementEventCallbacks)[] = [
    "AnimationStart",
    "AnimationComplete",
    "Update",
    "Unmount",
    "BeforeLayoutMeasure",
    "LayoutMeasure",
    "LayoutAnimationStart",
    "LayoutAnimationComplete",
]
```

packages/framer-motion/src/render/VisualElement.ts
```diff
@@ -0,0 +1,810 @@
...
+        callback: VisualElementEventCallbacks[EventName]
+    ) {
+        if (!this.events[eventName]) {
+            this.events[eventName] = new SubscriptionManager()
+        }
+
+        return this.events[eventName].add(callback)
+    }
+
+    notify<EventName extends keyof VisualElementEventCallbacks>(
```
nvh, line 804: Does it need to be `extends`?
Don't we always want it to be one of the `VisualElementEventCallbacks`?

packages/framer-motion/src/render/VisualElement.ts
```diff
@@ -0,0 +1,810 @@
...
+    /**
+     * Create an object of the values we initially animated from (if initial prop present).
+     */
+    private initialValues: ResolvedValues
+
+    /**
+     * An object containing a SubscriptionManager for each active event.
+     */
+    private events: {
+        [key: string]: SubscriptionManager<any>
```
nvh, line 304: This could also be stronger typed.

packages/framer-motion/src/motion/index.tsx
mattgperry: Ah yeah I had to just pass a few more generics around

packages/framer-motion/src/render/VisualElement.ts
mattgperry: I've just added this, is this saying the array doesnt change?

packages/framer-motion/src/render/VisualElement.ts
```diff
@@ -0,0 +1,810 @@
...
+]
+
+/**
+ * A VisualElement is an imperative abstraction around UI elements such as
+ * HTMLElement, SVGElement, Three.Object3D etc.
+ */
+export abstract class VisualElement<
+    Instance = unknown,
+    RenderState = unknown,
+    Options extends {} = {}
```
nvh, line 61: I find it hard to view the differences from the factory function in this file, but I assume apart from syntax, not much changed?

packages/framer-motion/src/render/html/HTMLVisualElement.ts
```diff
@@ -11,87 +9,52 @@ import { scrapeMotionValuesFromProps } from "./utils/scrape-motion-values"
...
-                (isCSSVariable(key)
-                    ? computedStyle.getPropertyValue(key)
-                    : computedStyle[key]) || 0
-
-            return typeof value === "string" ? value.trim() : value
-        }
-    },
-
-    sortNodePosition(a, b) {
+export abstract class DOMVisualElement<
```
nvh, line 23: Can we put this in it's own file?

packages/framer-motion/src/render/types.ts
nvh: Can we make it specific to those visual elements then?
Don't feel strongly, because I don't have full context, so feel free to leave as is too.

packages/framer-motion/src/render/utils/__tests__/animation-state.test.ts
```diff
@@ -1,47 +1,59 @@
...
 import { checkTargetForNewValues, getOrigin } from "../setters"
-import { visualElement } from "../../"
 import { MotionProps } from "../../../motion/types"
 import { createHtmlRenderState } from "../../html/utils/create-render-state"
 import { createBox } from "../../../projection/geometry/models"
+import { VisualElement } from "../../VisualElement"
+import { TargetAndTransition } from "../../.."
 
-const stateVisualElement = visualElement<
+class StateVisualElement extends VisualElement<
```
nvh, line 11: Is this one only for testing? Should we move it to it's own file?

packages/framer-motion/src/render/VisualElement.ts
```diff
@@ -0,0 +1,813 @@
...
+        [key: string]: SubscriptionManager<any>
+    } = {}
+
+    /**
+     * An object containing an unsubscribe function for each prop event subscription.
+     * For example, every "Update" event can have multiple subscribers via
+     * VisualElement.on(), but only one of those can be defined via the onUpdate prop.
+     */
+    private propEventSubscriptions: {
+        [key: string]: VoidFunction
```
nvh, line 313: Not sure if possible, but maybe?
```suggestion
        [key: keyof typeof propEventHandlers]: VoidFunction
```

packages/framer-motion/src/render/VisualElement.ts
mattgperry: Yeah I'm not sure how else to type it though. By doing it like this I can send `EventName` through to `callback: VisualElementEventCallbacks[EventName]` and get the correct type for that.

packages/framer-motion/src/render/VisualElement.ts
mattgperry: Yeah most of it is the same really

packages/framer-motion/src/render/html/HTMLVisualElement.ts
mattgperry: Ah yeah I will do - this was actually messing with VSC's parsing of the second class too

packages/framer-motion/src/render/utils/__tests__/animation-state.test.ts
mattgperry: Done!

packages/framer-motion/src/render/VisualElement.ts
mattgperry: This gives me 

```
An index signature parameter type cannot be a literal type or generic type. Consider using a mapped object type instead.
```

packages/framer-motion/src/render/VisualElement.ts
nvh: `as const` makes sure it is not typed as `string[]`, but a tuple of these string values

packages/framer-motion/src/render/VisualElement.ts
nvh: I looked into improving this, but Typescript won't let me.

nvh (approved):

PR #1750, opened 2022-11-02
Removing support for mouse and touch events

mattgperry:
Pointer events are extremely [well supported](https://caniuse.com/pointer), this PR removes support for mouse and touch event fallbacks. Saves almost 0.3kb from the `motion` component.


packages/framer-motion/package.json
```diff
@@ -73,31 +73,31 @@
     "bundlesize": [
         {
             "path": "./dist/size-rollup-motion.js",
-            "maxSize": "29.40 kB"
+            "maxSize": "29.12 kB"
```
mattgperry, line 76: This one is the "real" saving - the others saw a similar decrease with the `class` PR and are likely from a previous PR.

packages/framer-motion-3d/jest.setup.tsx
```diff
@@ -1,6 +1,6 @@
 import "@testing-library/jest-dom"
 // Get fireEvent from the native testing library
-// because @testing-library/react one switches out mouseEnter and mouseLeave
+// because @testing-library/react one switches out pointerEnter and pointerLeave
```
nvh, line 3: Is this true for pointerEnter and pointerLeave too?

packages/framer-motion/jest.setup.tsx
```diff
@@ -1,6 +1,6 @@
 import "@testing-library/jest-dom"
 // Get fireEvent from the native testing library
-// because @testing-library/react one switches out mouseEnter and mouseLeave
+// because @testing-library/react one switches out pointerEnter and pointerLeave
```
nvh, line 3: Same here.

packages/framer-motion/jest.setup.tsx
```diff
@@ -18,21 +18,21 @@ export const click = (element: Element) =>
...
+export const pointerDown = (element: Element) =>
     act(() => {
-        fireEvent.mouseDown(element)
+        fireEvent.pointerDown(element)
     })
-export const mouseUp = (element: Element) =>
+export const pointerUp = (element: Element) =>
     act(() => {
-        fireEvent.mouseUp(element)
+        fireEvent.pointerUp(element)
```
nvh, line 35: I assume these are now unused and we can just remove this whole file.

packages/framer-motion/src/gestures/PanSession.ts
```diff
@@ -141,12 +138,12 @@ export class PanSession {
...
 
     constructor(
-        event: AnyPointerEvent,
+        event: PointerEvent,
         handlers: Partial<PanSessionHandlers>,
         { transformPagePoint }: PanSessionOptions = {}
     ) {
         // If we have more than one touch, don't start detecting this gesture
-        if (isTouchEvent(event) && event.touches.length > 1) return
+        if (!event.isPrimary) return
```
nvh, line 146: Can't we just remove this completely

packages/framer-motion/src/gestures/PanSession.ts
```diff
@@ -198,12 +195,12 @@ export class PanSession {
...
     }
 
-    private handlePointerMove = (event: AnyPointerEvent, info: EventInfo) => {
+    private handlePointerMove = (event: PointerEvent, info: EventInfo) => {
         this.lastMoveEvent = event
         this.lastMoveEventInfo = transformPoint(info, this.transformPagePoint)
 
         // Because Safari doesn't trigger mouseup events when it's above a `<select>`
-        if (isMouseEvent(event) && event.buttons === 0) {
+        if (event.isPrimary) {
```
nvh, line 203: This comment is probably no longer true, and I wonder if we should keep the logic

nvh (approved):
I like the logic cleanup, but think that for the size saving it's almost not worth it.
Don't feel strongly though.

PR #1751, opened 2022-11-03
Fixing jumping boxes when switching relative projection parent

mattgperry:
There was code that could potentially reassign a relative projection parent without cleaning up outdated relative projection targets.

Contributes towards https://github.com/framer/company/issues/26070

### DIFF
--- a/dev/projection/animate-relative-skip-parent.html
+++ b/dev/projection/animate-relative-skip-parent.html
@@ -1,0 +1,107 @@
+<html>
+    <head>
+        <style>
+            body {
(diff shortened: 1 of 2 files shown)


nvh (approved):
Shouldn't we create a test for this?

mattgperry:
@nvh That's what [dev/projection/animate-relative-skip-parent.html](https://github.com/framer/motion/pull/1751/files#diff-be2467791f201fa9a829e363e43a96fd80d79cfa06e3e5860db318e6efa0760d) is

PR #1754, opened 2022-11-08
Tests for `position: sticky` layout animations

mattgperry:
Failing tests are currently skipped via `.skip`


nvh:
What's the plan for enabling them?

nvh (approved):
Why are these files in the `dev` folder?

mattgperry:
> Why are these files in the dev folder?

Because they leverage the webpack dev server 

> What's the plan for enabling them?

A series of small refactors getting the codebase closer to what I have in mind for a fix without breaking current tests, then a final PR that enables them and makes the final fixes

PR #1757, opened 2022-11-10
Consolidate pre/post-render layout measurements

mattgperry:
This PR unifies snapshot and layout into a unified data type


nvh (approved):
Look good!

PR #1761, opened 2022-11-10
Adding `frameTimestamp` to measurements

mattgperry:
This PR adds a `frameTimestamp` field to `Measurements`.

This will enable us to more reliably avoid double measurements within a given animation frame and more reliably understand the freshness of data when attempting to resolve relative bounding boxes.

### DIFF
--- a/packages/framer-motion/src/projection/node/create-projection-node.ts
+++ b/packages/framer-motion/src/projection/node/create-projection-node.ts
@@ -1,4 +1,4 @@
-import sync, { cancelSync, flushSync, Process } from "framesync"
+import sync, { cancelSync, flushSync, getFrameData, Process } from "framesync"
 import { mix } from "popmotion"
 import {
     animate,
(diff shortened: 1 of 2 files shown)


packages/framer-motion/src/projection/node/create-projection-node.ts
```diff
@@ -55,6 +55,11 @@ const transformAxes = ["", "X", "Y", "Z"]
  */
 const animationTarget = 1000
 
+/**
+ * A mutable state containing the latest animation frame timestamp.
+ */
+const frameData = getFrameData()
```
nvh, line 61: Why is this called data? Going at the comment I would say `frameState`

nvh (approved):

packages/framer-motion/src/projection/node/create-projection-node.ts
mattgperry: Yeah this can also work

PR #1762, opened 2022-11-10
Making `resetRotation` 2x faster when no rotations applied

mattgperry:
Before a layout animation we check the tree for any rotations and suspend them so we can get an accurate read from `getBoundingClientRect`. It's not ideal as it introduces an extra (batched) DOM write, but otherwise the bounding box is unsalvageable. 

99% of the time there will be no rotations. By unrolling the initial check we speed this function up to 2x.

### DIFF
--- a/packages/framer-motion/src/projection/node/create-projection-node.ts
+++ b/packages/framer-motion/src/projection/node/create-projection-node.ts
@@ -773,7 +773,7 @@ export function createProjectionNode<I>({
 
             const box = visualElement.measureViewportBox()
 
-            // // Remove viewport scroll to give page-relative coordinates
+            // Remove viewport scroll to give page-relative coordinates
             const { scroll } = this.root
             if (scroll) {
                 translateAxis(box.x, scroll.x)
--- a/packages/framer-motion/src/projection/node/create-projection-node.ts
+++ b/packages/framer-motion/src/projection/node/create-projection-node.ts
@@ -1438,8 +1438,7 @@ export function createProjectionNode<I>({
 
             // Check the rotate value of all axes and reset to 0
             for (let i = 0; i < transformAxes.length; i++) {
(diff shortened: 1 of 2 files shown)


packages/framer-motion/src/projection/node/create-projection-node.ts
```diff
@@ -1417,30 +1417,34 @@ export function createProjectionNode<I>({
...
-                    continue
-                }
-
-                hasRotate = true
+                const key = "rotate" + transformAxes[i]
 
                 // Record the rotation and then temporarily set it to 0
-                resetValues[key] = visualElement.getStaticValue(key)
+                resetValues[key] = latestValues[key]
                 visualElement.setStaticValue(key, 0)
```
nvh, line 1445: Should we check for `latestValues[key]` to be `0` here too?
Or wouldn't that lead to much preformance benifit?

nvh (approved):

packages/framer-motion/src/projection/node/create-projection-node.ts
mattgperry: Ah yeah it's true we don't need to do anything for those values

packages/framer-motion/src/projection/node/create-projection-node.ts
```diff
@@ -1417,30 +1417,34 @@ export function createProjectionNode<I>({
             // If there's no detected rotation values, we can early return without a forced render.
             let hasRotate = false
 
-            // Keep a record of all the values we've reset
+            /**
+             * An unrolled check for rotation values. Most elements don't have any rotation and
+             * skipping the nested loop and new object creation is 50% faster.
+             */
+            const { latestValues } = visualElement
```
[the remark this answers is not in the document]
mattgperry: Yeah

PR #1773, opened 2022-11-14
Adding `layoutRoot` prop

mattgperry:
Adds `layoutRoot` prop for instance layout animations relative to this element.


TmLev:
Any timeline on this?

mattgperry:
No not currently. Fixing some of these use-cases is quite difficult and aren’t currently surfacing as bugs within Framer itself.

PR #1781, opened 2022-11-18
Fixed layout animations without `position: sticky` (hack edition)

mattgperry:
Fixes https://github.com/framer/motion/issues/1535

This PR adds a `layoutRoot` prop. When applied to an element, it gets forced to perform an instant layout animation any time a child performs one. Then, any children can hook into its final layout to resolve relative layout animations.

A more universal fix for this class of animations would be preferable but this will do for a quick fix.


TmLev:
I have a sticky sidebar with navigation items which have `layout` prop for smooth horizontal motion when expanding/collapsing the sidebar. If user scrolls down on any page and navigates to a different page (Next.js app), navigation items "slide up" from the bottom to theirs supposed location.

Interested in trying out this PR with fix, but can't figure out how to correctly depend on this branch with Yarn. Tried `yarn add`-ing by branch/commit (got [build error](https://gist.github.com/TmLev/8e3faa20174d6e1ec673c3dd8c7d78f1)), also tried to `git clone` locally and specifying path to the package (also didn't quite work).

TrillCyborg:
@mattgperry any idea when this will get merged into `main`? I have a few fixed components that are totally incompatible with this lib right now. I'm not sure how to move forward since I love this lib but can't use it on most pages

mattgperry:
It is merged?

TrillCyborg:
@mattgperry is it in a release on NPM? Searching the main branch for `layoutRoot` gives no results

TmLev:
> It is merged?

It was merged into `fix/sticky-11`, not main branch

PR #1784, opened 2022-11-21
Skip projection calculations for unchanged subtrees

mattgperry:
This PR skips projection recalculations for nodes where neither they nor their ancestors have updated projection target or transforms. In the included stress test this is a **per-frame JS reduction of 90%** - this will scale proportionally with the amount of dirty vs clean subtrees.


hemlok (approved):
👍

PR #1787, opened 2022-11-24
Adding isTransformDirty optimisation

mattgperry:
This PR specifies between when a projection is dirty and when a transform is dirty. Ancestor transforms only matter for shared element transitions.

In the included stress test this leads to a **75% per-frame JS reduction**. As with previous optimisations, in the real world there is no lower or upper bound to this cost saving, it depends on the tree.


hemlok (approved):
👍

PR #1788, opened 2022-11-24
Treat `<motion.svg />` components as HTML

mattgperry:
For our purposes, `svg` elements are the same as HTML elements, rather than SVG elements.


hemlok (approved):
👏 🙇

PR #1790, opened 2022-11-24
Setting scale threshold

mattgperry:
This PR adds a threshold for rendered scales. In deep trees when many elements are effectively `scale: 1` this might get calculated (due to floating point inaccuracy) as `1.0000000001` etc. These actually get rendered as `scale(1)` when set in the browser so they're truly useless.

By not rendering these scales we achieve a 25% higher framerate in the included stress test.


hemlok (approved):
:heavy-breathing:

PR #1794, opened 2022-11-28
Use cached scroll if possible

mattgperry:
Reading scroll can trigger style recalculations. Here we use the cached scroll if possible. It might be possible to do this post-render too but some testing needs doing to see if the scroll event will fire before the useLayoutEffect (unlikely)


mattgperry:
Closing as the included test proves it isn't possible to reliably break the cache when dealing with synchronous code.

PR #1802, opened 2022-12-01
Removing `popmotion` as a dependency

mattgperry:
Moving the bits of Popmotion we use into Framer Motion to make way for future performance improvements like
- Replacing string values with data models
- animations -> classes
- Blending in WAAPI animations where possible

In the main, stuff has been brought over without changes, with the exception of `interpolate` and `cubicBezier` which have had size reductions based on my latest version in Motion One


hemlok:
Will we need a follow-up `FramerStudio` pr to change all `popmotion` imports to `framer-motion` to avoid file-size increase there?

hemlok (approved):

mattgperry:
@hemlok Yeah it'll be modest savings but worth doing

PR #1803, opened 2022-12-02
Remove `style-value-types` and `framesync` as dependencies

mattgperry:
This PR moves `style-value-types` and `framesync` inside of Framer Motion to ease the move to maintaining values as data structures vs strings.


packages/framer-motion/src/frameloop/data.ts
```diff
@@ -0,0 +1,4 @@
+export const frameData = {
```
nvh, line 1: Is this just a public const?

mattgperry: Yeah - the props are mutable so we don't write a new object every frame, it's actually always been this way and I think its mutable nature was masked slightly by the `getFrameData()` api

packages/framer-motion/src/utils/mix-color.ts
```diff
@@ -26,10 +29,10 @@ function asRGBA(color: Color | string) {
     let model = type!.parse(color)
 
     if (type === hsla) {
-        model = hslaToRgba(model)
+        model = hslaToRgba(model as HSLA)
```
nvh, line 32: Why is this cast now needed?

packages/framer-motion/src/value/types/numbers/index.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export const alpha = {
+    ...number,
+    transform: (v: number) => clamp(0, 1, v),
+}
+
+export const scale = {
+    ...number,
+    default: 1,
+}
```
nvh, line 17: It seems to me these don't really belong a `numbers` utils

packages/framer-motion/src/value/types/utils.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export const floatRegex = /(-)?([\d]*\.?[\d])+/g
+export const colorRegex =
+    /(#[0-9a-f]{6}|#[0-9a-f]{3}|#(?:[0-9a-f]{2}){2,4}|(rgb|hsl)a?\((-?[\d\.]+%?[,\s]+){2}(-?[\d\.]+%?)\s*[\,\/]?\s*[\d\.]*%?\))/gi
+export const singleColorRegex =
+    /^(#[0-9a-f]{3}|#(?:[0-9a-f]{2}){2,4}|(rgb|hsl)a?\((-?[\d\.]+%?[,\s]+){2}(-?[\d\.]+%?)\s*[\,\/]?\s*[\d\.]*%?\))$/i
+
+export function isString(v: any): v is string {
+    return typeof v === "string"
+}
```
nvh, line 13: These are not really "type" utils, right? Can we come up with a better name?

nvh (changes requested):

packages/framer-motion/src/utils/mix-color.ts
mattgperry: There's some extra casts in there because I believe Framer Motion is compiled with stricter settings than Popmotion was

packages/framer-motion/src/value/types/utils.ts
mattgperry: This folder is basically a drag-drop of `style-value-types`. The idea in the medium term is to give motion values "typed" data models - but not TypeScript types, more like https://developer.mozilla.org/en-US/docs/Web/API/CSS_Typed_OM_API. vs keeping values in strings as we currently do.

Maybe "model" or something would be a better name?

packages/framer-motion/src/value/types/numbers/index.ts
mattgperry: The idea is they're flavours of `number` like `int` would be. It's sibling to unit types which contains `"px"` `"%"` etc

packages/framer-motion/src/utils/mix-color.ts
nvh: Makes sense, can we add a todo to try to remove them?

packages/framer-motion/src/value/types/numbers/index.ts
nvh: 👌🏻

packages/framer-motion/src/value/types/utils.ts
nvh: Although it is pretty generic, I kind of like `model` 
I would also be ok with a comment explaining this on top of this file.

nvh (approved):

Issue #1807, opened 2022-12-03
[BUG] svg viewBox property no longer animates after 7.6.11


willhindson:
**2. Describe the bug**
Animating the svg viewBox property no longer works after framer motion 7.6.11 (release notes: "Treat <motion.svg /> components as HTML")

**3. IMPORTANT: Provide a CodeSandbox reproduction of the bug**
https://codesandbox.io/s/fervent-monad-s4eb9e?file=/src/App.js
Note — the svg is invisible, but if you click anywhere on the yellow and inspect you will see that the viewBox property on the svg animates, with framer motion 7.6.10 as a dependency (as it is in this codesandbox), if you change it to 7.6.11 or later, it will not animate.

**4. Steps to reproduce**
1. Create an array of viewbox properties within the animate prop on a motion.svg element, i.e.
```
animate={{
        viewBox: [
          '0 0 360 100',
          '0 0 360 100',
          '0 -50 360 100',
          '0 -50 360 100',
          '-50 0 360 100',
          '-50 0 360 100',
          '-80 -60 360 100',
          '-100 -60 360 100',
        ],
      }}
```
2. Check in the inspector to see it is working correctly, as it should on 7.6.10 and below
3. Use any version of motion from 7.6.11 to 7.6.18 and the viewBox property will not animate

**5. Expected behavior**
The viewBox prop should animate between items in the array

**7. Environment details**
Mac os ventura 13.0.1, MS Edge 107.0.1418.68

PR #1809, opened 2022-12-05
Fixing SVG viewBox animations

mattgperry:
Fixes https://github.com/framer/motion/issues/1807

### DIFF
--- a/packages/framer-motion/src/render/svg/SVGVisualElement.ts
+++ b/packages/framer-motion/src/render/svg/SVGVisualElement.ts
@@ -22,6 +22,8 @@ export class SVGVisualElement extends DOMVisualElement<
 > {
     type: "svg"
 
+    isSVGTag = false
+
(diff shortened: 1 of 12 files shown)


packages/framer-motion/src/render/svg/SVGVisualElement.ts
```diff
@@ -67,4 +70,9 @@ export class SVGVisualElement extends DOMVisualElement<
     ): void {
         renderSVG(instance, renderState, styleProp, projection)
     }
+
+    mount(instance: SVGElement) {
+        this.isSVGTag = instance.tagName.toLowerCase() === "svg"
```
hemlok, line 75: I'm on the fence about this suggestion but should we do:

`isSVGTag(tag: string): boolean`? We do use it in 3 places 🤔

hemlok (approved):

PR #1816, opened 2022-12-08
Optimised appear animations

mattgperry:
This PR adds the ability to play optimised appear animations via WAAPI and for Framer Motion to seamlessly take them over once React has loaded.

Animation data is shared via `window` and matching ids, so even if there's a hydration mismatch, the animation will continue unabated.


packages/framer-motion/rollup.config.js
```diff
@@ -56,6 +56,10 @@ const projection = Object.assign({}, config, {
         format: "umd",
         name: "Projection",
         exports: "named",
+        globals: {
```
mattgperry, line 59: This just removes a warning in the build script

packages/framer-motion/src/animation/legacy-popmotion/index.ts
```diff
@@ -124,7 +127,7 @@ export function animate<V = number>({
             isComplete = isForwardPlayback ? state.done : elapsed <= 0
         }
 
-        onUpdate?.(latest)
+        onUpdate && onUpdate(latest)
```
mattgperry, line 130: A slight aside to this PR but replacing `?.` for `foo &&` makes a surprising saving for transpilation.

packages/framer-motion/src/animation/legacy-popmotion/spring.ts
```diff
@@ -109,35 +111,6 @@ export function spring({
                             initialDelta * Math.cos(angularFreq * t))
                 )
             }
```
mattgperry, line 113: Here we're replacing the use of Motion One's `spring` function, which only supports under- and critically-dampened springs, with Framer Motion's which supports over-damped springs also.

To help offset the added bundlesize, we're replacing the closed-form velocity calculation with a second sample of the spring calculation.

packages/framer-motion/src/animation/legacy-popmotion/index.ts
nvh: 🤔 Why?

packages/framer-motion/src/animation/optimized-appear/types.ts
```diff
@@ -0,0 +1,7 @@
+export type MotionAppearAnimations = Map<string, Animation>
+
+declare global {
+    interface Window {
+        MotionAppearAnimations?: MotionAppearAnimations
+    }
+}
```
nvh, line 7: Why do we need to store these on window?

nvh (changes requested):
Think we should add a comment to the reason for the window globals.

packages/framer-motion/src/animation/legacy-popmotion/index.ts
mattgperry: It's just super-exact. `onUpdate?.()` becomes:

```
var _onUpdate;
(_onUpdate = onUpdate) === null || _onUpdate === void 0 ? void 0 : _onUpdate();
```

`??` is also upsettingly large. `const a = 0 ?? 1`:

```javascript
const a = (_ = 0) !== null && _ !== void 0 ? _ : 1;
```

packages/framer-motion/src/animation/optimized-appear/handoff.ts
```diff
@@ -0,0 +1,37 @@
...
+         *   1. So it continues to play until Framer Motion is ready to render
+         *      (avoiding a potential flash of the element's original state)
+         *   2. As all independent transforms share a single transform animation, stopping
+         *      it synchronously would prevent subsequent transforms from handing off.
+         */
+        sync.render(() => {
+            try {
+                animation.cancel()
+                MotionAppearAnimations.delete(animationId)
+            } catch (e) {}
```
shuangq, line 30: Should we add some logging for the errors?

packages/framer-motion/src/animation/optimized-appear/handoff.ts
```diff
@@ -0,0 +1,37 @@
...
+    const animation =
+        MotionAppearAnimations && MotionAppearAnimations.get(animationId)
+
+    if (animation) {
+        /**
+         * We allow the animation to persist until the next frame:
+         *   1. So it continues to play until Framer Motion is ready to render
+         *      (avoiding a potential flash of the element's original state)
+         *   2. As all independent transforms share a single transform animation, stopping
+         *      it synchronously would prevent subsequent transforms from handing off.
```
shuangq, line 24: I didn't quite understand this comment, could you give an example?

shuangq (approved):
Code looks good 👏

packages/framer-motion/src/animation/utils/transitions.ts
```diff
@@ -179,41 +182,52 @@ function getAnimation(
...
         !isTargetAnimatable ||
         valueTransition.type === false
-        ? set
-        : start
+
+    return useInstantAnimation
+        ? valueTransition.elapsed
+            ? () => delay(set, -valueTransition.elapsed)
+            : set()
+        : start()
```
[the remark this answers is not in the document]
mattgperry: This happens in https://github.com/framer/motion/pull/1818

packages/framer-motion/src/animation/optimized-appear/handoff.ts
mattgperry: I'll add a comment, this is expected behaviour. `Animation.cancel()` annoyingly throws, so we have to catch it here.

packages/framer-motion/src/animation/optimized-appear/handoff.ts
mattgperry: in psuedo code

```
animate("scale", 2)
 -> handoff from "transform" animation
 -> cancel "transform" animation
animate("x", 2)
 -> no "transform" animation present, start new animation
```

VS

```
animate("scale", 2)
 -> handoff from "transform" animation
 -> schedule cancel "transform" animation
animate("x", 2)
 -> handoff from "transform" animation
 -> schedule cancel "transform" animation
cancel "transform" animation
```

PR #1818, opened 2022-12-09
Refactoring internal animation functions

mattgperry:
This PR refactors the `startAnimation` function. Previously this served as a bridge between Framer Motion and Popmotion. Now these animations are inside Framer Motion we can more directly use the same options that we present to the user.

Likewise, the new function is more straightforward to follow and adds some logic to determine when to fire WAAPI animations. Note that in this PR the WAAPI animation is not integrated, nor does any value get marked as appropriate for WAAPI animations.


packages/framer-motion/src/animation/create-accelerated-animation.ts
```diff
@@ -0,0 +1,3 @@
+export function createAcceleratedAnimation() {
```
mattgperry, line 1: Stub for now, will not be called

packages/framer-motion/src/animation/legacy-popmotion/decay.ts
```diff
@@ -1,29 +1,32 @@
-import { Animation, AnimationState, DecayOptions } from "./types"
+import { DecayOptions } from "../types"
+import { Animation, AnimationState } from "./types"
 
 export function decay({
+    keyframes = [0],
```
nvh, line 5: This looks a bit weird, so I think we should
- Verify an array that contains a zero is actually what you mean here
- Add a comment why this is.

mattgperry: I've added a comment explaining this now, to expand on this it's less wrangling to coerce the different formats of `to: V[]` and `to: V, from: V` into a keyframes array as early as possible as this is used by both WAAPI and the majority of our JS animations. It looks a bit funny for `decay` but doing it this way keeps the interface compatible with all the other animations.

packages/framer-motion/src/animation/legacy-popmotion/inertia.ts
```diff
@@ -62,9 +61,9 @@ export function inertia({
         })
     }
 
-    if (isOutOfBounds(from)) {
+    if (isOutOfBounds(origin)) {
         // Start the animation with spring if outside the defined boundaries
-        startSpring({ from, velocity, to: boundaryNearest(from) })
+        startSpring({ velocity, keyframes: [origin, boundaryNearest(origin)] })
```
nvh, line 66: What does `boundaryNearest` mean?

packages/framer-motion/src/motion/__tests__/variant.test.tsx
```diff
@@ -702,7 +702,7 @@ describe("animate prop as variant", () => {
                 )
             }
         )
-
+        console.log(recordedOrder)
```
nvh, line 705: Stray console.log

packages/framer-motion/src/value/index.ts
```diff
@@ -34,6 +35,12 @@ export class MotionValue<V = any> {
      */
     version = "__VERSION__"
 
+    /**
+     * If a MotionValue has an owner, it was created internally within Framer Motion
+     * and therefore has no external listeners. It is therefore safe to animate via WAAPI.
+     */
+    owner?: VisualElement
```
nvh, line 42: Shouldn't we block or error adding external listeners?

nvh (approved):
I can't say I fully understand all the details, but the broad picture looks right to me.

There's one console.log in a test that should be removed.

packages/framer-motion/src/animation/legacy-popmotion/index.ts
```diff
@@ -62,28 +63,32 @@ export function animate<V = number>({
...
-        to = 100 as any
+        keyframes = [0, 100] as any
     }
 
-    const animation = animator({ ...options, from, to } as any)
+    const animation = animator({
+        ...options,
+        duration,
+        keyframes,
+    } as any)
```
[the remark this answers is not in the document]
mattgperry: I think it's probably possible if the types were revisited, basically any animation that `needsInterpolation` also only takes numbers and that's what it complains about here. Although it's already not inferring this correctly within the `if` so I'm not sure if being more explicit would fix this.

packages/framer-motion/src/animation/legacy-popmotion/inertia.ts
mattgperry: Renamed to `findNearestBoundary` - this is ultimately where drag constraints get applied, that kind of animation

packages/framer-motion/src/value/index.ts
mattgperry: No, it's not a problem if there are external listeners, it just means we run animations as before

Issue #1806, opened 2022-12-02
[FEATURE] GPU animations


mattgperry:
Start introducing WAAPI animations for select values, perhaps starting with opacity

finkrer:
Hey @mattgperry, thanks for this change! Do I understand correctly that only opacity is animated using WAAPI in the latest version?

I've noticed that framer animations were noticeably slow in Safari, and found this update. In the new version, opacity animations work great! But transforms are still done using styles, right? Can't wait to use `layoutId` animations in my project, but last time I checked they were too slow for production unfortunately, at least in my case.

mattgperry:
Hi there - yeah it’s just opacity for now. I’ll be looking at other values but transform is one of the least likely as layout animations require per-frame javascript. It’s be interesting to see what your perf issues are in a sandbox though

PR #1823, opened 2022-12-12
Enabling WAAPI animations for `opacity`

mattgperry:
Implements https://github.com/framer/motion/issues/1806


packages/framer-motion/src/animation/legacy-popmotion/index.ts
```diff
@@ -157,12 +158,15 @@ export function animate<V = number>({
         driverControls.start()
     }
 
-    play()
+    autoplay && play()
```
nvh, line 161: Heh, this was removed in the previous PR.

packages/framer-motion/src/animation/legacy-popmotion/index.ts
```diff
@@ -157,12 +158,15 @@ export function animate<V = number>({
...
 
     return {
         stop: () => {
             onStop && onStop()
             driverControls.stop()
         },
+        sampleForT: (t: number) => {
+            return animation.next(Math.max(0, t)).value
+        },
```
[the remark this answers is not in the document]
nvh: Can we add a comment or change the name?

packages/framer-motion/src/animation/waapi/create-accelerated-animation.ts
```diff
@@ -0,0 +1,98 @@
+import { EasingDefinition } from "../../easing/types"
+import type { VisualElement } from "../../render/VisualElement"
+import type { MotionValue } from "../../value"
+import { animate } from "../legacy-popmotion"
+import { spring } from "../legacy-popmotion/spring"
+import { AnimationOptions } from "../types"
+import { animateStyle } from "./"
+
+const sampleDelta = 10 //ms
```
nvh, line 9: Is there any science in picking this constant? Let's add a comment

packages/framer-motion/src/animation/waapi/create-accelerated-animation.ts
```diff
@@ -0,0 +1,98 @@
...
+        if (currentTime) {
+            const sampleAnimation = animate(options)
+            value.setWithVelocity(
+                sampleAnimation.sampleForT(currentTime - sampleDelta),
+                sampleAnimation.sampleForT(currentTime),
+                sampleDelta
+            )
+        }
+
+        animation.cancel()
```
nvh, line 96: Is there no `onCancel` callback?

nvh (approved):
A couple of questions/remarks, but nothing serious.

packages/framer-motion/src/animation/legacy-popmotion/index.ts
mattgperry: It was sods law that I went from "hey we've never once used this in Motion" to "i don't need to actually play this animation"

packages/framer-motion/src/animation/legacy-popmotion/index.ts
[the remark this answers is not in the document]
mattgperry: Sure maybe I'll just call it `sample` - `t` is commonly used in the code for "time" and `p` for "progress"

packages/framer-motion/src/animation/waapi/create-accelerated-animation.ts
mattgperry: Good point - added

```
/**
 * 10ms is chosen here as it strikes a balance between smooth
 * results (more than one keyframe per frame at 60fps) and
 * keyframe quantity.
 */
```

packages/framer-motion/src/animation/waapi/create-accelerated-animation.ts
mattgperry: This function is called by the `MotionValue` when a new animation interrupts this one (I'll add a comment) so any cancel callbacks will be handled by that (as of the MotionValue event PR)

Issue #1824, opened 2022-12-12
[FEATURE] adding `onStart` and `onComplete` callbacks to `useMotionValue`


kuus:
**Is your feature request related to a problem? Please describe.**
I would like to know if there is any intention or anything against adding an `onStart` and `onComplete` callbacks to `useMotionValue`, the use case is simply to be able to use a callback (maybe from a parent component) when the animation starts/ends. Another usecase is to set some styles depending on the current state of the animation. I know all of this is possible with the `animate` property of the `<m.{component} />` but it would be good I think being able to achieve the same as `onAnimationStart` and `onAnimationComplete` props with the "render-less" approach allowed by `useMotionValue`.

**Describe the solution you'd like**
From an API point of view I imagine this usage:
```
const marginValue = useMotionValue(0, {
  onStart: () => someRef.current && someRef.current.style.overflow = "hidden",
  onComplete: () => someRef.current && someRef.current.style.overflow = "auto",
);
```

mattgperry:
This is implemented, great suggestion https://twitter.com/mattgperry/status/1603371303775379456

PR #1826, opened 2022-12-13
Adding animation events to motion value

mattgperry:
Implements https://github.com/framer/motion/issues/1824

This PR adds a new event subscription method to motion values, `.on()`, along with three new motion value events, `"animationStart"`, `"animationComplete"` and `"animationCancel"`.

This allows multiple event handlers to be attached to the same motion value.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@ Framer Motion adheres to [Semantic Versioning](http://semver.org/).
 
 Undocumented APIs should be considered internal and may change without warning.
 
+## [7.10.0] 2022-12-15
+
+### Added
+
+-   `.on()` event method to `MotionValue`.
+-   `"animationStart"`, `"animationComplete"`, and `"animationCancel"` events for `MotionValue`.
+
 ## [7.9.1] 2022-12-14
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,19 @@ Undocumented APIs should be considered internal and may change without warning.
 ### Added
 
 -   `.on()` event method to `MotionValue`.
--   `"animationStart"`, `"animationComplete"`, and `"animationCancel"` events for `MotionValue`.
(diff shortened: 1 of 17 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,13 @@ Framer Motion adheres to [Semantic Versioning](http://semver.org/).
 
 Undocumented APIs should be considered internal and may change without warning.
 
+## [7.8.0] 2022-12-13
+
+### Added
+
+-   `.on()` event method to `MotionValue`.
+-   `"animationStart"`, `"animationComplete"`, and `"animationCancel"` events for `MotionValue`.
```
nvh, line 12: What about `"change"`?

packages/framer-motion/src/value/__tests__/index.test.ts
```diff
@@ -0,0 +1,73 @@
+import { motionValue } from "../"
+import { animate } from "../../animation/animate"
+
+describe("motionValue", () => {
+    test("change event fires when value changes", () => {
+        const value = motionValue(0)
+        const callback = jest.fn()
+
+        value.on("change", callback)
```
nvh, line 9: Are we keeping the old API around?

packages/framer-motion/src/value/index.ts
```diff
@@ -171,28 +157,34 @@ export class MotionValue<V = any> {
      * @param subscriber - A function that receives the latest value.
      * @returns A function that, when called, will cancel this subscription.
      *
-     * @public
+     * @deprecated
      */
     onChange(subscription: Subscriber<V>): () => void {
```
nvh, line 162: I guess there's my answer 😁

nvh (approved):

packages/framer-motion/src/value/index.ts
mattgperry: Ha. I think it's so widely used by us I won't even throw in a `warning` yet. I'm gonna drop it in the docs and then revisit in 6 months or so, maybe add a warning then.

PR #1834, opened 2022-12-16
Pregenerating keyframes for unsupported easing functions

mattgperry:
WAAPI doesn't support custom JavaScript easing functions and has a shorter list of supported easings than Framer Motion.

In https://github.com/framer/motion/pull/1830 we added some approximations of unsupported easings via `cubic-bezier`. This PR extends support to all easing functions by pre-generating the animation keyframes.

In the future it'll be preferable to change this approach to generating a custom `linear()` easing similar to Motion One, but this doesn't have wide browser support yet.

### DIFF
--- a/packages/framer-motion/src/animation/waapi/easing.ts
+++ b/packages/framer-motion/src/animation/waapi/easing.ts
@@ -2,8 +2,8 @@ import { BezierDefinition, Easing, EasingDefinition } from "../../easing/types"
 
 export function isWaapiSupportedEasing(easing?: Easing | Easing[]) {
     return (
-        !easing ||
-        Array.isArray(easing) ||
+        !easing || // Default easing
+        Array.isArray(easing) || // Bezier curve
         (typeof easing === "string" && supportedWaapiEasing[easing])
     )
 }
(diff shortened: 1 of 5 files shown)


packages/framer-motion/src/animation/waapi/easing.ts
```diff
@@ -1,4 +1,12 @@
-import { BezierDefinition, EasingDefinition } from "../../easing/types"
+import { BezierDefinition, Easing, EasingDefinition } from "../../easing/types"
+
+export function isEasingSupported(easing?: Easing | Easing[]) {
+    return (
+        !easing ||
+        Array.isArray(easing) ||
+        (typeof easing === "string" && supportedWaapiEasing[easing])
```
nvh, line 7: This is pretty weird logic to me.
Why is easing always supported if `easing` is an array or easing is falsy?

nvh (changes requested):
Let's rename a bit.

packages/framer-motion/src/animation/waapi/easing.ts
mattgperry: Easing is falsy === default array
Easing is array === bezier curve

nvh: Not 100% sure, but maybe we should add that as a comment. Not feeling strongly though, so do what you prefer

nvh (approved):

Issue #884, opened 2020-12-11
[FEATURE] Method to immediately set the spring value from `useSpring`, skipping the spring animation


DanielHZhang:
**Is your feature request related to a problem? Please describe.**

Currently, there is no documented way to immediately set the current value of a `MotionValue` returned by `useSpring`. Calling `.set(value, false)` continues to play the spring animation even if render subscribers are not notified. This results in strange animation when the spring value is updated frequently. 

See: https://codesandbox.io/s/framer-motion-hover-issue-59cd0?file=/src/App.tsx
In this example, setting the spring value in the `onMouseEnter` handler and then setting it again in response to subsequent `onMouseMove` events causes the animation to look janky, as these events fire in quick succession.

**Describe the solution you'd like**

There should be some public API on the `MotionValue` class that allows you to immediately set the current value of the spring while skipping the spring animation. Exposing `.current` and `.prev` as mutable public properties, skipping the animation on `.set(value, false)`, or introducing a new method such as `.immediate(value)` are possible solutions.

**Describe alternatives you've considered**
As far as I know, the only way to do this is to mutate the private property `.current` of the `MotionValue`. This works, but Typescript does not like it.

Sirk:
I agree with @DanielHZhang. This is a missing feature. For example with `react-spring` you can do : 
```javascript
set({ x: 20, immediate: true })
```
which skips the animation, mentionned [in the docs](https://www.react-spring.io/docs/hooks/api).

The only current solution in `framer-motion` I found is similar to what @DanielHZhang mentionned, mutating `current` & `prev` to prevent spring oscillation : 

```javascript
x.current = x.prev = offsetX;
x.set(offsetX);
```

mattgperry:
I assume you'd expect this to stop any current animations if they were running?

Sirk:
> I assume you'd expect this to stop any current animations if they were running?

Yes, it would be like forcing a new value. Can be useful for example to set a position on a mouseenter event and animating it on mousemove.

HammadB:
This would be useful. Is there a preferred API - I'd be happy to take a look at implementation.

robksawyer:
Any movement on this? Or is there a better approach now?

arlyon:
As a concrete usecase, I have an item that I want to be able to move around in response to outside state, while also being able to drag it to a new value. Currently, it all works, except if I drag somwhere else, as soon as I let go, due to how it is implemented, the object snaps back to its spring location before 'springing' over.

PR #1842, opened 2022-12-21
Added `MotionValue.jump` to immediately set value while resetting animations and velocity 

mattgperry:
Closes https://github.com/framer/motion/issues/884

This PR adds an `jump` method to `MotionValue`. This sets `MotionValue` in a way that resets `velocity` and any active passive effects (currently only the spring from `useSpring()`)

### DIFF
--- a/packages/framer-motion/src/render/dom/utils/unit-conversion.ts
+++ b/packages/framer-motion/src/render/dom/utils/unit-conversion.ts
@@ -144,7 +144,7 @@ const convertChangedValueTypes = (
     changedKeys.forEach((key) => {
         // Restore styles to their **calculated computed style**, not their actual
         // originally set style. This allows us to animate between equivalent pixel units.
-        const value = visualElement.getValue(key) as MotionValue
+        const value = visualElement.getValue(key)
 
         value.set(origin[key], true)
         target[key] = positionalValues[key](targetBbox, elementComputedStyle)
--- a/packages/framer-motion/src/render/dom/utils/unit-conversion.ts
+++ b/packages/framer-motion/src/render/dom/utils/unit-conversion.ts
@@ -146,7 +146,7 @@ const convertChangedValueTypes = (
         // originally set style. This allows us to animate between equivalent pixel units.
         const value = visualElement.getValue(key)
 
-        value.set(origin[key], true)
+        value.jump(origin[key])
         target[key] = positionalValues[key](targetBbox, elementComputedStyle)
     })
 
@@ -252,7 +252,7 @@ const checkAndConvertChangedValueTypes = (
                         ? transitionEnd[key]
                         : target[key]
 
-                value.set(to, true)
+                value.jump(to)
             }
         }
     })
--- a/packages/framer-motion/src/render/dom/utils/unit-conversion.ts
+++ b/packages/framer-motion/src/render/dom/utils/unit-conversion.ts
@@ -146,7 +146,7 @@ const convertChangedValueTypes = (
         // originally set style. This allows us to animate between equivalent pixel units.
         const value = visualElement.getValue(key)
 
-        value.jump(origin[key])
+        value && value.jump(origin[key])
         target[key] = positionalValues[key](targetBbox, elementComputedStyle)
     })
 
--- a/packages/framer-motion/src/render/dom/utils/unit-conversion.ts
+++ b/packages/framer-motion/src/render/dom/utils/unit-conversion.ts
@@ -27,13 +27,6 @@ const hasPositionalKey = (target: TargetWithKeyframes) => {
     return Object.keys(target).some(isPositionalKey)
 }
 
-const setAndResetVelocity = (value: MotionValue, to: string | number) => {
-    // Looks odd but setting it twice doesn't render, it'll just
-    // set both prev and current to the latest value
-    value.set(to, false)
-    value.set(to)
-}
-
 const isNumOrPxType = (v?: ValueType): v is ValueType =>
     v === number || v === px
 
@@ -153,7 +146,7 @@ const convertChangedValueTypes = (
         // originally set style. This allows us to animate between equivalent pixel units.
         const value = visualElement.getValue(key) as MotionValue
 
-        setAndResetVelocity(value, origin[key])
+        value.set(origin[key], true)
         target[key] = positionalValues[key](targetBbox, elementComputedStyle)
     })
 
@@ -259,7 +252,7 @@ const checkAndConvertChangedValueTypes = (
                         ? transitionEnd[key]
                         : target[key]
 
-                setAndResetVelocity(value, to)
+                value.set(to, true)
             }
         }
     })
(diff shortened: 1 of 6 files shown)


packages/framer-motion/src/render/VisualElement.ts
```diff
@@ -732,7 +732,7 @@ export abstract class VisualElement<
             this.addValue(key, value)
         }
 
-        return value as MotionValue
+        return value!
```
nvh, line 735: Any reason why we can't typecheck it here? We could throw an explicit error or add an assert, right?

packages/framer-motion/src/value/index.ts
```diff
@@ -223,12 +225,17 @@ export class MotionValue<V = any> {
      *
      * @public
      */
-    set(v: V, render = true) {
-        if (!render || !this.passiveEffect) {
-            this.updateAndNotify(v, render)
+    set(v: V, immediate = false) {
```
shuangq, line 228: The TSDoc also needs to be updated (line:224)

packages/framer-motion/src/value/use-spring.ts
```diff
@@ -56,7 +60,7 @@ export function useSpring(
             })
 
             return value.get()
-        })
+        }, stopAnimation)
```
nvh, line 63: Was this something that used to be part of this `useMemo` but was captured incorrectly?

packages/framer-motion/src/value/use-spring.ts
```diff
@@ -35,6 +35,12 @@ export function useSpring(
     const activeSpringAnimation = useRef<PlaybackControls | null>(null)
     const value = useMotionValue(isMotionValue(source) ? source.get() : source)
 
+    const stopAnimation = () => {
+        if (activeSpringAnimation.current) {
+            activeSpringAnimation.current.stop()
+        }
+    }
```
nvh, line 42: This is always going to be a new function, so this will negate the benefit of `useMemo` completely, right?

packages/framer-motion/src/render/VisualElement.ts
shuangq, line 735: Can we avoid the cast? It can be undefined if `defaultValue` is also undefined, right?

packages/framer-motion/src/render/dom/utils/unit-conversion.ts
```diff
@@ -153,7 +146,7 @@ const convertChangedValueTypes = (
         // originally set style. This allows us to animate between equivalent pixel units.
         const value = visualElement.getValue(key) as MotionValue
```
nvh, line 147: This cast seems to be no longer needed

packages/framer-motion/src/value/index.ts
```diff
@@ -223,12 +225,17 @@ export class MotionValue<V = any> {
...
+        if (immediate || !this.passiveEffect) {
+            this.updateAndNotify(v)
```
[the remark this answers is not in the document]
nvh: Yeah, and API-wise it does not _have_ to be a breaking change, right?
It's just nicer to be able to call it `immediate`.

I would suggest calling this new function something else, but that is a bit too disruptive I guess.

We could maybe as easy-backwards compatible version add a function that has the old behaviour?

packages/framer-motion/src/value/index.ts
```diff
@@ -237,7 +244,7 @@ export class MotionValue<V = any> {
         this.timeDelta = delta
     }
 
-    updateAndNotify = (v: V, render = true) => {
+    updateAndNotify = (v: V) => {
```
nvh, line 247: This is also a breaking change, I guess, or is it non-public API?

nvh (changes requested):
I think we should make more effort to keep the API stable.

packages/framer-motion/src/value/index.ts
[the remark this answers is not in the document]
mattgperry: IMO this might be a breaking change in the sense that I stupidly documented the second argument (although it was never expanded upon) even though it wasn't meant to be public.

I don't think it's required to keep the existing old functionality as its usage was never particularly explained but I'd be happy to bump to `9.0.0`.

And the way we used it internally was always to achieve the same effect we're using adding the `immediate` flag for:

```
value.set(100, false)
value.set(100)

// velocity = 0 
```

The "don't render" flag was only really a means to this end.

packages/framer-motion/src/value/index.ts
mattgperry: It isn't public

packages/framer-motion/src/value/use-spring.ts
mattgperry: No. `useMemo` (and `useCallback`, `useEffect`) always also create new functions.

In this particular instance it's being used in strictly off the book way, attaching a handler on the initial render. Given this is impure it would probably be better to change this to `useInsertionEffect`

packages/framer-motion/src/render/VisualElement.ts
mattgperry: Will revert from this PR and have fixed in https://github.com/framer/motion/pull/1843

packages/framer-motion/src/value/__tests__/use-spring.test.tsx
```diff
@@ -89,6 +89,43 @@ describe("useSpring", () => {
         expect(resolved).toEqual([0, 2, 4, 7, 10, 14, 19, 24, 29, 34])
     })
 
+    test.only("will not animate if immediate=true", async () => {
```
shuangq, line 92: We probably don't want this .only?

packages/framer-motion/src/value/index.ts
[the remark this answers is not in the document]
mattgperry: Thinking it through slightly more, just going to take a quick stab in a way that retains the existing functionality

packages/framer-motion/src/value/use-spring.ts
nvh: But what I meant is that you add a value that is always different to the dependency list of `useMemo` which  will never memoize the function result.

packages/framer-motion/src/value/index.ts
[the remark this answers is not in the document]
nvh: I think we should bump the version, albeit just for keeping semver

packages/framer-motion/src/value/index.ts
```diff
@@ -4,6 +4,7 @@ import { sync } from "../frameloop"
 import type { VisualElement } from "../render/VisualElement"
 import { SubscriptionManager } from "../utils/subscription-manager"
 import { velocityPerSecond } from "../utils/velocity-per-second"
+import { warning } from "hey-listen"
```
nvh, line 7: This import seems unused.

packages/framer-motion/src/value/use-spring.ts
mattgperry: `stopAnimation`? That's being sent to `attach` not the `useMemo`. Btw I also changed `useMemo` to `useInsertionEffect` as it's a more normal place to add a listener.

PR #1843, opened 2022-12-21
Stopping active animation if no change listeners on `MotionValue`

mattgperry:
Fixes https://github.com/framer/motion/issues/1048

This PR ensures that any animation running on a motion value is stopped if the motion value has no remaining listeners. Previously they were left to play out but if one has `repeat: Infinity` then this will play out for the duration of the page.

In the process this uncovered another bug resulting from our change from `useEffect` to `useLayoutEffect` for triggering animations. These will fire before `"animationStart"` event listeners are bound via `useMotionValueEvent` because effects resolve upwards through the tree. So this has been moved to `useInsertionEffect`, which run before `useLayoutEffect`s.


packages/framer-motion/src/motion/utils/use-visual-element.ts
```diff
@@ -59,10 +59,5 @@ export function useVisualElement<Instance, RenderState>(
         }
     })
 
-    useIsomorphicLayoutEffect(
```
mattgperry, line 62: This had no listeners

packages/framer-motion/src/value/use-on-change.ts
```diff
@@ -1,30 +0,0 @@
-import { useIsomorphicLayoutEffect } from "../three-entry"
```
mattgperry, line 1: The functions in this file were each only used in one other function

packages/framer-motion/src/render/VisualElement.ts
```diff
@@ -45,7 +45,6 @@ const propEventHandlers = [
     "AnimationStart",
     "AnimationComplete",
     "Update",
-    "Unmount",
```
shuangq, line 48: Do we also need to cleanup the interface here https://github.com/framer/motion/blob/2eb37fc5fac7bd94ff57492f6e4164e93d787b50/packages/framer-motion/src/render/types.ts#L137?

packages/framer-motion/src/utils/use-motion-value-event.ts
```diff
@@ -9,7 +9,7 @@ export function useMotionValueEvent<
     event: EventName,
     callback: MotionValueEventCallbacks<V>[EventName]
 ) {
-    useIsomorphicLayoutEffect(
+    useInsertionEffect(
```
shuangq, line 12: Maybe worth a comment to explain why we're using `useInsertionEffect` here.

shuangq (approved):

Issue #584, opened 2020-06-03
[FEATURE] Ship without polyfills (tslib) 


seivan:
Edit: Technically I'd consider this a bug, because your `Rollup` plugin with Typescript forces `emitHelpers: true` which shoves `tslib` into the final bundle and from my impression the author of said library has no intention of changing that. 

**Is your feature request related to a problem? Please describe.**
The library is shipping with `tslib` which adds to the bundle sizes and duplicates existing polyfils on the consumer side (application).
This can't be tree-shaked and is `3.73kb` that can be shaved off.
This constitutes `4.3%` of `motion` and `9%` of `popmotion`.

Though I am not sure if you would have to do it for the dependencies that seem to have the same issue. 

**Describe the solution you'd like**
Ship without any polyfills as those belongs on the consumer side.
An example of this is something like `ObjectAssign` which comes from`core-js`, `React` and now this library via `tslib`.

**Describe alternatives you've considered**
Excluding `ObjectAssign` from `core-js` and `React` via some build magic, but it seems weird that it should come from an animation library. 

You could do this by adding `exclude: ["tslib"]` to your `CJS`and `ES` builds.

mattgperry:
I'll take a look at this now!

mattgperry:
Looking into it, using tslib was precisely so we didn't need to include the helpers directly in our various libraries. As far as I can tell we don't get a choice in how Typescript transpiles code like spread operators so helpers need to be included one way or the other.

seivan:
@mattgperry I don't think I understood, when you say "various" libraries, do you mean other packages that rely on motion? 
Why would they need it? They would just be used by an application that already has those.

mattgperry:
Sorry I mean all the libraries motion depends on, like Popmotion. They all share the same tslib. And any Typescript projects using Motion will also share it.

Build systems typically don’t transpile libraries which is why all the ones you mentioned have their own polyfill for es6 features. It might be that this has changed in the last year?

seivan:
> Build systems typically don’t transpile libraries which is why all the ones you mentioned have their own polyfill for es6 features, as they ship pre-transpiled. It might be that this has changed in the last year?
No, you're correct, for libraries. 
Applications would and should do it.

>  They all share the same tslib. And any Typescript projects using Motion will also share it.
No, that's the issue. It's added next to my existing polyfills.
You don't need to distribute it. 

[...]

PR #1851, opened 2022-12-28
Unpin tslib dependency

sirreal:
Unpin the tslib dependency to help reduce duplication in dependants.

This dependency was pinned at 2.4.0 by #1663. No mention of why the
range was changed to an exact version is made there.

With the exact version, dependents will likely get versions 2.4.1
(latest at the moment) and 2.4.0 (required by framer-motion).

This goes against the intention of using `tslib`, which is to prevent
duplication of TypeScript helpers from being duplicated across the
application (see #584).


mattgperry:
Ah very nice, thanks!

PR #1860, opened 2023-01-02
Adding warning when reduced motion is enabled in development mode

mattgperry:


eelco (commented):
`warnOnce` always posts a message, correct? (Not just in development mode?)

mattgperry:
> `warnOnce` always posts a message, correct? (Not just in development mode?)

Ah good catch thanks, I’m used to warn and invariant handling this

Issue #1066, opened 2021-03-27
[BUG] Toggling a motion.div style based on a MotionValue does not work as expected in v3.4.0+


parahippo:
Bug Description:

The rendering behavior of a motion.div style based on a MotionValue changed starting with 3.4.0 and I believe the behavior is incorrect, although I am new to React and am not 100% confident in my conclusion. I have a boolean prop flag determining if the background color of a motion.div is a solid color or based on a MotionValue. This worked fine prior to 3.4.0 but starting with 3.4.0 the solid color value is ignored and the last MotionValue color is retained (from a cached value I'm guessing).

Code Sandbox:

https://codesandbox.io/s/wispy-surf-eievc

Steps to reproduce:
1. Go to the code sandbox (https://codesandbox.io/s/wispy-surf-eievc)
2. Scroll the browser frame. Note that the "Closed Menu" goes from transparent to white as you scroll
3. Click the "Toggle Menu" button. The "Closed Menu" will change to showing items in the menu. Based on the code, the items in the menu should be on an all white background, but instead the background is the last shade of the progressive scroll value.
4. Switch to framer-motion 3.3.0 under Dependencies
5. Repeat steps 2-3. Note that when the menu is opened, the background switches to all white, consistent with the code. When the menu is close, the background color switches back to the progressive MotionValue based on the scroll position.

Expected behavior:

I would expect that the behavior seen in 3.3.0 would be retained. Namely that when a prop change happens, the rendering respects that value of the prop flag and changes the corresponding style of the component.

Environment details:

Chrome (latest version), Mac OS Mojave

parahippo:
#1023 could be a related issue.

PR #1863, opened 2023-01-03
Fixing switching style between motion value and static value

mattgperry:
Fixes https://github.com/framer/motion/issues/1066

### DIFF
--- a/packages/framer-motion-3d/src/render/utils/scrape-motion-value.ts
+++ b/packages/framer-motion-3d/src/render/utils/scrape-motion-value.ts
@@ -17,14 +17,15 @@ const valueMap = {
 }
 
 export const scrapeMotionValuesFromProps: ScrapeMotionValuesFromProps = (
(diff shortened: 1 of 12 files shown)


packages/framer-motion-3d/src/render/utils/scrape-motion-value.ts
```diff
@@ -17,14 +17,15 @@ const valueMap = {
...
 ) => {
     const motionValues = {}
 
     for (const key in props) {
         const prop = props[key]
 
-        if (isMotionValue(prop)) {
+        if (isMotionValue(prop) || isMotionValue(prevProps[key])) {
             motionValues[valueMap[key] || key] = prop
         } else if (Array.isArray(prop)) {
```
shuangq, line 30: Do we need to handle the case where an array of motion values switches to an array of staic values?

shuangq (approved):

PR #1864, opened 2023-01-03
Removing browser-safe `process.env.NODE_ENV`

mattgperry:
This wasn't minifying correctly, meaning development-only error messages end up in production bundles (even though they never display).

Additionally, in https://github.com/framer/motion/blob/main/packages/framer-motion/src/render/utils/motion-values.ts#L34 there has been a `process.env.NODE_ENV` for a while that hasn't generated any error reports for us.

### DIFF
--- a/packages/framer-motion/src/value/scroll/use-element-scroll.ts
+++ b/packages/framer-motion/src/value/scroll/use-element-scroll.ts
@@ -6,9 +6,12 @@ import { useScroll } from "../use-scroll"
  * @deprecated useElementScroll is deprecated. Convert to useScroll({ container: ref })
  */
 export function useElementScroll(ref: RefObject<HTMLElement>) {
-    warnOnce(
-        false,
-        "useElementScroll is deprecated. Convert to useScroll({ container: ref })."
-    )
+    if (process.env.NODE_ENV === "development") {
+        warnOnce(
+            false,
+            "useElementScroll is deprecated. Convert to useScroll({ container: ref })."
+        )
+    }
+
     return useScroll({ container: ref })
 }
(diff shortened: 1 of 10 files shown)


huntercaron:
Is there no other way to fix this? This kinda sucks that it will not run in the browser without stubbing process now

mattgperry:
It couldn't have been running without stubbing because there was already a `process.env.NODE_ENV` in there. Given this, it feels like the course of action should be

1. Standardise usage and fix minification
2. If we receive bug reports (we already should have if this was a problem) open a PR that fixes without increasing reported bundlesizes

packages/framer-motion/src/value/scroll/use-element-scroll.ts
```diff
@@ -6,9 +6,12 @@ import { useScroll } from "../use-scroll"
  * @deprecated useElementScroll is deprecated. Convert to useScroll({ container: ref })
  */
 export function useElementScroll(ref: RefObject<HTMLElement>) {
-    warnOnce(
-        false,
-        "useElementScroll is deprecated. Convert to useScroll({ container: ref })."
-    )
+    if (process.env.NODE_ENV === "development") {
```
huntercaron, line 9: What is the reason to sometimes use `process.env.NODE_ENV === "development"` and sometimes `process.env.NODE_ENV !== "production"`?

huntercaron (approved):
Just one question on the check, otherwise I checked the bundles and didn't notice any bare usages ✅ nice lil bundle size reduction

packages/framer-motion/src/value/scroll/use-element-scroll.ts
mattgperry: There's some tests in there that check whether `console.warn` was called and this is to catch `NODE_ENV === "test"`

Issue #1867, opened 2023-01-04
[BUG] Interrupting WAAPI animations with repeat/delay options not working as expected


mattgperry:
Currently the `sample` method assumes  we can resolve animations for `time` but this is only true for the underlying animations, not the animation runner (which is what applies repeat and delay).

A refactor would be needed to fix this properly but in the meantime it's probably quick enough to run a low-resolution animation until `currentTime`.

PR #1868, opened 2023-01-04
Fixing animation sampling

mattgperry:
Fixes https://github.com/framer/motion/issues/1867

### DIFF
--- a/packages/framer-motion/jest.setup.tsx
+++ b/packages/framer-motion/jest.setup.tsx
@@ -30,6 +30,10 @@ if (!(global as any).ResizeObserver) {
     } as any
 }
 
+// Stub scrollTo
+const noop = () => {}
+Object.defineProperty(global, "scrollTo", { value: noop, writable: true })
+
 export const click = (element: Element) =>
     act(() => {
         fireEvent.click(element)
--- a/packages/framer-motion/src/animation/legacy-popmotion/index.ts
+++ b/packages/framer-motion/src/animation/legacy-popmotion/index.ts
@@ -68,7 +68,6 @@ export function animate<V = number>({
     let driverControls: DriverControls | undefined
     let repeatCount = 0
     let computedDuration: number | undefined = duration
-    let latest: V
     let isComplete = false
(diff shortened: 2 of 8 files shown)


packages/framer-motion/src/animation/waapi/create-accelerated-animation.ts
```diff
@@ -88,10 +89,10 @@ export function createAcceleratedAnimation(
          */
         const { currentTime } = animation
         if (currentTime) {
-            const sampleAnimation = animate(options)
+            const sampleAnimation = animate({ ...options, autoplay: false })
```
shuangq, line 92: What does `autoplay: false` do here?

packages/framer-motion/src/animation/legacy-popmotion/index.ts
```diff
@@ -164,10 +165,27 @@ export function animate<V = number>({
...
+         * temporal-resolution version.
+         */
+        sample: (t: number): V => {
+            elapsed = initialElapsed
+            const sampleResolution =
+                typeof duration === "number" ? duration * 0.5 : 50
+
+            let sampleElapsed = 0
+            update(0)
+            while (sampleElapsed <= t) {
```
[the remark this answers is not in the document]
mattgperry: Ah yeah this is a great point, let me add a test for this.

packages/framer-motion/jest.setup.tsx
```diff
@@ -30,6 +30,10 @@ if (!(global as any).ResizeObserver) {
     } as any
 }
 
+// Stub scrollTo
+const noop = () => {}
+Object.defineProperty(global, "scrollTo", { value: noop, writable: true })
+
```
shuangq, line 36: Why do we need this in the PR?

shuangq (approved):

packages/framer-motion/jest.setup.tsx
mattgperry: It's unrelated but removes some noise from the Jest console output

PR #1870, opened 2023-01-04
Removing ability to start animations within the render function

mattgperry:
In the early days of Framer, back before concurrent rendering, people would write code like this

```jsx
function Component() {
  const controls = useAnimationControls()

  controls.start({ x: 100 })
  
  return <motion.div animate={controls} />
}
```

Because it usually worked (even though it was always a poor idea). However now with concurrent rendering, this is likely to break as components can be rendered multiple times, or prospectively with props that don't end up committed to the DOM.

This PR changes the behaviour here so that rather than obfuscating this by saving up pending animations on the initial render, we throw an error.


hemlok (approved):
👍

Issue #1418, opened 2022-01-12
[BUG] transitions defined inside variants are not used in Framer Motion 3D


itssharlee:
**Describe the bug**

Framer Motion 3D uses the default `transition` instead of the one provided in each variant on version `5.5.0` and later.

**IMPORTANT: Provide a CodeSandbox reproduction of the bug**

[https://codesandbox.io/s/framer-motion-transitions-demo-l24qt?file=/src/App.js](https://codesandbox.io/s/framer-motion-transitions-demo-l24qt?file=/src/App.js)

**Steps to reproduce**

1. Click on the button 'To the left'
2. Click on the button 'To the right'
-> The transitions are the same, even if different transitions are set in the variants

**Expected behavior**

Transitions used must be the ones defined in each variant instead of using the default transitions.

**Environment details**

Should not work in all browsers (Tested in Chrome and Firefox)

itssharlee:
I found the fix, we now need to wrap our scene in a `MotionCanvas` instead of a simple `Canvas` component.

cdebotton:
I'm encountering this bug using `framer-motion-3d@6.2.8` with `@react-three/fiber@7.0.26`. The `transition` property is always ignored.

roelkok:
Can this issue be re-opened? I also have the issue with `framer-motion-3d@6.3.3`

PR #1874, opened 2023-01-05
Fixing `transition` in variants in Framer Motion 3D

mattgperry:
Fixes https://github.com/framer/motion/issues/1418

### DIFF
--- a/dev/examples/ThreeVariant.tsx
+++ b/dev/examples/ThreeVariant.tsx
@@ -1,0 +1,58 @@
+import * as React from "react"
+import { MotionConfig, motion as motionDom } from "framer-motion"
+import { motion, MotionCanvas } from "framer-motion-3d"
+import { Canvas, extend } from "@react-three/fiber"
+import {
+    AmbientLight,
+    PointLight,
+    Group,
+    BoxGeometry,
+    MeshStandardMaterial,
+    Mesh,
+} from "three"
+
+extend({
+    AmbientLight,
+    PointLight,
+    Group,
+    BoxGeometry,
+    MeshStandardMaterial,
+    Mesh,
+})
+
+/**
+ * An example of firing an animation onMount using the useAnimation hook
+ */
+
+function Box(props) {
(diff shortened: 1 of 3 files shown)


dev/examples/ThreeVariant.tsx
```diff
@@ -0,0 +1,58 @@
...
+} from "three"
+
+extend({
+    AmbientLight,
+    PointLight,
+    Group,
+    BoxGeometry,
+    MeshStandardMaterial,
+    Mesh,
+})
```
huntercaron, line 21: Why do you need this? All these types should come in from `@react-three/fiber` automatically

mattgperry: Since 8 I think they need pulling in manually https://docs.pmnd.rs/react-three-fiber/tutorials/v8-migration-guide#tree-shaking-via-extend

huntercaron: I've been using 8 for a while and never had this? I use it quite often. You also don't see it in recent sandboxes by their team https://codesandbox.io/s/meshepoxymaterial-hmgdjq?file=/src/App.js

I think what you linked to is an example if you want to do more aggressive tree shaking? Or maybe all these examples have some babel plugin hidden that you don't have in your setup here?

huntercaron: I've never used extend for these and I use framer-motion-3d and r3f with nextjs, vite, etc often, something must be up. Does it break if you remove?

hemlok (approved):

dev/examples/ThreeVariant.tsx
hemlok: Its just for tree shaking right?

mattgperry: @huntercaron Yeah I get an error otherwise. It used to point to the page I've linked but now it only mentions third-party components. Maybe I'll reach out to Paul see if he knows what's going on

Issue #1464, opened 2022-02-21
Forcing borderRadius as percentage breaks aspect ratio on resize


RickyRomero:
For layout animations, `framer-motion` reinterprets `borderRadius` as a percentage of the element's size as a performance optimization:

https://github.com/framer/motion/blob/2ab6d8bdb89cfc89214ccb0394a971c03625407d/packages/framer-motion/src/projection/styles/scale-border-radius.ts#L10-L16

However, in layouts where the element's size is flexible based on the viewport, framer-motion doesn't update these percentages when the element's size changes as a result of resizing the viewport.

[Please refer to this CodeSandbox](https://codesandbox.io/s/charming-haze-83m7x8?file=/src/App.js) for a simplified test case. To reproduce the issue, resize the viewport after clicking on the element. Notice that the `borderRadius` isn't corrected for the updated size.

PR #1876, opened 2023-01-06
Removing scale correction for values with no active projection transform

mattgperry:
Fixes https://github.com/framer/motion/issues/1464


dev/projection/script-assert.js
```diff
@@ -50,7 +50,7 @@ window.Assert = {
     },
     matchBorderRadius: (element, expected) => {
         let radius = element.style.borderRadius
-
+        console.log(radius)
```
ankon, line 53: Left-over debugging?

packages/framer-motion/src/projection/node/create-projection-node.ts
```diff
@@ -1624,7 +1624,17 @@ export function createProjectionNode<I>({
                 if (valuesToRender[key] === undefined) continue
 
                 const { correct, applyTo } = scaleCorrectors[key]
-                const corrected = correct(valuesToRender[key], lead)
+
+                /**
+                 * Only apply scale correction to the value if we have an
+                 * active projection transform. Otherwise these values become
+                 * vulnerable to distortion if the element changes size without
+                 * a corressponding layout animation.
```
ankon, line 1632:
```suggestion
                 * a corresponding layout animation.
```

PR #1877, opened 2023-01-06
Accept motion values as children

mattgperry:
This PR allows `motion` components to accept a `MotionValue` as its child.

When the `MotionValue` updates, its latest value is directly assigned to `element.textContent`, making it straightforward to animate content without triggering React state.

```jsx
const value = useMotionValue(0)

return <motion.span>{value}</motion.span>
```

Fixes https://github.com/framer/motion/issues/1476

### DIFF
--- a/packages/framer-motion/package.json
+++ b/packages/framer-motion/package.json
@@ -39,7 +39,7 @@
         "clean": "rm -rf types dist lib",
         "test": "yarn test-server && yarn test-client",
         "test-ci": "yarn test",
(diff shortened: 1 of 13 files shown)


packages/framer-motion/package.json
```diff
@@ -39,7 +39,7 @@
         "clean": "rm -rf types dist lib",
         "test": "yarn test-server && yarn test-client",
         "test-ci": "yarn test",
-        "test-client": "jest --config jest.config.json --max-workers=2 --detectOpenHandles",
```
mattgperry, line 42: This was erroneously left in from a previous PR

hemlok (approved):

PR #1880, opened 2023-01-07
staggerDirection Api Doc Fix

SyedImam1998:
#### Hi team,

I have noticed a small mistake in Api Documentation related to staggerDirection property. delayChildren should be replaced with staggerChildren without that Developers cannot notice the animation of staggerDirection.

#### Before:

framer1

#### After:

framer2


mattgperry (approved):

mattgperry:
Ah yeah good spot, thanks!

Issue #1485, opened 2022-03-13
[BUG] Non-MotionValue variables are still considered a MotionValue when using useMotionTemplate


thexpand:
**1. Describe the bug**
Using non-MotionValue variables in the interpolation/template string, triggers an error.
```const nonWorkingTemplateValue = useMotionTemplate`${n}px ${commonVariable}`;```
The code above produces a `values[i].get is not a function` error, because `framer-motion` doesn't check the type of the template string variables before calling the `[i].get()` function on them. You can check the CodeSandbox.

Probably a type check would be enough to be added here:
https://github.com/framer/motion/blob/069165ebd21ed5dcb56d4f219d7ceee5ca71c67b/packages/framer-motion/src/value/use-motion-template.ts#L41

**2. Provide a CodeSandbox reproduction of the bug**
https://codesandbox.io/s/usemotiontemplate-not-interpolating-right-ix9lb6?file=/src/App.js:262-282

**3. Steps to reproduce**
1. Go to the CodeSandbox
2. Check the definition of the non-working template variable:
```const nonWorkingTemplateValue = useMotionTemplate`${n}px ${commonVariable}`;```
4. See how it is not working and you get a type error `values[i].get is not a function` as `framer-motion` tries to call the function on the `commonVariable`.

**5. Expected behavior**
I expect only the MotionValue variables to be interpolated and interpreted as MotionValue, not other values like strings. I.e. the `commonVariable` should not be considered and interpreted as a MotionValue.

In the CodeSandbox you can switch for the `workingTemplateValue` and comment out the `nonWorkingTemplateValue` to see how it is supposed to work.

**6. Environment details**
- Windows 10 (10.0.19044 Build 19044)
- Chrome 99.0.4844.51 / Firefox 98.0 / Edge 99.0.1150.39
- framer-motion@6.2.8

PR #1887, opened 2023-01-09
Allow static values in useMotionValueTemplate

mattgperry:
This PR allows static values to be used in `useMotionValueTemplate`

Fixes https://github.com/framer/motion/issues/1485


hemlok (approved):
Nice, ran into this the other day to boot

Issue #1702, opened 2022-09-15
[BUG] 8-digit hex color not being applied on animate box-shadow prop


irangarcia:
**1. Read the [FAQs](#faqs) 👇**

**2. Describe the bug**

When trying to apply an 8-digit hexadecimal color to the box-shadow `animate` prop it doesn't work. The same value converted to `rgba` works.

**3. IMPORTANT: Provide a CodeSandbox reproduction of the bug**

https://codesandbox.io/s/funny-elgamal-ecdukn?file=/src/Example.tsx

**4. Steps to reproduce**

1. Go to the [sandbox](https://codesandbox.io/s/funny-elgamal-ecdukn?file=/src/Example.tsx)
2. Comment line 9
3. Uncomment line 11
4. See that the box-shadow was applied just with the `rgba` color

**5. Expected behavior**

Apply the box-shadow when using an 8-digit hexadecimal color

**6. Video or screenshots**

-

**7. Environment details**

Ubuntu 20.04
Chrome 105.0.5195.102

Issue #1677, opened 2022-08-23
[BUG] Animation between colors with alpha values not working


MasaLion:
I wanted to change the border color of an element using whileHover, from `#ffffff44` to `#f1f1f1aa`, which doesn't work. 
All the other changes happen, but the color stays the same. 

The color does change when the I stop using alpha values for the latter - `#f1f1f1`.
Give a clear and concise description of what the bug is.

To reproduce - 
- Go to CodeSandbox: https://codesandbox.io/s/sharp-frog-4w965m?file=/src/App.js
- Comment out line 15 and uncomment line 14

PR #1889, opened 2023-01-09
Fixing detection of hex colors with alpha in complex strings

mattgperry:
Fixes https://github.com/framer/motion/issues/1677
Fixes https://github.com/framer/motion/issues/1702

### DIFF
--- a/packages/framer-motion/src/value/types/utils.ts
+++ b/packages/framer-motion/src/value/types/utils.ts
@@ -9,9 +9,9 @@ export const sanitize = (v: number) => Math.round(v * 100000) / 100000
 
 export const floatRegex = /(-)?([\d]*\.?[\d])+/g
 export const colorRegex =
-    /(#[0-9a-f]{6}|#[0-9a-f]{3}|#(?:[0-9a-f]{2}){2,4}|(rgb|hsl)a?\((-?[\d\.]+%?[,\s]+){2}(-?[\d\.]+%?)\s*[\,\/]?\s*[\d\.]*%?\))/gi
+    /(#[0-9a-f]{2,8}|(rgb|hsl)a?\((-?[\d\.]+%?[,\s]+){2}(-?[\d\.]+%?)\s*[\,\/]?\s*[\d\.]*%?\))/gi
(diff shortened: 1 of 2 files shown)


packages/framer-motion/src/value/types/utils.ts
```diff
@@ -9,9 +9,9 @@ export const sanitize = (v: number) => Math.round(v * 100000) / 100000
 
 export const floatRegex = /(-)?([\d]*\.?[\d])+/g
 export const colorRegex =
-    /(#[0-9a-f]{6}|#[0-9a-f]{3}|#(?:[0-9a-f]{2}){2,4}|(rgb|hsl)a?\((-?[\d\.]+%?[,\s]+){2}(-?[\d\.]+%?)\s*[\,\/]?\s*[\d\.]*%?\))/gi
+    /(#[0-9a-f]{2,8}|(rgb|hsl)a?\((-?[\d\.]+%?[,\s]+){2}(-?[\d\.]+%?)\s*[\,\/]?\s*[\d\.]*%?\))/gi
```
nvh, line 12: Why is this `8` and not `7`?
I would assume that it can be 3 (`#fff`), `4` (`#fff0`) `6` (`#00aaff`) and `7` (`#00aaff0`)

nvh: So when exactly is it `2` or `8`?

nvh (changes requested):

packages/framer-motion/src/value/types/utils.ts
nvh: Ah, the opacity can be two hex digits too.

nvh: But then the minimum would still be `3`, right?

mattgperry: Ah yeah it is 3

mattgperry: (8 being 2 digit alpha)

nvh (approved):

Issue #1567, opened 2022-06-02
[BUG] reducedMotion not working with AnimatePresence components


panmona:
**2. Describe the bug**

When using the MotionConfig to enable reduced motion, it doesn't reduce motion on components that are animated using AnimatePresence.

**3. IMPORTANT: Provide a CodeSandbox reproduction of the bug**
https://codesandbox.io/s/framer-motion-image-gallery-forked-f3gic5 (**_Warning_**: contains animations that may be problematic for some viewers)

This is mostly the same as the codesandbox linked from the [docs](https://www.framer.com/docs/animate-presence/#usage) with the following changes:
- Updated framer to the latest stable version
- Added MotionConfig wrapper with reducedMotion="always"
- Added rotation and scale to transition and variants. (I would expect these to be disabled when using reducedMotion="always")

**4. Steps to reproduce**

1. Click on the 'next' button
2. See that the rotate and scale animation doesn't get disabled

**5. Expected behavior**
I would expect that transform and layout animations are disabled for AnimatePresence components, when using reducedMotion.

For other components it works [great](https://codesandbox.io/s/framer-motion-keyframes-forked-49j6jv)

panmau:
Were you able to confirm that this is not working as intended in this case?

I would be willing to contribute a pull request for the bugfix but I would need some pointers. I couldn't yet find out what exactly is going wrong from reading the code related to this.

felds:
I have the same problem. I tried wrapping the whole app, the AnimatePresence container, the individual motion elements, and all of them at the same time in MotionConfig elements but the result is the same.

nwesthoff:
Also experiencing this. @panmona's example was using `framer-motion@6.3.6`. Confirming this issue still exists after upgrading to current latest: `framer-motion@7.3.6`.

panmona:
@mattgperry Could you take a look at this? I would be willing to further investigate this and contribute a PR, but I would need some pointers on where to start. Maybe you have an idea what could be wrong here.

panmona:
Thanks for fixing this!
This seems to work perfectly now: https://codesandbox.io/s/framer-motion-image-gallery-forked-0qhbif

PR #1891, opened 2023-01-11
Fix `reducedMotion` when `transition` is explicitly defined

mattgperry:
Fixes https://github.com/framer/motion/issues/1567


yarn.lock
```diff
@@ -7909,8 +7909,8 @@ __metadata:
     cache-loader: ^1.2.5
     convert-tsconfig-paths-to-webpack-aliases: ^0.9.2
     fork-ts-checker-webpack-plugin: ^6.2.0
-    framer-motion: ^8.4.0
-    framer-motion-3d: ^8.4.0
+    framer-motion: ^8.4.1
```
shuangq, line 7912: Why yarn lock updates in this PR? 🤔

shuangq (approved):

yarn.lock
mattgperry: probably just uncommited from master

PR #1902, opened 2023-01-18
Fix primary pointer filtering for mice

mattgperry:


packages/framer-motion/src/events/utils/is-primary-pointer.ts
```diff
@@ -1,6 +1,15 @@
-/**
- * Specifically match against false here as incomplete versions of
- * PointerEvents in very old browser might have it set as undefined.
- */
-export const isPrimaryPointer = (event: PointerEvent) =>
-    event.isPrimary !== false
+export const isPrimaryPointer = (event: PointerEvent) => {
+    if (event.pointerType === "mouse") {
+        return typeof event.button !== "number" || event.button <= 0
```
nvh, line 3: Why `button <= 0`? Can button ever be less than `0`?

nvh (approved):
I think we should just check `=== 0` for the button, but don't feel strongly

packages/framer-motion/src/events/utils/is-primary-pointer.ts
mattgperry: Yeah it can be `-1` for no button, I wasn't super confident that all browsers will implement it this way which is why there's also a type check

PR #1905, opened 2023-01-19
Defer real optimised animation until Chrome is ready to paint

mattgperry:
https://bugs.chromium.org/p/chromium/issues/detail?id=1406850

### DIFF
--- a/dev/package.json
+++ b/dev/package.json
@@ -1,6 +1,6 @@
 {
     "name": "framer-motion--dev",
-    "version": "8.5.3-alpha.0",
+    "version": "8.5.2-alpha.4",
     "private": true,
     "scripts": {
         "dev": "webpack serve --config ./webpack/config.js --hot"
@@ -8,8 +8,8 @@
     "dependencies": {
         "@react-three/drei": "^7.27.3",
         "@react-three/fiber": "^8.2.2",
-        "framer-motion": "^8.5.3-alpha.0",
-        "framer-motion-3d": "^8.5.3-alpha.0",
+        "framer-motion": "^8.5.2-alpha.4",
+        "framer-motion-3d": "^8.5.2-alpha.4",
         "path-browserify": "^1.0.1",
         "react": "^18.2.0",
         "react-dom": "^18.2.0",
--- a/dev/package.json
+++ b/dev/package.json
@@ -1,6 +1,6 @@
 {
     "name": "framer-motion--dev",
-    "version": "8.5.2-alpha.2",
+    "version": "8.5.2-alpha.3",
     "private": true,
     "scripts": {
         "dev": "webpack serve --config ./webpack/config.js --hot"
@@ -8,8 +8,8 @@
     "dependencies": {
         "@react-three/drei": "^7.27.3",
         "@react-three/fiber": "^8.2.2",
-        "framer-motion": "^8.5.2-alpha.2",
-        "framer-motion-3d": "^8.5.2-alpha.2",
+        "framer-motion": "^8.5.2-alpha.3",
+        "framer-motion-3d": "^8.5.2-alpha.3",
         "path-browserify": "^1.0.1",
         "react": "^18.2.0",
         "react-dom": "^18.2.0",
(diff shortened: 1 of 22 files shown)


dev/package.json
```diff
@@ -1,15 +1,15 @@
 {
     "name": "framer-motion--dev",
-    "version": "8.5.1",
+    "version": "8.5.2-alpha.4",
```
nvh, line 3: This needs to be non-alpha before we merge, right?

packages/framer-motion/src/animation/optimized-appear/handoff.ts
```diff
@@ -1,24 +1,43 @@
...
-    value: MotionValue
+    value: MotionValue,
+    /**
+     * This function is loaded via window by startOptimisedAnimation.
+     * By accepting `sync` as an argument, rather than using it via
+     * import, it can be kept out of the first-load Framer bundle,
+     * while also allowing this function to not be included in
+     * Framer Motion bundles where it's not needed.
+     */
+    sync: Sync
```
nvh, line 18: Is this the old `framesync`?

packages/framer-motion/src/render/utils/animation.ts
```diff
@@ -171,15 +171,16 @@ function animateTarget(
          * If this is the first time a value is being animated, check
          * to see if we're handling off from an existing animation.
          */
-        if (!value.hasAnimated) {
+        if (window.HandoffAppearAnimations && !value.hasAnimated) {
```
nvh, line 174: When are these set on window?

packages/framer-motion/src/animation/optimized-appear/start.ts
```diff
@@ -2,21 +2,60 @@ import { appearStoreId } from "./store-id"
...
+    onReady?: (animation: Animation) => void
+): void {
     const id = element.dataset[optimizedAppearDataId]
-    const animation = animateStyle(element, name, keyframes, options)
 
-    if (id && animation) {
-        window.MotionAppearAnimations.set(appearStoreId(id, name), animation)
+    if (!id) return
+
+    window.HandoffAppearAnimations = handoffOptimizedAppearAnimation
```
nvh, line 19: Ah, found it :)

packages/framer-motion/src/animation/optimized-appear/start.ts
```diff
@@ -2,21 +2,60 @@ import { appearStoreId } from "./store-id"
...
+
+    const storeId = appearStoreId(id, name)
+
+    /**
+     * Use a dummy animation to detect when Chrome is ready to start
+     * painting the page and hold off from triggering the real animation
+     * until then.
```
nvh, line 26: Shall we link to the chrome bug here?

packages/framer-motion/src/animation/optimized-appear/start.ts
```diff
@@ -2,21 +2,60 @@ import { appearStoreId } from "./store-id"
...
+     */
+    const readyAnimation = animateStyle(
+        element,
+        name,
+        [keyframes[0] as number, keyframes[0] as number],
+        { duration: 10000, ease: "linear" }
```
nvh, line 32: Why is the duration 10 secs here?

nvh (approved):
Couple of questions, but looks good in general

dev/package.json
mattgperry: Before publish but not so fussed about merge, normally the version bumps happen on `main`

packages/framer-motion/src/animation/optimized-appear/handoff.ts
mattgperry: Yeah I moved it in house so to speak

dev/package.json
nvh: 👌🏻

packages/framer-motion/src/animation/optimized-appear/start.ts
mattgperry: I'll add a comment, it's basically long enough for a `ready` to fire - obviously this will usually be low `ms`

nvh: Ah, right, so we're waiting until the _start_ of this "fake" animation, right? Yeah, we should add a comment about that.

PR #1912, opened 2023-01-25
Fix optimised appear animation handoff in Firefox

mattgperry:


packages/framer-motion/src/animation/optimized-appear/handoff.ts
```diff
@@ -66,7 +68,13 @@ export function handoffOptimizedAppearAnimation(
...
         sync.render(cancelOptimisedAnimation)
 
-        return animation.currentTime || 0
+        /**
+         * We use main thread timings vs those returned by Animation.currentTime as it
+         * can be the case, particularly in Firefox, that currentTime doesn't return
+         * an updated value for several frames, even as the animation plays smoothly via
+         * the GPU.
+         */
+        return sampledTime - startTime || 0
```
nvh, line 77: Is this particularly or only in Firefox? Is there any (future) benefit of using `.currentTime`?

nvh (approved):

packages/framer-motion/src/animation/optimized-appear/handoff.ts
mattgperry: I think Firefox is most egregious but in practise there is a slight mismatch in every browser between the time we're returning here now vs `.currentTime`.

`.currentTime` is what allowed us to write tests like

```jsx
animation.pause()
animation.currentTime = 500

React.hydrateRoot(<App />)
```

And know we should expect to handoff at `500` ms exactly. Whereas this approach doesn't take into consideration any changes in the animation playback. We know these optimised animations aren't getting paused or fastforwarded outside the tests though.

PR #1913, opened 2023-01-27
Limit dirty node propagation

mattgperry:
This PR changes the propagation of "dirty" nodes to limit recalculation to:
1. Dirty nodes
2. Nearest **projecting** children of dirty nodes
3. Shared nodes where parent projections or transforms have been dirtied

In the included stress test this is a 90% reduction in executed JS per frame - this will vary per tree.

Closes https://github.com/framer/motion/pull/1798

### DIFF
--- a/dev/projection/script-assert.js
+++ b/dev/projection/script-assert.js
@@ -13,6 +13,15 @@ function showError(element, msg) {
     }
 }
 
+window.ProjectionFrames = []
+window.MotionDebug = {
+    record: (action) => {
+        if (action.type === "projectionFrame") {
+            window.ProjectionFrames.push({ ...action })
+        }
+    },
+}
+
 window.Assert = {
     matchViewportBox: (element, expected, threshold = 0.01) => {
         const bbox = element.getBoundingClientRect()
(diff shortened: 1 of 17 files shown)


dev/projection/script-assert.js
```diff
@@ -13,6 +13,15 @@ function showError(element, msg) {
     }
 }
 
+window.ProjectionFrames = []
+window.MotionDebug = {
```
nvh, line 17: Should we prefix this with a motion-specific prefix?

dev/projection/script-assert.js
```diff
@@ -83,6 +92,32 @@ window.Assert = {
             left: left - x,
         }
     },
+    checkFrames(element, frameIndex, expected) {
```
nvh, line 95: If we pass an index, shouldn't this be `checkFrame`?

dev/projection/script-assert.js
nvh: Can't this script be TypeScript, btw?

packages/framer-motion/src/projection/node/create-projection-node.ts
```diff
@@ -672,9 +695,18 @@ export function createProjectionNode<I>({
          * the next step.
          */
         updateProjection = () => {
+            // Reset projection frame data
+            projectionFrameData.totalNodes =
+                projectionFrameData.resolvedTargetDeltas =
+                projectionFrameData.recalculatedProjection =
+                    0
```
nvh, line 702: Is this significantly faster? Because it reads weird.

packages/framer-motion/src/projection/node/create-projection-node.ts
```diff
@@ -1192,6 +1248,8 @@ export function createProjectionNode<I>({
 
                 this.notifyListeners("projectionUpdate", target)
             }
+
+            projectionFrameData.recalculatedProjection++
```
nvh, line 1252: What happens here? Can we add a comment?

nvh (changes requested):
Have some remarks on naming an comments, but approach looks good

packages/framer-motion/src/debug/types.ts
```diff
@@ -0,0 +1,16 @@
+export interface ProjectionFrame {
+    type: "projectionFrame"
```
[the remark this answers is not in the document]
mattgperry: Anticipating more payload types in the future

packages/framer-motion/src/projection/node/create-projection-node.ts
mattgperry: Not faster but no object creation per frame to reduce GC

nvh:
Feel free to re-request once the requested changes are adressed.

dev/projection/script-assert.js
mattgperry: It could be TypeScript but it would just add more overhead to what is meant to be a simple test script. Likewise the `ProjectionFrames` and `MotionDebug` vars are just read in the test files, rather than something that will be used by Motion

packages/framer-motion/src/projection/node/create-projection-node.ts
mattgperry: Added comments to all debug counters

nvh (approved):

PR #1918, opened 2023-01-31
Making tap events keyboard accessible

mattgperry:
- Add `tabindex` if an element has tap events to allow element to receive focus.
- Fires `onTapStart` if Enter is pressed on a focussed element.
- Fires `onTap` if Enter is released on a focussed element.
- Fires `onTapCancel` is `blur`ed before Enter is released.
- Starts/stops `whileTap` as appropriate.

### DIFF
--- a/packages/framer-motion/src/gestures/use-tap-gesture.ts
+++ b/packages/framer-motion/src/gestures/use-tap-gesture.ts
@@ -8,9 +8,15 @@ import { isDragActive } from "./drag/utils/lock"
 import { FeatureProps } from "../motion/features/types"
 import { pipe } from "../utils/pipe"
 import { addDomEvent, useDomEvent } from "../events/use-dom-event"
-import { extractEventInfo } from "../events/event-info"
+import {
+    EventListenerWithPointInfo,
+    extractEventInfo,
+} from "../events/event-info"
 
-function fireSyntheticPointerEvent(name: string, handler?: any) {
+function fireSyntheticPointerEvent(
+    name: string,
+    handler?: EventListenerWithPointInfo
+) {
     if (!handler) return
     const syntheticPointerEvent = new PointerEvent("pointer" + name)
     handler(syntheticPointerEvent, extractEventInfo(syntheticPointerEvent))
--- a/packages/framer-motion/src/gestures/use-tap-gesture.ts
+++ b/packages/framer-motion/src/gestures/use-tap-gesture.ts
@@ -37,6 +37,7 @@ export function useTapGesture({
     const hasPressListeners = onTap || onTapStart || onTapCancel || whileTap
     const isPressing = useRef(false)
     const cancelPointerEndListeners = useRef<Function | null>(null)
+    const suspendedTabIndex = useRef<string | null>(null)
 
     /**
      * Only set listener to passive if there are no external listeners.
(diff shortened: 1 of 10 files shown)


packages/framer-motion/src/gestures/use-tap-gesture.ts
```diff
@@ -70,36 +78,44 @@ export function useTapGesture({
...
-            )
+        latestProps.onTapStart?.(event, info)
+    }
 
-            const latestProps = visualElement.getProps()
+    const callbackDependencies = [
+        Boolean(onTapStart),
+        Boolean(onTap),
+        Boolean(whileTap),
+        visualElement,
```
nvh, line 98: Why shouldn't we use the explicit tap handlers as dependencies here?

packages/framer-motion/src/gestures/use-tap-gesture.ts
```diff
@@ -7,6 +7,14 @@ import { AnimationType } from "../render/utils/types"
 import { isDragActive } from "./drag/utils/lock"
 import { FeatureProps } from "../motion/features/types"
 import { pipe } from "../utils/pipe"
+import { addDomEvent, useDomEvent } from "../events/use-dom-event"
+import { extractEventInfo } from "../events/event-info"
+
+function fireSyntheticPointerEvent(name: string, handler?: any) {
```
nvh, line 13: Can't we type `handler` stronger?

nvh (changes requested):

packages/framer-motion/src/render/html/use-props.ts
```diff
@@ -82,6 +82,10 @@ export function useHTMLProps(
                 : `pan-${props.drag === "x" ? "y" : "x"}`
     }
 
+    if (props.onTap || props.onTapStart) {
+        htmlProps.tabIndex = 0
```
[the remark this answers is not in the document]
mattgperry: A `div` with `tabindex="0"` is the equivalent of, for instance a `button` with no `tabIndex`. It will be focusable in document order: https://developer.mozilla.org/en-US/docs/Web/Accessibility/Keyboard-navigable_JavaScript_widgets

packages/framer-motion/src/gestures/use-tap-gesture.ts
mattgperry: Ah yeah we can

packages/framer-motion/src/gestures/use-tap-gesture.ts
mattgperry: Because we don't use these directly but via `visualElement.getProps()`. It's only their presence that informs whether we create a new event handler.

mattgperry: If we used them directly they would probably change every render.

packages/framer-motion/src/render/html/use-props.ts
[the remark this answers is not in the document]
mattgperry: Also updated to only set if it's undefined

packages/framer-motion/src/render/dom/use-render.ts
```diff
@@ -18,7 +18,7 @@ export function createUseRender(forwardMotionProps = false) {
             : useHTMLProps
 
         const visualProps = useVisualProps(
-            props,
+            props as any,
```
nvh, line 21: 🤔

nvh (approved):

Issue #1930, opened 2023-02-03
[BUG] Reorder does not work with <LazyMotion strict={true}>


james-wallis:
**2. Describe the bug**
Unable to use the Reorder component when using LazyMotion with `strict` (`<LazyMotion strict={true}>`).

```
Error: You have rendered a `motion` component within a `LazyMotion` component. This will break tree shaking. Import and render a `m` component instead.
```

I've followed https://www.framer.com/motion/guide-reduce-bundle-size/ to reduce the Framer Motion bundle size and have changed all my `motion` components to `m`. However, when I switch on `strict` mode, my app is breaking because there is no way to switch `Reorder` to use `m` instead of `motion`. 

Is the method to make it use `m` undocumented or does this require a code change?

**3. IMPORTANT: Provide a CodeSandbox reproduction of the bug**
https://codesandbox.io/p/sandbox/flamboyant-taussig-p2i0pu?file=%2Fpages%2F_app.tsx

A CodeSandbox minimal reproduction will allow us to quickly follow the reproduction steps. **Without one, this bug report won't be accepted.**

**4. Steps to reproduce**

1. Go to the Codesandbox
2. See error
3. You can open `pages/_app.tsx` to see that removing `strict` works.

**5. Expected behavior**

I can tell the `Reorder` component to use `m` instead of `motion` so the app doesn't crash in strict mode.

james-wallis:
I fixed this on my fork by making the change in https://github.com/framer/motion/pull/1931

PR #1931, opened 2023-02-03
feat: add motionComponent prop to Reorder Group and Item components

james-wallis:
Resolves: https://github.com/framer/motion/issues/1930

I've tested this locally and it resolves the linked issue. I'm open to feedback/changes and can add tests/docs once the implementation details are good.

Edit: Changed this to add an `ignoreStrict` prop onto the `motion` component. When `ignoreStrict` is given, it will output a `warn` instead of throwing an error so that users are still aware of what using `motion` inside `LazyMotion` will do.

### DIFF
--- a/packages/framer-motion/src/components/Reorder/Item.tsx
+++ b/packages/framer-motion/src/components/Reorder/Item.tsx
@@ -11,6 +11,7 @@ import {
 import { ReorderContext } from "../../context/ReorderContext"
 import { Box } from "../../projection/geometry/types"
 import { motion } from "../../render/dom/motion"
+import { m } from "../../render/dom/motion-minimal"
 import { HTMLMotionProps } from "../../render/html/types"
 import { useConstant } from "../../utils/use-constant"
 import { useMotionValue } from "../../value/use-motion-value"
@@ -39,6 +40,13 @@ export interface Props<V> {
      * @default true
      */
     layout?: true | "position"
+
+    /**
+     * The motion component to use. Defaults to `motion`.
+     *
+     * @public
+     */
+    motionComponent?: typeof motion | typeof m
 }
 
 function useDefaultMotionValue(value: any, defaultValue: number = 0) {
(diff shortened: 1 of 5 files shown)


packages/framer-motion/src/components/Reorder/Item.tsx
```diff
@@ -53,11 +61,12 @@ export function ReorderItem<V>(
         as = "li",
         onDrag,
         layout = true,
+        motionComponent = motion,
```
mattgperry, line 64: This will fix the error message but the bundle size will still go up

mattgperry:
Hi James, thanks for the PR!

The problem with this approach is it will remove the error message but because the prop can't be resolved until runtime, it will still increase the bundlesize in the way it's warning about.

Perhaps a cleaner approach would be to have a `ignoreStrict` option on `MotionConfig` or `motion` that could be used to disable the check, although we'd still have the same problem of an increased bundlesize. 

Alternatively the `invariant` could become a `warning` - on one hand I think it's reasonable for `Reorder` to expect the drag functionality to be present but on the other I think there's value in allowing it to still be lazy-loaded.

james-wallis:
Hey @mattgperry, yeah, that makes sense!

I want to check my understanding of `LazyMotion` with you if that is alright. 
If I use `LazyMotion` and load asynchronously using `import()` and use `m` in all places in my app apart from on the `/reorder` page (using Next.js if that makes a difference) where `Reorder` is used - I'll still get the reduced bundle size when not on `/reorder`?
- I'm hoping this is the case so I'll work on a change to add the `ignoreStrict` prop to a `motion` component. Think it makes sense to prevent the error when using a built-in component?

Or if my app is going to `Reorder` I should use the `motion` component always?

mattgperry:
@james-wallis Yeah this is the case - `Reorder` will come with `motion` and all the features. But this means when you're not on this page, the features won't be loaded.

mattgperry (approved):
Great PR! Thanks

PR #1947, opened 2023-02-13
Moving optional features from React components to `class`

mattgperry:
Hooks can't be called conditionally, so to run conditional code we currently use conditionally-rendered "renderless" components.

```jsx
function Hover() {
  useHover()
  return null
}

function motion() {
  return (
    <div>
      {isHover ? <Hover /> : null}
    </div>
  )
}
```

This incurs some overhead by involving React. This PR changes the architecture to move from components to custom vanilla classes. In result is a faster architecture that reduces blocking time on our SaaS template by 10% and will improve overall runtime. 

image

Instead, features are now classes with the interface

```
class Feature {
  mount() {}
  unmount() {}
  update() {}
}
```

Where `mount` is called the first `useEffect` where a feature is present and `update` on subsequent effects.

Fixes https://github.com/framer/motion/issues/1949


CHANGELOG.md
```diff
@@ -4,6 +4,14 @@ Framer Motion adheres to [Semantic Versioning](http://semver.org/).
 
 Undocumented APIs should be considered internal and may change without warning.
 
+### Updated
+
+-   Refactored `motion` component features to load without React components.
+
+### Removed
+
+-   Fallback support for `IntersectionObserver`.
```
mattgperry, line 13: We currently run `IntersectionObserver` without polyfill on Framer sites so this is safe for us to do.

packages/framer-motion/src/events/use-dom-event.ts
```diff
@@ -1,15 +1,5 @@
-import { useEffect, RefObject } from "react"
-
-export function addDomEvent(
```
mattgperry, line 3: I split this out into its own file as we no longer use `useDomEvent` but we do export it via `index.js` and use it within Framer. This allows us to code split the two functions.

packages/framer-motion/src/gestures/drag/__tests__/index.test.tsx
```diff
@@ -146,7 +147,7 @@ describe("dragging", () => {
         pointer.end() // + 3 = 14
         await onDragEnd.promise
 
-        expect(count).toBeGreaterThanOrEqual(12)
```
mattgperry, line 149: I've taken a look at why this is happening and the feature is getting unmounted a frame earlier than before. This could be explained by before it was getting cleaned up by the returned Effect in a child component whereas now we clean up features on unmounting via the `ref` prop.

But the thing we're testing for is working as expected and I don't expect this to have any practical ramifications.

packages/framer-motion/src/gestures/types.ts
```diff
@@ -192,10 +189,7 @@ export interface PanHandlers {
      *   - `point`: Relative to the device or page.
      */
 
-    onPanSessionStart?(
-        event: MouseEvent | TouchEvent | PointerEvent,
```
mattgperry, line 196: These were incorrect

packages/framer-motion/src/utils/__tests__/use-instant-transition.test.tsx
```diff
@@ -14,6 +14,7 @@ describe("useInstantTransition", () => {
 
             const Parent = () => {
                 const [state, setState] = React.useState(false)
+                console.log({ state })
```
shuangq, line 17: Is this console log intended?

shuangq (approved):

packages/framer-motion/src/motion/features/types.ts
```diff
@@ -3,65 +3,55 @@ import { MotionProps } from "../types"
...
 import { VisualElement } from "../../render/VisualElement"
 import { CreateVisualElement } from "../../render/types"
+import type { Feature } from "./Feature"
+import { MeasureLayout } from "./layout/MeasureLayout"
 
-/**
- * @public
- */
-export interface FeatureProps<T = unknown> extends MotionProps {
-    visualElement: VisualElement<T>
```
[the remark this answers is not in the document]
mattgperry: I don't think any of this is public/documented

CHANGELOG.md
```diff
@@ -4,6 +4,20 @@ Framer Motion adheres to [Semantic Versioning](http://semver.org/).
...
 
+## [9.0.3] 2023-02-14
+
+### Updated
+
+-   Refactored `motion` component features to load without React components.
+
+### Removed
+
+-   Fallback behavior for missing `IntersectionObserver`.
```
nvh, line 15: Does this match our browser support matrix?

packages/framer-motion-3d/src/render/gestures/use-hover.ts
```diff
@@ -21,11 +21,11 @@ export function useHover(
     return {
         onPointerOver: (event: ThreeEvent<any>) => {
             visualElement.animationState?.setActive(AnimationType.Hover, true)
-            onPointerOver?.(event)
+            onPointerOver && onPointerOver(event)
         },
         onPointerOut: (event: ThreeEvent<any>) => {
             visualElement.animationState?.setActive(AnimationType.Hover, false)
-            onPointerOut?.(event)
+            onPointerOut && onPointerOut(event)
```
nvh, line 28: Does this matter that much? I like the readability of the `?.` operator.

CHANGELOG.md
mattgperry: I only know we already don't provide a fallback or polyfill on Framer sites

packages/framer-motion-3d/src/render/gestures/use-hover.ts
mattgperry: Yes - it leads to substantial size savings across the package

mattgperry: If you look at the output of this operator you can see why - it is quite verbosely transpiled but also in a way that doesn't allow for good minification or gzipped (lots of unique code)

Screenshot 2023-02-15 at 11 00 31

mattgperry: Also chained ones can get wild

```
const a = object?.prop?.method()

const a = (_a = object === null || object === void 0 ? void 0 : object.prop) === null || _a === void 0 ? void 0 : _a.method()
```

PR #1950, opened 2023-02-14
Adding clipPath, filter and transform to waapi values

mattgperry:


packages/framer-motion/src/animation/utils/default-transitions.ts
```diff
@@ -48,9 +38,11 @@ export const getDefaultTransition = (
 ): Partial<AnimationOptions> => {
     if (keyframes.length > 2) {
         return keyframesTransition
-    } else {
-        const factory =
-            defaultTransitions[valueKey] || defaultTransitions.default
-        return factory(keyframes[1])
+    } else if (transformProps.has(valueKey)) {
```
mattgperry, line 41: `opacity` `backgroundColor` and `color` now use this slightly smoother default easing curve

nvh: So is this a breaking change? (I think we can treat it as not?)

nvh (approved):

packages/framer-motion/src/animation/utils/default-transitions.ts
mattgperry: Yeah technically this does change the defaults but personally I don't think anyone is going to notice this difference

hemlok (approved):
:heavy-breathing:

PR #1956, opened 2023-02-16
25% speed increase for `transform` string creation

mattgperry:


packages/framer-motion/src/render/html/utils/build-transform.ts
```diff
@@ -37,12 +30,13 @@ export function buildTransform(
     // The transform string we're going to build into.
     let transformString = ""
 
-    // Transform keys into their default order - this will determine the output order.
```
mattgperry, line 40: This is the meat and potatoes of the PR

PR #1964, opened 2023-02-21
Remove external dependencies

mattgperry:
This PR removes Motion One and `hey-listen` as external deps


packages/framer-motion/package.json
```diff
@@ -39,7 +39,7 @@
         "clean": "rm -rf types dist lib",
         "test": "yarn test-server && yarn test-client",
         "test-ci": "yarn test",
-        "test-client": "jest --config jest.config.json --max-workers=2",
+        "test-client": "jest --config jest.config.json --max-workers=2 src/animation/__tests__/css-variables.test.tsx",
```
shuangq, line 42: Was this for debugging? should we remove it?

shuangq (approved):

packages/framer-motion/package.json
mattgperry: thanks good catch

PR #1969, opened 2023-02-23
Replace `type enum` with `const enum`

mattgperry:
Closes https://github.com/framer/motion/pull/1968


packages/framer-motion/src/render/dom/utils/unit-conversion.ts
```diff
@@ -30,15 +30,6 @@ const hasPositionalKey = (target: TargetWithKeyframes) => {
...
 
-export enum BoundingBoxDimension {
-    width = "width",
-    height = "height",
-    left = "left",
-    right = "right",
-    top = "top",
-    bottom = "bottom",
-}
-
```
hemlok, line 41: Was this unused?

mattgperry: yeah and actually bundled too

hemlok (approved):

PR #1978, opened 2023-02-24
Animate `background-color` via WAAPI

mattgperry:
Always generate keyframes for `background-color` to ensure we use our linear RGB interpolation rather than the browsers'

### DIFF
--- a/packages/framer-motion/src/animation/waapi/create-accelerated-animation.ts
+++ b/packages/framer-motion/src/animation/waapi/create-accelerated-animation.ts
@@ -27,6 +27,14 @@ const acceleratedValues = new Set<string>([
  */
 const sampleDelta = 10 //ms
 
+const requiresPregeneratedKeyframes = (
+    valueName: string,
+    options: AnimationOptions
+) =>
+    options.type === "spring" ||
+    valueName === "backgroundColor" ||
+    !isWaapiSupportedEasing(options.ease)
+
 export function createAcceleratedAnimation(
     value: MotionValue,
     valueName: string,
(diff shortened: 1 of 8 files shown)


packages/framer-motion/src/animation/waapi/create-accelerated-animation.ts
```diff
@@ -45,14 +46,16 @@ export function createAcceleratedAnimation(
     /**
      * If this animation needs pre-generated keyframes then generate.
      */
-    if (options.type === "spring" || !isWaapiSupportedEasing(options.ease)) {
-        /**
-         * If we need to pre-generate keyframes and repeat is infinite then
-         * early return as this will lock the thread.
-         */
-        if (options.repeat === Infinity) return
```
shuangq, line 53: Was this fixed in a different PR?

packages/framer-motion/src/animation/waapi/create-accelerated-animation.ts
```diff
@@ -45,14 +46,16 @@ export function createAcceleratedAnimation(
...
-
-        const sampleAnimation = animateValue({ ...options, elapsed: 0 })
+    if (
+        options.type === "spring" ||
+        valueName === "backgroundColor" ||
+        !isWaapiSupportedEasing(options.ease)
```
shuangq, line 52: nit: at this point it might be more readable to create a helper `needsPregeneratedKeyframes`?

shuangq (approved):

packages/framer-motion/src/animation/waapi/create-accelerated-animation.ts
mattgperry: Yes and no. `repeat = 0` would have fixed it in this PR. But we did add a 20 second limit a long time ago that did fix this infinite loop. However the true fix in this case is `repeat = 0` - we want to pre-generate the unrepeated keyframes and only repeat them when we play them via WAAPI.

shuangq: I see! Might worth a comment

PR #1982, opened 2023-02-24
Reducing time spent on keyframe pregeneration by ~75%

mattgperry:
This PR removes an unnecessary nested loop from keyframe pregeneration, speeding it up by an amount proportional to the duration of the animation (in tests with a spring, 75%)

### DIFF
--- a/packages/framer-motion/src/frameloop/types.ts
+++ b/packages/framer-motion/src/frameloop/types.ts
@@ -17,13 +17,7 @@ export interface Step {
     process: (frame: FrameData) => void
 }
 
-export type StepId =
-    | "read"
-    | "update"
-    | "postUpdate"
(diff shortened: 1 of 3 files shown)


packages/framer-motion/src/frameloop/types.ts
```diff
@@ -17,13 +17,7 @@ export interface Step {
...
 }
 
-export type StepId =
-    | "read"
-    | "update"
-    | "postUpdate"
-    | "preRender"
-    | "render"
-    | "postRender"
+export type StepId = "read" | "update" | "preRender" | "render" | "postRender"
```
mattgperry, line 20: This removed step unimplemented

shuangq (approved):

PR #1984, opened 2023-02-27
Vanilla entrypoint

mattgperry:
This PR adds an entrypoint for using Framer Motion's low-level APIs without React

```
import { animate } from "framer-motion/dom"
```


packages/framer-motion/rollup.size.config.js
```diff
@@ -0,0 +1,76 @@
+import resolve from "@rollup/plugin-node-resolve"
```
mattgperry, line 1: This splits the distributed bundles from the bundles we generate for size. Makes `make build` faster when not measuring

o-alexandrov:
Could you please share your plans on documenting vanilla JS usage of framer-motion?
Personally, I’d love to see layout animation working for Vue.js, as it’d bring a lot of new users to framer-motion.

o-alexandrov:
@mattgperry thank you for the reply. I'll hope it'll turn out to be different over time, as you know yourself that if you figure out how not to rely on React, you'd improve performance dramatically. I believe this PR along with #1947 are a positive tendency increasing probability of getting a framework-agnostic `framer-motion`'s, thank you very much for the library

mattgperry:
The React touchpoints are more about scheduling. We're barely threading the needle on having the right lifecycles to pull off layout animations, but the performance overhead is almost exclusively layout measurement and projection calculations.

hemlok (commented):
👏

PR #1993, opened 2023-03-03
Make `animate()` resolve for time

mattgperry:
This PR replaces the `animate` function from the previous Popmotion approach which consumes time delta per frame (helpfully illustrated by this):

Screenshot 2023-03-08 at 08 57 37

Into the Motion One stateless approach where the timestamp itself is passed.

This allows for:
- **Playback functions.** Currently we only return `stop` but for arbitrary seeking, pause, play of springs/inertia this stateless approach works better.
- **API parity with WAAPI.** In a future PR this will allow `animate` to easily return a set of playback controls that wraps many JS and/or WAAPI animation controls.
- **Quicker seeking.** Currently we have to simulate a low-res version of animations when sampling to ensure repeats are incorporated correctly whereas now this can be instant, which leads to faster optimised appear animation handoff and faster WAAPI interruption.


packages/framer-motion/package.json
```diff
@@ -103,10 +103,6 @@
             "path": "./dist/size-rollup-animate.js",
             "maxSize": "10 kB"
         },
-        {
```
mattgperry, line 106: Dupe of previous size check

packages/framer-motion/src/animation/index.ts
```diff
@@ -84,12 +84,6 @@ export const createMotionValueAnimation = (
              * or this is simply defined as an instant transition, return an instant transition.
              */
             return createInstantAnimation(options)
-        } else if (valueTransition.type === "inertia") {
```
mattgperry, line 87: `animate()` now handles inertia animations

packages/framer-motion/src/animation/js/index.ts
```diff
@@ -0,0 +1,264 @@
...
+    ...options
+}: AnimationOptions<V>): MainThreadAnimationControls<V> {
+    let animationDriver: DriverControls
+
+    const generatorFactory = types[type] || keyframesGeneratorFactory
+
+    /**
+     * If this isn't the keyframes generator and we've been provided
+     * strings as keyframes, we need to interpolate these.
+     * TODO: Support velocity for units and complex value types/
```
mattgperry, line 62: This is a "in life" todo not a "this PR" todo

packages/framer-motion/src/animation/animate.ts
```diff
@@ -61,12 +52,10 @@ export type AnimationOptions<V> = (Tween | Spring) &
...
-    transition: AnimationOptions<V> = {}
+    transition: Transition & AnimationPlaybackLifecycles<V> = {}
 ): AnimationPlaybackControls {
     const value = isMotionValue(from) ? from : motionValue(from)
     value.start(createMotionValueAnimation("", value, to as any, transition))
 
-    return {
-        stop: () => value.stop(),
-    }
+    return value.animation!
```
nvh, line 60: Is the `!` necessary?

packages/framer-motion/src/animation/legacy-popmotion/__tests__/inertia.test.ts
```diff
@@ -1,203 +0,0 @@
...
-        ...options,
-        onUpdate: (v) => output.push(Math.round(v)),
-        onComplete: () => {
-            expect(output).toEqual(expectation)
-            resolve()
-        },
-    })
-}
-
-describe("inertia", () => {
```
nvh, line 24: Did these tests move somewhere else?

packages/framer-motion/src/animation/js/__tests__/animate.test.ts
```diff
@@ -771,11 +808,51 @@ describe("animate", () => {
...
+            autoplay: false,
+            ease: noop,
+        })
+
+        expect(animation.sample(0).value).toEqual(0)
+        expect(animation.sample(250).value).toEqual(103.65507873893438)
+        expect(animation.sample(4100).value).toEqual(96.10257237444083)
+    })
+
+    test.only("Correctly sets and gets currentTime", async () => {
```
nvh, line 835:
```suggestion
    test("Correctly sets and gets currentTime", async () => {
```

packages/framer-motion/src/animation/animate.ts
nvh: Is there a way we can get to the animation in a "safer" way (my first thought was returning it from `start`, but that already returns a promise), so we have more guarantee than "yeah we really started the animation, I'm sure"

packages/framer-motion/src/animation/generators/__tests__/utils.ts
```diff
@@ -0,0 +1,35 @@
...
+
+    /**
+     * If generating an animation that didn't actually move,
+     * generate a second keyframe so we have an origin and target.
+     */
+    if (keyframes.length === 1) keyframes.push(state.value)
+
+    return {
+        keyframes,
+        duration: duration / 1000,
```
nvh, line 33: are there times when we're only interested in the duration? Because we could skip the modifying of the array, leading to a potential performance win.

packages/framer-motion/src/motion/types.ts
```diff
@@ -73,6 +73,19 @@ export interface CustomStyles {
     image?: string
 }
 
+type CSSPropertiesWithoutTransitionOrSingleTransforms = Omit<
+    CSSProperties,
+    "transition" | "rotate" | "scale" | "perspective"
+>
```
nvh, line 79: One could argue that `perspective` is not really a single transfrom, but I don't know a better name.

packages/framer-motion/src/motion/__tests__/variant.test.tsx
```diff
@@ -205,7 +205,7 @@ describe("animate prop as variant", () => {
         return expect(promise).resolves.toBe("rgba(190, 60, 60, 1)")
     })
 
-    test("respects orchestration props in transition prop", async () => {
+    test.only("respects orchestration props in transition prop", async () => {
```
nvh, line 208:
```suggestion
    test("respects orchestration props in transition prop", async () => {
```

packages/framer-motion/src/animation/waapi/create-accelerated-animation.ts
```diff
@@ -69,7 +72,8 @@ export function createAcceleratedAnimation(
          */
         let t = 0
         while (!state.done && t < 20000) {
-            state = sampleAnimation.sample(t, true)
+            // TODO Reinstate sample
+            state = sampleAnimation.sample(t)
```
nvh, line 76: I don't understand this TODO, could you make it a little bit more elaborate?

packages/framer-motion/src/animation/generators/spring/index.ts
```diff
@@ -0,0 +1,176 @@
...
+            }
+
+            state.value = state.done ? target : current
+
+            return state
+        },
+    }
+}
+
+const zero = (_t: number) => 0
```
nvh, line 176: Why define this as a separate const? Performance?

packages/framer-motion/src/animation/index.ts
```diff
@@ -62,7 +62,7 @@ export const createMotionValueAnimation = (
             keyframes,
             velocity: value.getVelocity(),
             ...valueTransition,
-            elapsed,
+            delay: -elapsed,
```
nvh, line 65: I don't really get this one.

packages/framer-motion/src/animation/js/index.ts
```diff
@@ -0,0 +1,261 @@
...
+    tween: keyframesGeneratorFactory,
+    keyframes: keyframesGeneratorFactory,
+    spring,
+}
+
+function calculateDuration(generator: KeyframeGenerator<unknown>) {
+    let duration = 0
+    const timeStep = 50
+    let state = generator.next(duration)
+    while (!state.done && duration < 20000) {
```
nvh, line 28: What is this constant? Can't we go higher than 20s? Maybe we should extract it into a const, and can we write it like `20_000` in typescript?

nvh (changes requested):
Left some questions, and there are still some `.only(` tests sprinkled here and there.

packages/framer-motion/src/animation/legacy-popmotion/__tests__/inertia.test.ts
mattgperry: Yeah `inertia()` used to be a function that ran its own animations so the tests are different (taken from Motion One) and now in `animation/generators/__tests__/inertia.ts`

packages/framer-motion/src/animation/create-instant-animation.ts
```diff
@@ -1,16 +1,28 @@
...
+            stop: () => {},
+            currentTime: 0,
+        }
     }
 
-    return elapsed ? { stop: delay(setValue, -elapsed) } : setValue()
+    return delayBy
+        ? animateValue({
+              keyframes: [0, 1],
+              duration: delayBy,
```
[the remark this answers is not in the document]
mattgperry: In the old code `stop: () => {}` was the only playback control. These will be expanded - by returning `animateValue` this function will remain fungible with the JS/WAAPI animations.

packages/framer-motion/src/animation/generators/__tests__/utils.ts
mattgperry: This one is just a utility function for tests and I don't think it often only returns `duration` however you've reminded me that we do have two similar functions that we might be able to unify and I'll keep this in mind if it's only being called for duration calculations

packages/framer-motion/src/animation/waapi/create-accelerated-animation.ts
mattgperry: Actually I can remove - this was from before I redid `sample` to take advantage of the stateless animation

packages/framer-motion/src/animation/generators/utils/velocity-sample-duration.ts
```diff
@@ -0,0 +1 @@
+export const velocitySampleDuration = 5 // ms
```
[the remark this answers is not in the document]
mattgperry: Ah yeah it was being used in multiple files but not now, I can co-locate

packages/framer-motion/src/animation/generators/spring/index.ts
mattgperry: Basically yeah - this file is identical to the old `spring` and this is just a holdover. But I've just removed this and replaced with `let resolveSpring: (v: number) => number`

packages/framer-motion/src/animation/js/index.ts
mattgperry: (But will abstract to constant)

packages/framer-motion/src/animation/animate.ts
mattgperry: The idea is in a later iteration for `animate()` to be able to return `GroupPlaybackControls` which can control multiple animations so I could implement that. It shouldn't care if there are none or many underlying animations.

mattgperry: Implemented in https://github.com/framer/motion/pull/1993/commits/c926c957c23b28e37c61369f95ce6adffcc1c0f4

packages/framer-motion/src/motion/types.ts
mattgperry: Removed

packages/framer-motion/src/animation/GroupPlaybackControls.ts
```diff
@@ -0,0 +1,29 @@
+import { AnimationPlaybackControls } from "./types"
+
+export class GroupPlaybackControls implements AnimationPlaybackControls {
+    animations: AnimationPlaybackControls[]
+
+    constructor(animations: Array<AnimationPlaybackControls | undefined>) {
+        this.animations = animations.filter(
+            Boolean
+        ) as AnimationPlaybackControls[]
```
nvh, line 9: I'm still so disappointed that typescript can't figure this out 😞 
Is there no way to go without the cast?

packages/framer-motion/src/animation/GroupPlaybackControls.ts
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    get currentTime() {
+        return this.animations[0].currentTime
+    }
```
nvh, line 14: Shouldn't it be the `max()` of currentTimes or something?

packages/framer-motion/src/animation/animate.ts
```diff
@@ -61,12 +46,10 @@ export type AnimationOptions<V> = (Tween | Spring) &
...
-    transition: AnimationOptions<V> = {}
+    transition: Transition & AnimationPlaybackLifecycles<V> = {}
 ): AnimationPlaybackControls {
     const value = isMotionValue(from) ? from : motionValue(from)
     value.start(createMotionValueAnimation("", value, to as any, transition))
 
-    return {
-        stop: () => value.stop(),
-    }
+    return value.animation || new GroupPlaybackControls([value.animation])
```
nvh, line 54: Why would we add the animation if it is falsy? I'd say we either always return a GroupPlaybackControls, or return an empty one when the animation doesn't exist.

packages/framer-motion/src/frameloop/index.ts
```diff
@@ -46,7 +45,7 @@ const processFrame = (timestamp: number) => {
     runNextFrame = false
 
     frameData.delta = useDefaultElapsed
-        ? defaultTimestep
+        ? 16.666
```
nvh, line 48: Any reason for not making this a constant? I'd say that 1000/60 is clearer than this.

nvh: (Don't feel strongly though, as it is kinda common knowledge)

nvh (approved):
Looks good now, still have my doubts about the GroupedAnimationControl, but as it is obviously in flux I'll leave it up to you to fix now or later.

packages/framer-motion/src/animation/GroupPlaybackControls.ts
mattgperry: That would be correct for `duration`, as I realised after implementing `max` originally too. In theory all these animations should be in sync, some could be interrupted at which point this is a bit of a fudge, we probably want to filter out animations that aren't running or paused and take the duration from the first one of those. But I do want to take a second look at `currentTime` before making an official API, this is just for internal use and it doesn't currently represent more than one animation, but I'll add a comment

packages/framer-motion/src/animation/GroupPlaybackControls.ts
mattgperry: It actually didn't even work with a more explicit function either (that checked specficially for undefined)

packages/framer-motion/src/frameloop/index.ts
mattgperry: Works for me - done

Issue #1997, opened 2023-03-07
[BUG] Build fails with error caused by incorrect type declaration


JethroNatividad:
## Problem
When building my Next.js application with Chakra UI, I encountered a build error caused by an incorrect type declaration in the Framer Motion library. Specifically, the typeof Feature<unknown> type caused the error "Type error: ';' expected" on line 1212 of the index.d.ts file.

```shell
./node_modules/.pnpm/framer-motion@10.1.0_biqbaboplfbrettd7655fr4n2y/node_modules/framer-motion/dist/index.d.ts:1212:28
Type error: ';' expected.

  1210 | declare type HydratedFeatureDefinition = {
  1211 |     isEnabled: (props: MotionProps) => boolean;
> 1212 |     Feature: typeof Feature<unknown>;
       |                            ^
  1213 |     ProjectionNode?: any;
  1214 |     MeasureLayout?: typeof MeasureLayout;
  1215 | };
 ELIFECYCLE  Command failed with exit code 1.
```
I have tried building with pnpm v7.26.2, npm v8.19.3, and yarn v1.22.1, but all attempts resulted in the same error.
### I am using the following dependencies:
next: 13.2.1
react: 18.2.0
framer-motion: 10.1.0
@chakra-ui/react: 2.5.1
typescript: 4.6.4

## How to reproduce
The error only occurs during the build process and not during development, so it may be difficult to reproduce. However, the error can be seen by building a project that uses the dependencies above.

## Solution
After investigating the issue, I found that the problem was with the type definition for `HydratedFeatureDefinition`. Specifically, the `Feature` property was defined as `typeof Feature<unknown>`, which was causing a type error during the build process.

To fix the issue, I updated the type definition for `HydratedFeatureDefinition` to use a generic `FeatureClass` interface, like so:
```ts
interface FeatureClass<Props> {
    new (props: Props): Feature<Props>;
}
```
And replaced the `typeof Feature<unknown>` in `HydratedFeatureDefinition` to use the `FeatureClass` interface
https://github.com/framer/motion/blob/main/packages/framer-motion/src/motion/features/types.ts#L11
```diff
declare type HydratedFeatureDefinition = {
    isEnabled: (props: MotionProps) => boolean;
+   Feature: FeatureClass<unknown>;
-   Feature: typeof Feature<unknown>;
    ProjectionNode?: any;
    MeasureLayout?: typeof MeasureLayout;
};
```

This change resolved the build error and allowed my project to build successfully.

[...]

PR #1998, opened 2023-03-07
Fix hydrated feature definition type

JethroNatividad:
#1997


mattgperry:
Great contribution, thanks!

PR #2007, opened 2023-03-13
Adding `animation.then`

mattgperry:
This PR adds a method of awaiting animations.

```javascript
const animation = animate()

animation.then(() => {})

// or
await animation
```

### DIFF
--- a/packages/framer-motion/src/animation/js/index.ts
+++ b/packages/framer-motion/src/animation/js/index.ts
@@ -64,6 +64,11 @@ export function animateValue<V = number>({
     let resolveFinishedPromise: VoidFunction
     let currentFinishedPromise: Promise<void>
 
+    /**
+     * Create a new finished Promise every time we entered the
+     * finished state and resolve the old Promise. This is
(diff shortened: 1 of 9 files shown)


packages/framer-motion/src/animation/js/index.ts
```diff
@@ -61,6 +61,22 @@ export function animateValue<V = number>({
...
+     * finished state and resolve the old Promise. This is
+     * WAAPI-compatible behaviour.
+     */
+    const updateFinishedPromise = () => {
+        currentFinishedPromise = new Promise((resolve) => {
+            resolveFinishedPromise = resolve
+        })
+    }
+
+    updateFinishedPromise()
```
hemlok, line 78: I guess this saves a few bytes because you don't have to initialise `currentFinishedPromise` with the same code in `updateFinishedPromise` but it confused me for a sec.

Maybe just a lil comment above saying `// Initialize the finished promise`

hemlok (approved):

PR #2016, opened 2023-03-15
Add `.complete()` method to `animate()`

mattgperry:
This PR adds a `.complete()` method to `animate()`.

When called, animations will skip to the end.

```javascript
animation.complete()
```

### DIFF
--- a/packages/framer-motion/src/animation/animators/js/index.ts
+++ b/packages/framer-motion/src/animation/animators/js/index.ts
@@ -312,6 +312,10 @@ export function animateValue<V = number>({
             onStop && onStop()
             cancel()
         },
+        complete: () => {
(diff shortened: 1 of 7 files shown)


packages/framer-motion/src/animation/animators/js/index.ts
```diff
@@ -233,6 +227,12 @@ export function animateValue<V = number>({
             holdTime === null &&
             (playState === "finished" || (playState === "running" && done))
 
+        if (onUpdate) {
```
mattgperry, line 230: By firing external listeners before we've determined whether the animation has finished, we can inadvertently end animations a frame early.

FrancescaB33 (approved):

PR #2018, opened 2023-03-16
Adding `.speed` property to `animate()`

mattgperry:
This PR adds a `.speed` property to `animate()`.

This is a factor that affects the playback speed of the animation.

```javascript
// Half speed
animation.speed = 1

// Normal speed
animation.speed = 1

// Double speed
animation.speed = 2

// Reverse
animation.speed = -1
```


packages/framer-motion/src/animation/types.ts
```diff
@@ -26,6 +26,7 @@ export type ElementOrSelector =
  */
 export interface AnimationPlaybackControls {
     time: number
+    speed: number
```
nvh, line 29: Speed is a bit undescriptive as to what the unit of it is, what about `playbackRate`?

nvh: That would match WAAPI terminology: https://developer.mozilla.org/en-US/docs/Web/API/Animation/playbackRate

packages/framer-motion/src/animation/GroupPlaybackControls.ts
```diff
@@ -16,22 +18,38 @@ export class GroupPlaybackControls implements AnimationPlaybackControls {
     /**
      * TODO: Filter out cancelled or stopped animations before returning
      */
+    private getAll(propName: PropNames) {
+        return this.animations[0][propName]
```
nvh, line 22: How do make sure all animations have the same value per prop?

packages/framer-motion/src/animation/__tests__/GroupPlaybackControls.test.ts
```diff
@@ -182,4 +183,20 @@ describe("GroupPlaybackControls", () => {
...
+        const a = createTestAnimationControls({
+            speed: 2,
+        })
+        const b = createTestAnimationControls({
+            speed: 2,
+        })
+
+        const controls = new GroupPlaybackControls([a, b])
+
+        controls.speed = -1
```
nvh, line 197: This is only testing the setting, no the getting, I think that would be better in a separate tests.

packages/framer-motion/src/animation/animators/waapi/create-accelerated-animation.ts
```diff
@@ -166,6 +166,12 @@ export function createAcceleratedAnimation(
         set time(newTime: number) {
             animation.currentTime = secondsToMilliseconds(newTime)
         },
+        get speed() {
+            return animation.playbackRate
+        },
+        set speed(newSpeed: number) {
+            animation.playbackRate = newSpeed
+        },
```
nvh, line 174: Oh haha, missed this part. Why not stay with this naming?

packages/framer-motion/src/animation/animators/js/index.ts
```diff
@@ -297,12 +308,21 @@ export function animateValue<V = number>({
...
+                startTime = animationDriver.now() - newTime / speed
             }
         },
+        get speed() {
+            return speed
+        },
+        set speed(newSpeed: number) {
+            if (newSpeed === speed || !animationDriver) return
+            speed = newSpeed
+            startTime = animationDriver.now() - time / speed
```
nvh, line 323: Why not use `set time()` here?

nvh: We also should check for speed `0` here, otherwise we'll get an error.

packages/framer-motion/src/animation/types.ts
mattgperry: IMO `speed`/`scale`/`rate`/`playback-` are all roughly equivalent, `playbackRate` is probably the most descriptive but it is also kind of ugly, and when things are fairly equal I would rather go with APIs that are aesthetically pleasing vs verbose/correct.

```javascript
await animation.finished 🚫
await animation ✅

animation.currentTime = 2 🚫
animation.time = 2 ✅

animation.playbackRate = 0.5 🚫
animation.speed = 0.5 ✅
```

packages/framer-motion/src/animation/animators/js/index.ts
mattgperry: Ah yeah great point also that function does check for `0`

packages/framer-motion/src/animation/types.ts
nvh: Ok, I don't really feel strong about it, so if you're certain about this API, go for it! 👌🏻

packages/framer-motion/src/animation/animators/js/index.ts
```diff
@@ -297,12 +308,21 @@ export function animateValue<V = number>({
...
+                startTime = animationDriver.now() - newTime / speed
             }
         },
+        get speed() {
+            return speed
+        },
+        set speed(newSpeed: number) {
+            if (newSpeed === speed || !animationDriver) return
+            speed = newSpeed
+            controls.time = millisecondsToSeconds(time)
```
nvh, line 323: It might look a bit weird, but we could do 
```suggestion
            controls.time = controls.time
```

nvh (approved):

PR #2033, opened 2023-03-20
`stagger()`

mattgperry:
This PR adds a `stagger()` function for use with `animate()`.

When multiple elements are provided/selected, `stagger` will increase the `delay` for each.

```javascript
animate("div", { opacity: 1 }, { delay: stagger(0.1) })
```

Additional options can be provided.

### `ease`

Redistributes stagger duration based on easing.

```javascript
stagger(0.1, { ease: "easeIn" })
```

### `start`

A base delay to start staggering from - naming suggestions welcome. 

```javascript
stagger(0.1, { start: 1 })
```

### `origin`

Define an element in the set to stagger out from. 

```
from: "first" | "last" | "center" | number
```

```javascript
stagger(0.1, { from: "last" })
```

### DIFF
--- a/packages/framer-motion/src/animation/animate.ts
+++ b/packages/framer-motion/src/animation/animate.ts
@@ -17,8 +17,7 @@ import { GenericKeyframesTarget } from "../types"
 import { createVisualElement } from "./utils/create-visual-element"
 import { animateSingleValue } from "./interfaces/single-value"
 
-export interface DOMAnimationOptions
-    extends Omit<AnimateOptions<any>, "delay"> {
+export interface DOMAnimationOptions extends Omit<AnimateOptions, "delay"> {
     delay?: number | DynamicOption<number>
 }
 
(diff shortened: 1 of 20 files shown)


packages/framer-motion/src/animation/animate.ts
```diff
@@ -89,27 +106,31 @@ export const createScopedAnimate = (scope?: AnimationScope) => {
...
+        options: AnimateOptions<V> | DOMAnimationOptions = {}
     ): AnimationPlaybackControls {
         let animation: AnimationPlaybackControls
 
         if (isDOMKeyframes(keyframes)) {
             animation = animateElements(
                 valueOrElement as ElementOrSelector,
                 keyframes,
-                options,
+                options as DOMAnimationOptions,
```
mattgperry, line 122: Would prefer no cast but likewise runtime checks here are pointless beyond `isDOMKeyframes`

PR #2045, opened 2023-03-23
Adding `.duration` to `animate()`

mattgperry:
This PR adds `.duration` as a read-only property for animations.

```javascript
const animation = animate(0, 100)

animation.duration // .3
```

The `duration` is the duration of time calculated for the active part of the animation without delay or repeat, which may be added as an extra prop at a later date.

### DIFF
--- a/packages/framer-motion/src/animation/GroupPlaybackControls.ts
+++ b/packages/framer-motion/src/animation/GroupPlaybackControls.ts
@@ -1,6 +1,6 @@
 import { AnimationPlaybackControls } from "./types"
 
-type PropNames = "time" | "speed"
+type PropNames = "time" | "speed" | "duration"
 
 export class GroupPlaybackControls implements AnimationPlaybackControls {
     animations: AnimationPlaybackControls[]
@@ -42,6 +42,14 @@ export class GroupPlaybackControls implements AnimationPlaybackControls {
 
     set speed(speed: number) {
         this.setAll("speed", speed)
+    }
+
+    get duration() {
+        let max = 0
+        for (let i = 0; i < this.animations.length; i++) {
+            max = Math.max(max, this.animations[i].duration)
+        }
+        return max
     }
 
     private runAll(
--- a/packages/framer-motion/src/animation/animators/js/__tests__/animate.test.ts
+++ b/packages/framer-motion/src/animation/animators/js/__tests__/animate.test.ts
@@ -1242,4 +1242,56 @@ describe("animate", () => {
 
         await animation
     })
+
+    test("Correctly returns duration", async () => {
+        expect(
+            animateValue({
+                keyframes: [0, 100],
+                duration: 1000,
+            }).duration
+        ).toEqual(1)
+    })
+
+    test("Correctly returns duration when delay is defined", () => {
+        expect(
+            animateValue({
+                keyframes: [0, 100],
+                delay: 1000,
+                duration: 1000,
+            }).duration
+        ).toEqual(1)
+    })
+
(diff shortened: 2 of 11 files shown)


packages/framer-motion/src/animation/GroupPlaybackControls.ts
```diff
@@ -44,6 +44,14 @@ export class GroupPlaybackControls implements AnimationPlaybackControls {
         this.setAll("speed", speed)
     }
 
+    get duration() {
```
nvh, line 47:
```suggestion
    /* The duration is the duration of time calculated for the active part of the animation without delay or repeat, which may be added as an extra prop at a later date. */
    get duration() {
```

packages/framer-motion/src/animation/animators/js/__tests__/animate.test.ts
```diff
@@ -1242,4 +1242,56 @@ describe("animate", () => {
...
+            animateValue({
+                keyframes: [0, 100],
+                delay: 1000,
+                duration: 1000,
+                type: "spring",
+            }).duration
+        ).toEqual(1)
+    })
+
+    test("Correctly returns duration when animation is dynamic spring", () => {
```
nvh, line 1287: What happens with an undamped spring?

packages/framer-motion/src/animation/animators/js/index.ts
```diff
@@ -158,7 +158,9 @@ export function animateValue<V = number>({
         }
 
         // Rebase on delay
-        time = Math.max(time - delay, 0)
+        const timeWithDelay = time - delay
```
nvh, line 161: If it is `time - delay`, I would expect this constant to be called `timeWithoutDelay`

packages/framer-motion/src/animation/animators/js/index.ts
```diff
@@ -323,6 +332,14 @@ export function animateValue<V = number>({
                 startTime = animationDriver.now() - newTime / speed
             }
         },
+        get duration() {
```
nvh, line 335: Ah wait, maybe the comment needs to go with this getter.

packages/framer-motion/src/animation/animators/js/index.ts
nvh: Shouldn't `time` be called `currentTime`?

packages/framer-motion/src/animation/interfaces/motion-value.ts
```diff
@@ -110,6 +97,19 @@ export const animateMotionValue = (
...
+            instantAnimationState.current ||
+            valueTransition.type === false
+        ) {
+            /**
+             * If we can't animate this value, or the global instant animation flag is set,
+             * or this is simply defined as an instant transition, return an instant transition.
+             */
+            return createInstantAnimation(options)
+        }
+
```
nvh, line 112: Why was this bit moved?

nvh (commented):
Left a bunch of comments and questions, nothing really pressing

packages/framer-motion/src/animation/GroupPlaybackControls.ts
mattgperry: I'll put a comment in the underlying types so it applies to all implementations

packages/framer-motion/src/animation/animators/js/__tests__/animate.test.ts
mattgperry: I think for now I'm going to just consider anything over 20 seconds to be `Infinity`

nvh (approved):

PR #2047, opened 2023-03-23
Resize MotionCanvas when container resizes (Fixes #1444)

### LINKED ISSUE #1444: [BUG] framer-motion-3d MotionCanvas does not resize/update with parent container

**2. Describe the bug**

When resizing the element containing the canvas, the canvas should resize as well.

In this r3f example, the element stays centered when resizing the iframe:
https://codesandbox.io/s/react-spring-animations-6hi1y

In this framer-motion-3d example, the element retains its initial coordinates:
https://codesandbox.io/s/framer-motion-3d-variants-9173p

**4. Steps to reproduce**

1. Load a page that contains a MotionCanvas that takes up the full window size.
2. Resize the window
3. Notice that the canvas does not resize to fit the new window size.

**5. Expected behavior**

The canvas should resize with its parent and the scene should update to reflect the new size and aspect ratios.

**6. Video or screenshots**

image

sdacunha:
I was facing the issue in #1444 and attempted to fix it locally via patch-package. It seems to correct the behaviour described in the issue with regards to updating the canvas size as the parent resizes, and utilizes react-use-measure as does the react-three implementation.

I did not find any relevant MotionCanvas tests I could use as a starting point, so I did not any tests for this case, but given it seems like a minor fix, that might be okay. Happy to add it if someone can point me in the right direction / contribute to this PR.

### DIFF
--- a/packages/framer-motion-3d/package.json
+++ b/packages/framer-motion-3d/package.json
@@ -47,7 +47,8 @@
     },
     "dependencies": {
         "framer-motion": "^10.8.5",
-        "react-merge-refs": "^2.0.1"
+        "react-merge-refs": "^2.0.1",
+        "react-use-measure": "^2.1.1"
     },
     "peerDependencies": {
         "@react-three/fiber": "^8.2.2",
--- a/packages/framer-motion-3d/src/components/MotionCanvas.tsx
+++ b/packages/framer-motion-3d/src/components/MotionCanvas.tsx
@@ -1,7 +1,7 @@
 import * as React from "react"
 import {
     useContext,
-    useLayoutEffect,
+    useEffect,
     useRef,
     forwardRef,
     MutableRefObject,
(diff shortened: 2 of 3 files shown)


mattgperry:
My initial thought is the reason `MotionCanvas` isn't so good with this is to support layout animations with `LayoutCamera`. It might be a good idea to grab an example from the docs and set up a test with that to see if layout animations work correctly with this.

sdacunha:
> My initial thought is the reason `MotionCanvas` isn't so good with this is to support layout animations with `LayoutCamera`. It might be a good idea to grab an example from the docs and set up a test with that to see if layout animations work correctly with this.


@mattgperry Interesting - will try to take a look at this. Looking at the [LayoutCamera example](https://codesandbox.io/s/framer-motion-3d-layoutcamera-i84e3?from-embed), it seems like resizing the window has some weird behaviour currently after clicking to start the animation. Nonetheless, will look into testing this behaviour when I can

kavitalikesfajitas:
Is there any progress on this?

https://codesandbox.io/s/framer-motion-3d-layoutcamera-i84e3?from-embed even your own example does not resize 
https://www.framer.com/motion/layoutcamera/

mattgperry:
Just closing this for now just to keep things clean as I think without incorporating `LayoutCamera` it isn't possible to merge in. 

Apologises that the issue is long-standing - Framer Motion 3D only has 6k weekly downloads vs 3.9m for normal Framer Motion so it's currently difficult to justify too much time on it.

PR #2051, opened 2023-03-24
Fixed typo in LazyMotion component and its types

faustinozanetto:
While using the LazyMotion component from Framer Motion I encountered a small typo. It says that I should import domAnimations but I could not find such file exported from motion, and then I realized that It was domAnimation. So Im making this basic PR to address this typo fix.


mattgperry:
Thanks for the PR!

Kori000:
Hi, i noticed that the [source code](https://github.com/framer/motion/blob/main/packages/framer-motion/src/components/LazyMotion/types.ts) has been updated, but there is no update in the [documentation](https://www.framer.com/motion/lazy-motion/#props).
image

PR #2057, opened 2023-03-29
Improving scroll animation startup time

mattgperry:
Screenshot 2023-03-29 at 13 54 46

Screenshot 2023-03-29 at 13 56 02

### DIFF
--- a/packages/framer-motion/src/value/use-spring.ts
+++ b/packages/framer-motion/src/value/use-spring.ts
@@ -63,6 +63,7 @@ export function useSpring(
                 onUpdate: set,
             })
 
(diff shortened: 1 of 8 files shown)


packages/framer-motion/src/value/use-spring.ts
```diff
@@ -63,6 +63,7 @@ export function useSpring(
                 onUpdate: set,
             })
 
+            console.log(frameData.isProcessing)
```
hemlok, line 66: 👀

hemlok (approved):
Nice one!

shuangq (approved):
code looks good!

Issue #2059, opened 2023-03-31
[BUG] The new useAnimate sequence feature causes errors and warnings


sukjae:
**1. Describe the bug**

Hello. 
In the example below, I expected sequence 1, 2, and 3 to all work the same.
However, some errors and warnings are coming out. 

### sequence error
An error is coming out when using sequence. 
Am I misunderstanding the sequence? 

The sequence that I understood is the same concept as timeline, and I thought it was similar to timeline in gsap and sequence in motion one. 

However, if you run multiple motions in series in the sequence one example, the following error occurs. 

```js
animate([
  [".one", { x: 100 }],
  [".one", { y: 100 }]
]);
```

> Error
All keyframes must be of the same type

<img width="525" alt="image" src="https://user-images.githubusercontent.com/8156543/229001614-ad7f43b0-c29e-4c60-84e1-3912d17ad049.png">

### transform warning
When performing transform sequentially, the following warning occurs. 
The app doesn't shut down, but it doesn't do anything. 

```js
await animate([".three", { transform: "translateX(100px)" }]);
await animate([".three", { transform: "translateY(100px)" }]);
```

> You are trying to animate  from ".three,[object Object]" to "undefined". .three,[object Object] is not an animatable value - to enable this animation set .three,[object Object] to a value animatable to undefined via the `style` property.

<img width="1453" alt="image" src="https://user-images.githubusercontent.com/8156543/229001643-3d74ddda-76a2-46cb-bf8a-d8d928dd7ab4.png">

**2. IMPORTANT: Provide a CodeSandbox reproduction of the bug**

- https://codesandbox.io/s/great-cherry-pxgogh?file=/src/App.js

**3. Steps to reproduce**

- click play buttons in codesandbox.

sukjae:
The second problem was not an error. It was an issue that had to be wrapped up in another arrangement.

### as-is

```js
await animate([".three", { transform: "translateX(100px)" }]);
await animate([".three", { transform: "translateY(100px)" }]);
```

### to-be

```js
await animate([[".three", { transform: "translateX(100px)" }]]);
await animate([[".three", { transform: "translateY(100px)" }]]);
```

mattgperry:
The first issue will be fixed in `10.11.0` going out now.

```javascript
await animate([[".three", { transform: "translateX(100px)" }]]);
```

Can just be

```javascript
await animate(".three", { transform: "translateX(100px)" });
```

[...]

PR #2069, opened 2023-04-06
Fix support of `none` keyframes

mattgperry:
Fixes https://github.com/framer/motion/issues/2059

This PR improves keyframe parsing, converting incoming values to something animatable. Specifically, better typings for non-zero `null` keyframes, non-zero indexed `"0"` and `"none"` keyframes are also converted into animatable values based on other values in the list.

```javascript
["none", "translateX(100px)", "none"] -> ["translateX(0px)", "translateX(100px)", "translateX(0px)"]
[null, 1, null] -> [defaultOrigin, 1, 1]
[0, "20px"] -> ["0px", "20px"]
```


hemlok (approved):
This some hot hot heat 🌶️

PR #2084, opened 2023-04-13
Remove CSS variables before interpolating

mattgperry:
This PR ensures CSS variable tokens are removed from complex values before interpolation, so we don't end up animating numbers in tokens.

This PR doesn't allow us to animate strings containing multiple tokens, but these styles will still be set as expected if left unresolved.

### DIFF
--- a/packages/framer-motion/src/value/types/complex/index.ts
+++ b/packages/framer-motion/src/value/types/complex/index.ts
@@ -39,7 +39,13 @@ export function analyseComplexValue(v: string | number) {
         values.push(...numbers.map(number.parse))
     }
 
-    return { values, numColors, numNumbers, tokenised: v }
+    return {
+        values,
+        numColors,
+        numNumbers,
+        tokenised: v,
+        containsCSSVars: v.includes("var(--"),
+    }
 }
 
 function parse(v: string | number) {
--- a/packages/framer-motion/src/value/types/complex/index.ts
+++ b/packages/framer-motion/src/value/types/complex/index.ts
@@ -59,23 +59,24 @@ function tokenise(
     info: ComplexValueInfo,
     { regex, countKey, token, parse }: Tokeniser
 ) {
-    const matches = info.value.match(regex)
+    const matches = info.tokenised.match(regex)
 
     if (!matches) return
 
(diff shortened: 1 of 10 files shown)


packages/framer-motion/src/value/types/complex/index.ts
```diff
@@ -16,50 +18,99 @@ function test(v: any) {
...
+    if (!matches) return
+
+    info["num" + countKey] = matches.length
+    info.value = info.value.replace(regex, token)
+    info.values.push(...(matches.map(parse) as any))
+}
+
+export function analyseComplexValue(value: string | number): ComplexValueInfo {
+    const info = {
+        value: typeof value === "number" ? "" + value : value,
```
shuangq, line 73:
```suggestion
        value:value.toString(),
```

packages/framer-motion/src/value/types/complex/index.ts
```diff
@@ -16,50 +18,99 @@ function test(v: any) {
...
+    info: ComplexValueInfo,
+    { regex, countKey, token, parse }: Tokeniser
+) {
+    const matches = info.value.match(regex)
+
+    if (!matches) return
+
+    info["num" + countKey] = matches.length
+    info.value = info.value.replace(regex, token)
+    info.values.push(...(matches.map(parse) as any))
```
shuangq, line 68: You've prolly thought this through, but why do we rely on the index instead of updating the shape of `values`? For example, `values: {vars: [], colors: [], nums: []}`, this way, we won't need the countKey nor the counts?

shuangq (approved):
Code looks good, left a question regarding the API, but I don't think it's a blocker.

packages/framer-motion/src/value/types/complex/index.ts
mattgperry: If I have thought this through I think it was six years ago and I was basically just copying what I did then. You're right. I'll take another look.

mattgperry: As I'm refactoring this I remember it's so I can use a generic `mixArray` on the single values array

PR #2100, opened 2023-04-26
Fix relative animations with mixed instant/time: 0 transitions

mattgperry:
This PR disables relative animations for animations that have already completed.

This is essentially a quick fix that checks whether an animation has already completed before resolving a relative box. Reason being, the box can be resolved against a parent box which, if animating using a `duration: 0` animation, will itself complete a frame later leaving the relative box incorrect for subsequent frames.

### DIFF
--- a/dev/projection/animate-relative-mixed-transition.html
+++ b/dev/projection/animate-relative-mixed-transition.html
@@ -1,0 +1,101 @@
+<html>
+    <head>
+        <style>
+            body {
+                padding: 0;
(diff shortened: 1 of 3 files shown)


mattgperry:
No it just came in via Slack

shuangq (approved):
The code and test look good, but I'm not entirely certain about the correlation between the fix and the mix of instant and duration 0 transitions. Could you provide some more context in the PR description?

PR #2104, opened 2023-05-02
Fix instant layout animation with React double renders

mattgperry:
This PR fixes `useInstantLayoutAnimation` after a double render by only unlocking layout animations in a microtask, which fires after all synchronous double renders have flushed.


shuangq (approved):

shuangq:
Can we cover a test for this?

mattgperry:
Closed in favour of https://github.com/framer/motion/pull/2105

PR #2105, opened 2023-05-03
Move layout animations to `queueMicrotask`

mattgperry:
- Prevents early node mounting via `data-projection-id`.
- Skips unnecessary/duplicate layout animations when setting state in `useLayoutEffect`.

Future PRs will genericise the read/write lifecycles so we can hook into them with other bits that would benefit from batching like SVG measurements.


hemlok (approved):
very cool, nice that projection id is going away!

PR #2113, opened 2023-05-08
Use ref for exiting children (Fixes #1914)

### LINKED ISSUE #1914: [BUG]
Done

**2. Describe the bug**
`AnimatePresence` item is never unmounted if its exit is started during the animation of another item.

[Code Sandbox](https://codesandbox.io/s/animatepresence-usepresence-forked-8z1qgr)

**4. Steps to reproduce**
1. Open the sandbox
2. Click the "Remove" button three times rapidly (before any of the items is fully removed)

**5. Expected behavior**
I expect all items to be removed from the DOM. To produce the expected behavior:
1. Open the sandbox
2. Click the "Remove" button, waiting for each element to be removed from the page before clicking again.

**6. Video or screenshots**
### Working
Each item animates out if the previous removal completes
CleanShot 2023-01-27 at 11 58 33

### Bug
Any items removed during the exit animation fail to be completely removed
CleanShot 2023-01-27 at 11 58 48

**7. Environment details**
Tested in Firefox. Dependencies are in the sandbox.

hpersson:
Fixes https://github.com/framer/motion/issues/1914

The `onExit` callback for exiting children could have stale values for the set of `exiting` keys. This would lead to the last "exiting" element to not realize they were the last exiting element, and therefore not trigger a force render (specific code below)

```
if (!exiting.size) {
                presentChildren.current = filteredChildren

                if (isMounted.current === false) return

                forceRender()
                onExitComplete && onExitComplete()
            }
```

This means already-exited elements would still be visible until the next render (if there was one)

I looked into adding a test for this case. However, in tests, the use of `React.StrictMode` created a lot of extra renders which would "fix" this broken state before I could assert against it. Even experimenting with not using React.StrictMode, I still saw more than 1 render when calling `rerender` and could not get a clean failing test. Let me know if you have any suggestions!

### DIFF
--- a/packages/framer-motion/src/components/AnimatePresence/index.tsx
+++ b/packages/framer-motion/src/components/AnimatePresence/index.tsx
@@ -100,7 +100,7 @@ export const AnimatePresence: React.FunctionComponent<
     const filteredChildren = onlyElements(children)
     let childrenToRender = filteredChildren
 
-    const exiting = new Set<ComponentKey>()
+    const exitingChildren = useRef(new Map<ComponentKey, ReactElement<any> | undefined>()).current
 
     // Keep a living record of the children we're actually rendering so we
     // can diff to figure out which are entering and exiting
@@ -125,7 +125,7 @@ export const AnimatePresence: React.FunctionComponent<
     useUnmountEffect(() => {
         isInitialRender.current = true
         allChildren.clear()
-        exiting.clear()
+        exitingChildren.clear()
     })
 
     if (isInitialRender.current) {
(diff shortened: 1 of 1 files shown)


mattgperry:
Hey, great PR, thanks! Could you commit the test with multiple renders and no StrictMode so I can look into why it's rerendering?

hpersson:
sure - i've pushed the test up! (this test does not fail on main, so it's not reproducing the bug)

Looking at it a bit more, I actually think React Strict Mode is unrelated to the difference in rendering - both the sandboxed app and the tests run in strict mode, with different results.

I adding some console.log statements to try to demonstrate with screenshots - one at the top of the "AnimatePresence" component and one in the onExit callback

when i run against my local version of the [sandboxed example](https://codesandbox.io/s/animatepresence-usepresence-forked-8z1qgr?file=/src/index.js), I see the expected sequence of events that leads to the onExit cb being the last thing executed (showing the bug: we fail the check `!exiting.size` that should be calling forceRender() and there isn't a final render after the final onExit cb)

Screenshot 2023-05-08 at 9 18 25 PM

when i run the unit test i've just committed (+ `console.log("toggle clicked!")` on the lines above where i call `rerender`) , I see the following:
Screenshot 2023-05-08 at 9 40 56 PM

in the test code specifically, both cbs execute in the same tick so there _is_ a final render that shows the correct state.

let me know if that makes sense!

mattgperry:
Ah yeah that makes sense thanks, I'll take a look!

mattgperry:
I think I've got it breaking correctly by changing the test to:

```javascript

        // Remove the second-to-last element from the array
        await act(async () => {
            rerender(<Component visibleKeys={[0]} />)

            await new Promise<void>((resolve) => {
                //  Resolve after all animation is expected to have completed
                setTimeout(() => {
                    resolve()
                }, 1000)
            })
        })
```

mattgperry:
Ok I've got a test that fails on main and passes on this branch.

https://github.com/framer/motion/pull/2119

Once I've merged that I'll merge this and remove the Jest test. Thanks again!

Issue #2132, opened 2023-05-15
[BUG] `animate()` produces incorrect values if a `layout` element is present (wrong `frameData.timestamp`)


mikirejf:
**1. Read the [FAQs](#faqs) 👇**

✅ 

**2. Describe the bug**

If we have a component with an `useEffect` that conditionally invokes `animate()` and a markup with a layout motion component, the animate doesn't produce all values. It seems like the `frameData` gets invoked as soon as the component mounts hence the animate producing only the start and the end value on updates. 

E.g. if we have an animation that takes 4s, and we wait for 2s before we invoke it, the function will produce only the values from 2s to 4s. If we wait for 4s or more before invoking, only the start and the end value will be produced.

**3. IMPORTANT: Provide a CodeSandbox reproduction of the bug**

[Codesandbox](https://codesandbox.io/s/https-github-com-framer-motion-issues-2132-v5rwqo)

**4. Steps to reproduce**

1. Open the js console.
2. Click on the "increment" button. The animate function runs on every `count % 2 !== 0`.
3. Observe how only the start and the end values are logged.
4. Use one of the workarounds (see the code).
5. Increment the counter and observe how the value gets interpolated.

**5. Expected behavior**

The value we are animating should be interpolated accordingly.

**6. Video or screenshots**

https://github.com/framer/motion/assets/8895894/1b6246c9-a12b-4607-8fc7-638e5ca05d10

**7. Environment details**

The issue happens cross-browser.

PR #2133, opened 2023-05-16
Fixing timing bug with `animate()` and `layout` prop

mattgperry:
Fixes https://github.com/framer/motion/issues/2132

This PR fixes a bug with layout animations manually flushing a single animation frame whereby a stale timestamp might be used in some situations. This causes a conflict with `animate()` where if it schedules an animation before layout animations are flushed, the second animation frame will report a much larger time delta than could have occurred.

### DIFF
--- a/packages/framer-motion/src/projection/node/create-projection-node.ts
+++ b/packages/framer-motion/src/projection/node/create-projection-node.ts
@@ -47,6 +47,7 @@ import { ValueAnimationOptions } from "../../animation/types"
 import { frameData } from "../../dom-entry"
 import { isSVGElement } from "../../render/dom/utils/is-svg-element"
 import { animateSingleValue } from "../../animation/interfaces/single-value"
+import { clamp } from "../../utils/clamp"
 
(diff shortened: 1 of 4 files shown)


packages/framer-motion/src/projection/node/create-projection-node.ts
```diff
@@ -626,10 +626,17 @@ export function createProjectionNode<I>({
             this.nodes!.forEach(notifyLayoutUpdate)
             this.clearAllSnapshots()
 
-            // Flush any scheduled updates
+            /**
+             * Manually flush any pending updates. Ideally
+             * we could leave this to the following requestAnimationFrame but this seems
+             * to leave a flash of incorrectly styled content.
+             */
+            frameData.timestamp = performance.now()
```
shuangq, line 634: Do we also need to update frameData.delta here?

mattgperry: Good point, I'll take a look

shuangq (approved):

PR #2138, opened 2023-05-16
Use `0` string values to template numerical values

mattgperry:


packages/framer-motion/src/animation/utils/keyframes.ts
```diff
@@ -36,7 +36,14 @@ export function getKeyframes(
...
         if (isNone(keyframes[i])) {
             noneKeyframeIndexes.push(i)
-        } else if (typeof keyframes[i] === "string") {
+        }
+
+        // TODO: Clean this conditional, it works for now
+        if (
+            typeof keyframes[i] === "string" &&
+            keyframes[i] !== "none" &&
+            keyframes[i] !== "0"
```
shuangq, line 45: do we need to consider "0px" as well?

mattgperry: no we want that to be used as a template

shuangq (approved):

PR #2150, opened 2023-05-22
Fix scale correction after animation finishes

mattgperry:
Test

### DIFF
--- a/packages/framer-motion/src/projection/node/create-projection-node.ts
+++ b/packages/framer-motion/src/projection/node/create-projection-node.ts
@@ -493,6 +493,7 @@ export function createProjectionNode<I>({
                              * finish it immediately. Otherwise it will be animating from a location
                              * that was probably never commited to screen and look like a jumpy box.
                              */
+
                             if (!hasLayoutChanged) {
                                 finishAnimation(this)
                             }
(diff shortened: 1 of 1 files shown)


packages/framer-motion/src/projection/node/create-projection-node.ts
```diff
@@ -1225,6 +1226,12 @@ export function createProjectionNode<I>({
              */
             copyBoxInto(this.layoutCorrected, this.layout.layoutBox)
 
+            /**
+             * Record previous tree scales before updating.
```
mattgperry, line 1230: Moving this before updating treeScale as apparently this would have been completely useless before - potentially leading to skipped renders (which would be hard to capture in a test)

packages/framer-motion/src/projection/node/create-projection-node.ts
```diff
@@ -1245,6 +1264,7 @@ export function createProjectionNode<I>({
                  * a render to ensure the elements reflect the removed transform.
                  */
                 if (this.projectionTransform) {
+                    this.projectionDelta = createDelta()
```
shuangq, line 1267: Why do we need this?

shuangq: And do we need to reset `this.projectionDeltaWithTransform` as well?

shuangq (approved):

packages/framer-motion/src/projection/node/create-projection-node.ts
mattgperry: Basically by setting `projectionTransform` to `"none"` we want to reset the underlying data structure to (`createDelta` returns a none transform)

PR #2177, opened 2023-06-04
Handled some errors that only happen when using server-side rendering…

abdoachhoubi:
… frameworks/libraries


packages/framer-motion/src/render/dom/utils/unit-conversion.ts
```diff
@@ -109,7 +109,7 @@ const convertChangedValueTypes = (
 ) => {
     const originBbox = visualElement.measureViewportBox()
     const element = visualElement.current
-    const elementComputedStyle = getComputedStyle(element!)
```
mattgperry, line 112: Can you write a failing test for this? See the `ssr.test.tsx` file for examples

mattgperry:
Closing for now but feel free to reopen with failing test

PR #2198, opened 2023-06-22
Improve useInstantTransition

adamseckel:
`useInstantTransition` can get into a bad state if its callback is called in subsequent frames.


shuangq (approved):
Is it possible to capture this issue in a test?

adamseckel:
> Is it possible to capture this issue in a test?

Wondering the same. Let me see if there are test utilities to run the frames in a controlled way 🤔

Issue #2209, opened 2023-06-27
[BUG] ``useInView`` Update hook called on initial render


ariesclark:
Error: Update hook called on initial render. ~~This is likely a bug in React. Please file an issue.~~

```ts
const reference = useRef<HTMLDivElement>(null);
const inView = useInView(reference);
```

mattgperry:
This is expected behaviour. On initial render there's no way of knowing whether the component is in view.

ariesclark:
> This is expected behaviour. On initial render there's no way of knowing whether the component is in view.

Throwing an error on render is **not expected behaviour**. It should either default to false, or be able to receive a fallback value on initial render, It is absurd to throw an error like this as _intended behaviour_, seriously..

ariesclark:
For anyone facing this issue, I suggest using [react-intersection-observer](https://github.com/thebuilder/react-intersection-observer) instead, as it reasonably provides a ``initialInView`` property instead of throwing an error.

PR #2216, opened 2023-07-02
Issue number: 2209, useInView Update hook called on initial render

Arun-chaitanya:


Arun-chaitanya:
close #2209

mattgperry:
This would leave components that are actually in view with an incorrect state.

Issue #2172, opened 2023-06-02
[BUG] layoutId and AnimatePresence together fail to unmount content


onnlucky:
```ts
import * as React from "react";
import { createRoot } from "react-dom/client";
import { AnimatePresence, motion } from "framer-motion";

const rootElement = document.getElementById("root");
const root = createRoot(rootElement);

const animation = {
  x: 0, // Comment this out and it works.
  opacity: 0.5
};

export const Screen = function Component() {
  const [visible, setVisible] = React.useState(false);
  return (
    <>
      <motion.div
        className={"button"}
        onTap={() => setVisible(!visible)}
        style={{ width: 100, height: 100, background: "green" }}
      />
      <AnimatePresence>
        {visible && (
          <motion.div
            layoutId={"An2J7Niah"} // Comment this out and it works.
            style={{ width: 100, height: 100, background: "blue" }}
            className={"BLUE"}
            exit={animation}
          />
        )}
      </AnimatePresence>
    </>
  );
};

root.render(<Screen />);
```

With the above code (code sandbox: https://codesandbox.io/s/aged-leaf-1tlyr2?file=/src/index.tsx ), click the green square, will show a blue square, click it again, it should disappear after an exit animation. But instead it will never go away.

This is a regression in 10.12.8. And one way to fix it is to change `create-projection-node.ts` use of the `queueMicrotask`, like this:
```ts
        didUpdate() {
            if (!this.updateScheduled) {
                this.updateScheduled = true
                // queueMicrotask(() => this.update())
                this.update()
            }
        }
```

onnlucky:
This seems related to: https://github.com/framer/motion/issues/2145 and https://github.com/framer/motion/issues/2110

PR #2220, opened 2023-07-10
Fixing `Promise` race condition

mattgperry:
Fixes https://github.com/framer/motion/issues/2172
Fixes https://github.com/framer/company/issues/27486

### DIFF
--- a/packages/framer-motion/src/animation/animators/js/index.ts
+++ b/packages/framer-motion/src/animation/animators/js/index.ts
@@ -65,7 +65,6 @@ export function animateValue<V = number>({
      * WAAPI-compatible behaviour.
      */
     const updateFinishedPromise = () => {
-        resolveFinishedPromise && resolveFinishedPromise()
         currentFinishedPromise = new Promise((resolve) => {
             resolveFinishedPromise = resolve
         })
@@ -272,6 +271,7 @@ export function animateValue<V = number>({
     const cancel = () => {
         playState = "idle"
         stopAnimationDriver()
+        resolveFinishedPromise()
         updateFinishedPromise()
         startTime = cancelTime = null
     }
@@ -280,7 +280,7 @@ export function animateValue<V = number>({
         playState = "finished"
         onComplete && onComplete()
         stopAnimationDriver()
-        updateFinishedPromise()
+        resolveFinishedPromise()
     }
 
     const play = () => {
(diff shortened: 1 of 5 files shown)


packages/framer-motion/src/components/AnimatePresence/index.tsx
```diff
@@ -181,7 +183,7 @@ export const AnimatePresence: React.FunctionComponent<
         const insertionIndex = presentKeys.indexOf(key)
 
         let exitingComponent = component
-        if (!exitingComponent) {
```
mattgperry, line 184: This file is just Prettier

packages/framer-motion/src/animation/animators/js/index.ts
```diff
@@ -65,7 +65,6 @@ export function animateValue<V = number>({
      * WAAPI-compatible behaviour.
      */
     const updateFinishedPromise = () => {
-        resolveFinishedPromise && resolveFinishedPromise()
```
shuangq, line 68: The comment above needs to be updated as well.

packages/framer-motion/src/animation/animators/js/index.ts
```diff
@@ -280,7 +280,7 @@ export function animateValue<V = number>({
         playState = "finished"
         onComplete && onComplete()
         stopAnimationDriver()
-        updateFinishedPromise()
+        resolveFinishedPromise()
```
shuangq, line 283: Could you elaborate a little more on the race condition? Why do we stop updating the finished promise in `finish()`?

mattgperry: For reasons slightly beyond my comprehension, but at least related to adding `queueMicrotasks` (which is when Promises are resolved), swapping to a new `Promise` as soon as we resolve the old one messes with `Promise.all` - seemingly at random. As noted in Onne's linked ticket, even adding a `console.log` can affect timings in a way that more consistently fixes the observed bug. So instead we swap to a new `Promise` later - if and when we start a new animation.

shuangq: > swapping to a new Promise as soon as we resolve the old one messes with Promise.all
Would it have the same impact when `cancel()` is called?

mattgperry: I haven't see a bug related to this but it's added there because it's been removed from `updateFinishedPromise()`

shuangq (approved):
Let's add a comment why we skip it in `finish()`, I think it requires context to understand why the code is written this way.

pofixifop:
i still have issues with exiting components in quick succession getting stuck, and never getting to the proper final component, unless i comment out all `exit` props

PR #2225, opened 2023-07-12
Fix initial animation of background

mattgperry:
Fixes https://github.com/framer/company/issues/27789

This makes sure we try to read an initial value from component props before reading from the DOM. These values are more likely to be animatable.

### DIFF
--- a/dev/tests/layout-exit.tsx
+++ b/dev/tests/layout-exit.tsx
@@ -9,6 +9,8 @@ export const App = () => {
         opacity: 0.5,
     }
 
+    React.useEffect(() => setVisible(!visible), [])
+
     return (
         <>
             <AnimatePresence>
@@ -16,7 +18,6 @@ export const App = () => {
                     <motion.div
                         id="box"
                         layout
-                        onTap={() => setVisible(!visible)}
                         style={{ width: 100, height: 100, background: "blue" }}
                         exit={animation}
                     />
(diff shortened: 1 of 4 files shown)


dev/tests/layout-exit.tsx
```diff
@@ -9,14 +9,15 @@ export const App = () => {
         opacity: 0.5,
     }
 
+    React.useEffect(() => setVisible(!visible), [])
```
mattgperry, line 12: Unrelated, fixing test from previous PR

packages/framer-motion/src/render/VisualElement.ts
```diff
@@ -827,7 +827,8 @@ export abstract class VisualElement<
     readValue(key: string) {
         return this.latestValues[key] !== undefined || !this.current
             ? this.latestValues[key]
-            : this.readValueFromInstance(this.current, key, this.options)
+            : this.getBaseTargetFromProps(this.props, key) ||
+                  this.readValueFromInstance(this.current, key, this.options)
```
shuangq, line 831: Should this be a nullish check instead? In case the returned value is an empty string or 0.

shuangq (approved):

packages/framer-motion/src/render/VisualElement.ts
mattgperry: Yeah great catch

PR #2227, opened 2023-07-13
Transform newly-read values

mattgperry:
Values read from the DOM weren't being passed to `transformValues` if this was present.

Now initial values are being fished out of `style` before `getComputedStyle()` values like `"red"` may not correctly be animated in Framer.

### DIFF
--- a/packages/framer-motion/src/render/dom/DOMVisualElement.ts
+++ b/packages/framer-motion/src/render/dom/DOMVisualElement.ts
@@ -54,7 +54,17 @@ export abstract class DOMVisualElement<
         }
 
         if (isMounted) {
-            checkTargetForNewValues(this, target, origin as any)
+            const newValues = checkTargetForNewValues(
+                this,
+                target,
+                origin as any
+            )
+
+            // If new values have been added, for instance named color values, transform these
+            if (transformValues && newValues) {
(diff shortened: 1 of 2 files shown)


packages/framer-motion/src/render/dom/DOMVisualElement.ts
```diff
@@ -54,7 +54,17 @@ export abstract class DOMVisualElement<
...
-            checkTargetForNewValues(this, target, origin as any)
+            const newValues = checkTargetForNewValues(
+                this,
+                target,
+                origin as any
+            )
+
+            // If new values have been added, for instance named color values, transform these
+            if (transformValues && newValues) {
+                origin = transformValues(origin as any)
```
nvh, line 65: Is there any way to avoid this cast?

mattgperry:
We could but I would rather kick this in, it's basically `FramerStudio` functionality and the test there successfully caught this, it just should have run on the upgrade PR

packages/framer-motion/src/render/dom/DOMVisualElement.ts
mattgperry: Maybe but I'm just copy pasting the existing line above and would rather just get it in

PR #2231, opened 2023-07-15
Fix incorrect Reorder.Item order calculation

### LINKED ISSUE #2101: [BUG] Reordering randomly stuck
**2. Describe the bug**

Simple dnd for tabs reordering implemented based on the mantine framework and Reorder component from framer. 
At some conditions tab does not return to its position and stuck half way after mouse release. 

[](https://codesandbox.io/s/unruffled-smoke-jz0wic?file=/src/App.js:655-828)

**4. Steps to reproduce**

1. Reorder tabs with different speed and direction. Release mouse button randomly.
3. Tab does not return to its original place

Note: issue is flaky, it may not be reproduced from the first time. Need to try multiple times.
Note2: issue also reproducible with simple divs, but it much harder to achieve than with this tab component.  

**5. Expected behavior**

Tab should return to its original position.

**6. Video or screenshots**
image

**7. Environment details**

Chrome (Version 111.0.5563.146 (Official Build) (64-bit))
Chrome on iPad (latest versions)

obeattie:
## Problem

I've been experiencing an issue with `Reorder` lately, where items can become "stuck" while being dragged. It's a little hard to describe in words, but here's a recording of an example from an app I am building:

Before

As you can see here, the "Sydney" item is first dragged over "Vancouver", and then "London" without a problem. But when "Vancouver" is then dragged backwards over "London", "Vancouver" also moves, even though the user hasn't moved the cursor that far to the left yet. As the dragging continues, it becomes totally impossible to reorder the items. This is reliably reproducible in my example, but not in more trivial examples, leading to much frustration.

## Cause

The underlying issue is that `Reorder.Group` has an incorrect view of the current order of its items on the page. To determine that order, [relies on the layout measurements it is passed by its child `Reorder.Item`s](https://github.com/framer/motion/blob/10bfeace9183f344b2098ee659dc0c6848f0d705/packages/framer-motion/src/components/Reorder/Group.tsx#L87-L98) via a `registerItem` callback. But [the code in `Reorder.Item` to pass the layout measurements](https://github.com/framer/motion/blob/10bfeace9183f344b2098ee659dc0c6848f0d705/packages/framer-motion/src/components/Reorder/Item.tsx#L74-L100) has a race condition. The intended flow seems to be:

1. Layout measurements are taken and passed to the items' `onLayoutMeasure` handler
2. The handler stores the measurements in a ref
3. An effect reads the measurements from the ref on each render and invokes the `registerItem` callback

But there is a race condition between 1 and 3, and sometimes 3 can execute before 1. I'm not familiar enough with React's internals to understand why the two hooks execute in a non-deterministic order, but they do. And when that happens, stale layout values from the last render are passed to `registerItem` and the `Reorder.Group` gets an incorrect order.

## Solution

Instead of using a ref to temporarily hold the layout information, I'm just invoking `registerItem` directly in the `onLayoutMeasure` callback.

With this patch, the issue disappears and the reordering works as you would expect:

After

I've also tried this fix with the example in #2101. I can reproduce that issue pretty easily with framer-motion 10.12.21, and it seems fixed with this patch. Perhaps some of the other open Reorder bugs can be attributed to this, too. (Fixes #2101).

## Notes

There aren't any comments in the code to explain why the extra indirection with the ref is necessary, and [the commit that introduced the code](https://github.com/framer/motion/commit/be5c949daf652167ff3f4365135ac6fc42369c02) doesn't contain any helpful description either. Being unsure why that choice was made, I'm left wondering if it was meant to solve a problem I am now overlooking. But, things seem to work just as well as before with my change, and the bug is fixed too.

### DIFF
--- a/packages/framer-motion/src/components/Reorder/Group.tsx
+++ b/packages/framer-motion/src/components/Reorder/Group.tsx
@@ -85,16 +85,15 @@ export function ReorderGroup<V>(
     const context: ReorderContextProps<any> = {
         axis,
         registerItem: (value, layout) => {
-            /**
-             * Ensure entries can't add themselves more than once
-             */
-            if (
-                layout &&
-                order.findIndex((entry) => value === entry.value) === -1
-            ) {
-                order.push({ value, layout: layout[axis] })
-                order.sort(compareMin)
+            if (!layout) return;
+            // If the entry was already added, update it rather than adding it again
+            const idx = order.findIndex((entry) => value === entry.value)
+            if (idx !== -1) {
+                order[idx].layout = layout[axis]
+            } else {
+                order.push({ value: value, layout: layout[axis] })
             }
+            order.sort(compareMin)
         },
         updateOrder: (id, offset, velocity) => {
             if (isReordering.current) return
--- a/packages/framer-motion/src/components/Reorder/Item.tsx
+++ b/packages/framer-motion/src/components/Reorder/Item.tsx
@@ -4,12 +4,9 @@ import {
     ReactHTML,
     FunctionComponent,
     useContext,
-    useEffect,
-    useRef,
     forwardRef,
 } from "react"
 import { ReorderContext } from "../../context/ReorderContext"
-import { Box } from "../../projection/geometry/types"
 import { motion } from "../../render/dom/motion"
 import { HTMLMotionProps } from "../../render/html/types"
 import { useConstant } from "../../utils/use-constant"
@@ -71,15 +68,9 @@ export function ReorderItem<V>(
         latestX || latestY ? 1 : "unset"
     )
 
-    const measuredLayout = useRef<Box | null>(null)
-
     invariant(Boolean(context), "Reorder.Item must be a child of Reorder.Group")
 
     const { axis, registerItem, updateOrder } = context!
-
-    useEffect(() => {
-        registerItem(value, measuredLayout.current!)
-    }, [context])
 
     return (
         <Component
@@ -95,9 +86,7 @@ export function ReorderItem<V>(
 
                 onDrag && onDrag(event, gesturePoint)
             }}
-            onLayoutMeasure={(measured) => {
-                measuredLayout.current = measured
-            }}
+            onLayoutMeasure={(measured) => registerItem(value, measured)}
             ref={externalRef}
             ignoreStrict
         >


nick-keller (approved):

nick-keller:
Awesome work, thx!

nick-keller:
Hey @mattgperry,
This looks like a fix to an impacting issue, with minimal footprint on existing code. What is the status on this, anyway we can help?

chuganzy:
Sharing what I found.
I was investigating another race condition where `transitionEnd` styles are applied _after_ new animation starts and accidentally overrides the latest motion values.

It seems to be introduced in https://github.com/framer/motion/pull/2025. Before it lands, calling `stop()` never completes animation Promise, which is used to apply the `transitionEnd` styles, so there was no race condition..

I did not dig into how it is related to Reorder (though I can imagine it is related), but rolling back it somehow seems to solve this issue.

obeattie:
That's great sleuthing 🕵️‍♂️

In my opinion it would still be a good thing to remove this indirection from the code anyway, regardless of what causes the execution of the two hooks to be reordered. Passing the layout measurements between two hooks via a ref seems unnecessary and makes the code brittle to an order of execution that it doesn't control.

chuganzy:
I agree that this PR's approach looks more robust. One thing I can come up with regarding the existing approach is for the better performance? but not quite sure..😅

While I was investigating it further, I found that `Reorder` got unstable after https://github.com/framer/motion/pull/1993 is merged (v10.2.4), and another fix is needed to make it more stable. 

It happens randomly when you drag an item, stop the cursor for a while, and drop it as attached bellow:

https://github.com/framer/motion/assets/2619092/d29be0a0-1456-48a7-bd72-d127fa51beb0

I have confirmed that this still happens even after applying this PR, and the fix will be soon ready to submit!

obeattie:
@chuganzy: I've tried your patch from #2264. For clarity, I don't think it fixes the same issue that is described here. The inertial animation code you have modified is only invoked when an item is dropped (ie. on mouseup), but the issue I have tried to fix happens while the item is being dragged (ie. while the mouse button is down).

chuganzy:
Sorry I was not clear enough. Yes, your fix is still needed, mine fixes the end of the animation.

nick-keller:
Hey @mattgperry, 
Anything we can do to help you merge this PR?

obeattie:
In case it's helpful to share, I'm using this with [patch-package](https://www.npmjs.com/package/patch-package) until if/when it's merged.

robertotcestari:
Any updates here?

nick-keller:
Are there any blockers to merge this PR?

obeattie:
I messaged the maintainer on Twitter (fine, X). The hold-up is that this PR isn't mergeable without an integration test.

I don't have time immediately to work on this, but I'll try in the next few weeks.

Alternatively others could feel free to try and write this.

thewildandy:
It would be great to see this move forward, it's a shame to not see the benefit of any of the work being done on motion due to such a major feature being broken despite a solution existing. Appreciate @nick-keller for persisting with this 🙌

chuganzy:
Hi! JFYI: I just opened a PR that cherry-picks the changes here, and adds integration test! 
https://github.com/framer/motion/pull/2359

PR #2239, opened 2023-07-18
`scroll()`

mattgperry:
This PR adds a new universal API, `scroll()`. This function provides a scroll progress value to callbacks. It is powered by `ScrollTimeline`, where available, for 40% less per-frame JS and no DOM measurements.

The intention is to, in the future, expand this function to:
- Support element tracking via `ViewTimeline`
- Accept `animate()` animations for hardware accelerated animations


packages/framer-motion/src/render/dom/scroll/track.ts
```diff
@@ -0,0 +1,109 @@
+import { frame, cancelFrame, frameData } from "../../../frameloop"
```
mattgperry, line 1: This is the previous `scroll/index` and shouldn't need a review.

adamseckel:
Just confirming: `scroll` now returns only a progress number, and the previous `scroll` is now `scrollInfo`?

mattgperry:
@adamseckel Yeah that's right. Ideally this will be the main API going forward as it can migrate off of the `scrollInfo` fallback over time.

packages/framer-motion/src/utils/__tests__/delay.test.ts
```diff
@@ -16,7 +16,7 @@ describe("delay", () => {
     test("provides overshoot duration", async () => {
         return new Promise<void>((resolve) => {
             delay((overshoot) => {
-                expect(overshoot).toBeLessThan(20)
+                expect(overshoot).toBeLessThan(50)
```
adamseckel, line 19: Why did this change?

mattgperry: Oh flaky test

adamseckel (approved):
Code LGTM (strange browser api choices aside 🙈)

What's your plan for rolling this out? It's a breaking change since the `scroll` export is now quite different. Do we need a plan to address that in Framer where there may be user code built on existing `scroll` behaviour?

PR #2259, opened 2023-07-27
Hardware accelerated scroll animations

mattgperry:


packages/framer-motion/src/render/dom/scroll/observe.ts
```diff
@@ -0,0 +1,32 @@
...
+            update(progress)
+        }
+
+        prevProgress = progress
+    }
+
+    frame.update(onFrame, true)
+
+    return () => {
+        console.log("cancel timeline observation")
```
adamseckel, line 29: Shall we drop this before merging?

adamseckel (approved):

packages/framer-motion/src/render/dom/scroll/observe.ts
mattgperry: ty

PR #2265, opened 2023-07-29
`transitionEnd` should not be applied when new animations caused the animation to stop

chuganzy:
This PR aims to fix an issue where `transitionEnd` styles are applied after new animation starts and accidentally overrides the latest motion values.

When new animation starts while animating, `MotionValue` stops the ongoing animation and immediately starts the new one. However as `transitionEnd` waits for all animations to be resolved, some `transitionEnd` styles overrides new values.

It seems to be introduced in https://github.com/framer/motion/pull/2025. Before it lands, calling `stop()` never resolves animation Promises, which are used to apply the `transitionEnd` styles at the end, so there was no race condition..

Closes: https://github.com/chakra-ui/chakra-ui/issues/7775

### DIFF
--- a/packages/framer-motion/src/animation/interfaces/visual-element-target.ts
+++ b/packages/framer-motion/src/animation/interfaces/visual-element-target.ts
@@ -43,6 +43,7 @@ export function animateTarget(
     if (transitionOverride) transition = transitionOverride
 
     const animations: AnimationPlaybackControls[] = []
+    const promises: (Promise<void>)[] = []
 
     const animationTypeState =
         type &&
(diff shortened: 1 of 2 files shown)


mattgperry:
Can you add a test for this behaviour? It also looks like this would stop all of `transitionEnd` values being applied if even any one of the values is animating.

chuganzy:
@mattgperry Sure, added a test and fixed the logic to check if animations completed vs canceled [here](https://github.com/framer/motion/pull/2265/commits/9011c7d680cbb1cd7efc0922673430399bbc1536)

jonahallibone:
Any update on this PR? I believe I am experiencing the issue this PR would fix

stefanweck:
This fix would resolve https://github.com/framer/motion/issues/2317. Any update on merging this fix?

jonahallibone:
Bumping this, any update Framer team?

AndreiRailean:
Hey @mattgperry what do you think needs to happen to get this PR merged?
I assume at a minimum it needs to be brought up-to-date with master. Is there anything else?

TranslucentDragon:
Would like to bump this as well. I came here from the earlier linked accordion chakra UI issue. What else needs to happen to get this merged? @mattgperry

jonahallibone:
It seems like this isn't a problem in the newest versions of framer motion. I haven't noticed it in in the most recent couple versions. Can anyone else confirm?

PR #2273, opened 2023-08-03
Fix: add useScroll offset param to dependencies array

### LINKED ISSUE #2272: [BUG] `useScroll` doesn't update when `offset` is updated
`useScroll` doesn't update when `offset` param is updated

```javascript
const DELTA = 100;
const targetRef = useRef();
const headerRef = useRef();
const headerHeight = useComponentSize(headerRef).height;
const { scrollYProgress } = useScroll({
  target: ref,
  offset: [`start ${headerHeight + DELTA}px`, `start ${headerHeight}px`],
});
``` 

in this case when `headerHeight` changes `scrollYProgress` should change but `useScroll` doesn't notice that change due to empty dependency array in useScroll code. 
https://github.com/framer/motion/blob/main/packages/framer-motion/src/value/use-scroll.ts#L60
a quick fix could be stringify the offset param and pass it to the dependency array

```typescript
export function useScroll({
    container,
    target,
    layoutEffect = true,
    ...options
}: UseScrollOptions = {}) {
    const values = useConstant(createScrollMotionValues)

    const useLifecycleEffect = layoutEffect
        ? useIsomorphicLayoutEffect
        : useEffect

    useLifecycleEffect(() => {
        refWarning("target", target)
        refWarning("container", container)

        return scrollInfo(
            ({ x, y }) => {
                values.scrollX.set(x.current)
                values.scrollXProgress.set(x.progress)
                values.scrollY.set(y.current)
                values.scrollYProgress.set(y.progress)
            },
            {
                ...options,
                container: container?.current || undefined,
                target: target?.current || undefined,
            }
        )
    }, [JSON.stringify(options.offset)])

    return values
}
```

kalimantos:
Fixes #2272
add useScroll offset param to useEffect dependencies array to allow offset change

### DIFF
--- a/packages/framer-motion/src/value/use-scroll.ts
+++ b/packages/framer-motion/src/value/use-scroll.ts
@@ -57,7 +57,7 @@ export function useScroll({
                 target: target?.current || undefined,
             }
         )
-    }, [])
+    }, [JSON.stringify(options.offset)])
 
     return values
 }


mattgperry:
Great, thanks!

PR #2289, opened 2023-08-11
Batch SVG read/writes

mattgperry:
SVG `transform-origin` is always relative to the parent `svg` element rather than the SVG element itself. To normalise this between HTML and SVG we measure the SVG on mount. This PR batches reads and writes of this process to reduce layout thrashing.

### DIFF
--- a/dev/examples/SVG-transform.tsx
+++ b/dev/examples/SVG-transform.tsx
@@ -1,10 +1,6 @@
 import * as React from "react"
 import { motion } from "framer-motion"
 
-/**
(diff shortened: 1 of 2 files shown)


dev/examples/SVG-transform.tsx
```diff
@@ -0,0 +1,42 @@
+import * as React from "react"
+import { motion } from "framer-motion"
+
+/**
+ * An example of providing a MotionValue to an SVG component via its props
```
adamseckel, line 5: Where's the motion value?

PR #2293, opened 2023-08-11
Add queue for animation when visualEl is deferred

### LINKED ISSUE #2292: Delayed visualElement subscription can miss animations in using LazyMotion
When utilizing the LazyMotion component, there's a scenario where the visualElement might not be immediately available for the animation controls. In the original animationControls code, animations get started without ensuring the presence of any subscribed visualElement, potentially causing missed animations.

**Steps to reproduce:**

- Integrate the LazyMotion component into a project.
- Use m-Components.
- Call controls.start() before the visualElement has subscribed - in the effect.
- Observe that animations may be skipped or missed.

```
Lazy-Load.ts
export { domAnimation as default } from 'framer-motion';
```

```
import { m, LazyMotion, useAnimation, useMotionValue } from 'framer-motion';
import { FC, useEffect } from 'react';

// only way to keep bundle size small
const loadFeaturesAsync = async () =>
  await import('./Lazy-Load').then(res => res.default);

// it works when features are loaded synchronously like, but bundle size increases
// import { domAnimation } from 'framer-motion';
// <LazyMotion features={domAnimation}>

export const AnimationTest: FC = () => {
  const controls = useAnimation();

  const variants = {
    foo: { x: 100 }
  };

  const childVariants = {
    foo: { backgroundColor: '#fff' }
  };

  const x = useMotionValue(0);
  const backgroundColor = useMotionValue('#000');

  useEffect(() => {
    controls.start('foo').then(() => {
      console.log('variants result: ', [x.get(), backgroundColor.get()]);
    });
  }, []);

  return (
    <LazyMotion features={loadFeaturesAsync}>
      <m.div
        animate={controls}
        variants={variants}
        style={{ x }}
        onAnimationStart={() => {
          console.log('onAnimationStart outer');
        }}
        onAnimationComplete={() => {
          console.log('onAnimationComplete outer');
        }}
      >
        <m.div
          variants={childVariants}
          style={{ backgroundColor }}
          onAnimationStart={() => {
            console.log('onAnimationStart inner');
          }}
          onAnimationComplete={() => {
            console.log('onAnimationComplete inner');
          }}
        >
          Hello
        </m.div>
      </m.div>
    </LazyMotion>
  );
};
```

Here is the sandbox https://codesandbox.io/p/sandbox/polished-sunset-pg25zr

**Expected Behavior:**
Animations should always be played regardless of **when** the visualElement subscribes.

**Actual Behavior:**
Animations might be skipped if started before the visualElement subscription.

**Suggested Fix:**
Improve animationControls to queue animation start calls and processes them once the visual element is available, see referenced PR

gurkerl83:
**What's changed?**
This pull request introduces changes to the animationControls function to handle scenarios where the visualElement might not immediately subscribe, especially in the context of the LazyMotion component.

**Introduced Start Queue:** An array that temporarily holds animation start calls. This ensures animations are deferred until their respective visual elements have subscribed to the animation controller.

**Flush Mechanism:** Once a visualElement subscribes, any queued animations are immediately processed and played.

**Error Handling:** Added robust error handling to cater to potential issues during the animation process. This ensures promises are either resolved upon successful completion or rejected in case of errors.

**Benefits:**
This change ensures animations are consistently played in scenarios involving delayed visualElement subscriptions, enhancing user experience and ensuring predictable behavior.

**How to test:**

Integrate the LazyMotion component, described in #2292

- Integrate the LazyMotion component into a project.
- Import features dynamically to keep bundle size small.
- Use m-Components.
- Call controls.start() before the visualElement has subscribed - in the effect.

Ensure that animations play consistently, irrespective of when the controls.start() is called in relation to the visualElement subscription.

Fixes #2292

### DIFF
--- a/packages/framer-motion/src/animation/hooks/animation-controls.ts
+++ b/packages/framer-motion/src/animation/hooks/animation-controls.ts
@@ -1,8 +1,27 @@
 import { invariant } from "../../utils/errors"
 import { setValues } from "../../render/utils/setters"
 import type { VisualElement } from "../../render/VisualElement"
-import { AnimationControls } from "../types"
+import type {
+    AnimationControls,
+    AnimationDefinition,
+    Transition,
+} from "../types"
 import { animateVisualElement } from "../interfaces/visual-element"
+
+/**
+ * Represents an individual animation start call.
+ * Each item encapsulates:
+ *  - `definition`: The details of the desired animation.
+ *  - `transitionOverride` (optional): Overrides for default transition settings.
+ *  - `resolve`: Promise resolution function, invoked when the animation completes successfully.
+ *  - `reject`: Promise rejection function, invoked upon animation error or failure.
+ */
+type StartQueueItem = {
+    definition: AnimationDefinition
+    transitionOverride?: Transition
+    resolve: (value?: any) => void
+    reject: (reason?: any) => void
+}
 
 function stopAnimation(visualElement: VisualElement) {
     visualElement.values.forEach((value) => value.stop())
@@ -22,9 +41,19 @@ export function animationControls(): AnimationControls {
      */
     const subscribers = new Set<VisualElement>()
 
+    /**
+     * `startQueue` is an array that temporarily holds animation start calls, ensuring they are
+     * deferred until their respective visual elements have subscribed to the animation controller.
+     */
+    const startQueue: Array<StartQueueItem> = []
+
     const controls: AnimationControls = {
         subscribe(visualElement) {
             subscribers.add(visualElement)
+
+            // Upon a new subscription, handle any queued animations.
+            flushStartQueue()
+
             return () => void subscribers.delete(visualElement)
         },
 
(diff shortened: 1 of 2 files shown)


mattgperry:
We actually used to have something like this in `AnimationControls` for when people in Framer would call `.start()` inside the render function body.

My concern with this, it might work ok in instances where the features are expected to arrive shortly after we trigger the animations but conceivably they could load a long time after animations are triggered at which point, are the animations still relevant? Would it be better to instantly finish animations until this? I'm not sure.

gurkerl83:
@mattgperry 

This is what you mentioned, right https://github.com/framer/motion/commit/fc2dda0d980ab4c74225713c3a88a8b783e89c5d 

**The key difference from the previous implementation**, see the commit mentioned and the current PR, **is that framer-motion subscribe triggers the start of queued animations, not a react lifecycle hook**.

**New implementation**
1. **Animation control "start"** adds pending animations to the queue if no subscribers are registered yet
2. **Pending animations get started when the function "subscribe" executes**. When framer motion has prepared the respective visualElement (meaning also lazy feature loading is finished) - the animation feature - triggers the subscribe call; "pending" animations get started.

**Difference to the previous behavior**
1. Animation control "start" adds pending animations to the queue, but if not yet mounted. Note: **The current PR** implementation **looks for a reaction of framer-motion as described above, not an outside (React dependence)**
2. **Mount starts pending animations.** This **does not guarantee** that visualElement is complete and that the lazy feature loading has been completed. This is because the mount gets called from a react side effect, which does not know anything about the status of visualElement and framer-motion internals.

**About your questions**
_Would it be better to instantly finish animations until this?_
I am not sure if I understand correctly, but the visualElement is not there yet, and the same goes for the animation feature, which triggers subscribe.

**So my answer might be:** 
Animations have not even started because subscribe has never fired.

_Are the animations still relevant?_
I think so because, for the same reason as above, animations are not executed yet; you can not skip the animation and go directly to the end state of an animation right?

Thx!

gurkerl83:
@mattgperry Maybe you can find a bit of time to think about the provided concept again? Would be awesome if this or something similar could land somehow. Thx!

mattgperry:
Closing this for now as `useAnimationControls()` is deprecated so in that sense it isn't technically possible to run animations before a component has loaded.

PR #2300, opened 2023-08-15
Add `computed`-inspired API for `useTransform`

mattgperry:
Inspired by Preact's Signals API. Instead of declaring `MotionValue`s, i.e:

```javascript
// Single values
useTransform(x, (latest) => latest * 2)

// Multiple values
useTransform([x, y], ([latestX, latestY]) => latestX + latestY)
```

This API allows `MotionValue`s to be accessed within the callback via `.get()`, and these will be subscribed to automatically.

```javascript
// Single values
useTransform(() => x.get() * 2)

// Multiple values
useTransform(() => x.get() + y.get())
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,11 +4,21 @@ Framer Motion adheres to [Semantic Versioning](http://semver.org/).
 
 Undocumented APIs should be considered internal and may change without warning.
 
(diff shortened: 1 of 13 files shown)


adamseckel (approved):
coooooool

PR #2311, opened 2023-08-23
Use type instead of interface to avoid declaration merging

leroykorterink:
The `RefObject` interface declaration in `framer-motion` is merged with the one in `react` this can cause issues if other packages (e.g. `styled-components`) use the original `react` types.

> MyComponent.styles.tsx:7:14 - error TS4023: Exported variable 'MyComponent' has or is using name 'RefObject' from external module "/my-project/node_modules/framer-motion/dist/index" but cannot be named.
>
> 7: export const MyComponent = styled(motion.div)``;

This issue can easily be resolve if we use types instead of interfaces in `framer-motion`.


mattgperry:
Interesting, thanks for the fix!

PR #2322, opened 2023-08-30
Improve performance of frameloop

mattgperry:
As the `length` of an array grows, the performance of `indexOf` degrades. This makes it more expensive to schedule new jobs as queues grow. This PR switches lookup to a `Set` which has a fixed lookup cost.

A further optimisation at a later date could be to remove the lookup entirely by removing the assumption that a callback can't be called twice.


packages/framer-motion/src/frameloop/render-step.ts
```diff
@@ -1,16 +1,39 @@
...
+            return true
+        }
+    }
+
+    remove(process: Process) {
+        const index = this.order.indexOf(process)
+        if (index !== -1) {
+            this.order.splice(index, 1)
+            this.scheduled.delete(process)
+        }
```
shuangq, line 20: I wonder if we should do a `this.scheduled.has` check first so we only run the indexOf lookup when needed, but if we're removing the lookup entirely soon then it probably doesn't matter as much.

shuangq (approved):

packages/framer-motion/src/frameloop/render-step.ts
mattgperry: I think also remove happens so rarely it's probably not worth the check, whereas schedule can happen thousands of times a frame

PR #2325, opened 2023-08-31
Reducing calls to frameloop

mattgperry:
Also makes a change that can create fewer objects per frame.

### DIFF
--- a/packages/framer-motion/src/projection/node/create-projection-node.ts
+++ b/packages/framer-motion/src/projection/node/create-projection-node.ts
@@ -664,8 +664,12 @@ export function createProjectionNode<I>({
             this.sharedNodes.forEach(removeLeadSnapshots)
         }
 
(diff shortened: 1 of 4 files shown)


packages/framer-motion/src/projection/node/create-projection-node.ts
```diff
@@ -1741,16 +1747,18 @@ export function createProjectionNode<I>({
...
                 if (this.hasProjected && !hasTransform(this.latestValues)) {
                     emptyStyles.transform = transformTemplate
                         ? transformTemplate({}, "")
                         : "none"
                     this.hasProjected = false
                 }
+                console.log("returning empty")
                 return emptyStyles
             }
+            console.log("returning prpojhection")
```
shuangq, line 1761: This needs to be cleanup?

shuangq (approved):

Issue #2371, opened 2023-10-18
[BUG] Node v21+ implements global navigator without userAgent prop, which breaks SSR


apostolos:
Recently released Node v21 partially implements the Navigator interface, however it doesn't include the userAgent property.

This breaks code like:
https://github.com/framer/motion/blob/main/packages/framer-motion/src/projection/node/create-projection-node.ts#L2099-L2101

### Release notes:
https://nodejs.org/en/blog/announcements/v21-release-announce#navigator-object-integration

### Repro:
Render a component with framer-motion with a SSR framework like Next.js using Node v12

### Suggested fix:
Test for `window` before touching the `navigator` object or additional check if `userAgent !== undefined`?

picasocro1:
I have the same issue.

@apostolos I'm only wondering whether in your fix it should be an additional casting to the boolean, to keep the consistency of the function result (always boolean, not boolean | undefined)

```
const userAgentContains = (string: string) =>
    typeof navigator !== "undefined" &&
    !!navigator.userAgent?.toLowerCase().includes(string)
```

Uzlopak:
In node 21.1.0 navigator has now userAgent. Is your issue now solved?

apostolos:
@Uzlopak It is!

Confirmed fixed in Node v21.1.0

hirbod:
We bumped the framer-motion resolution to v10.16.5 (used by Moti, a React-Native library).
The undefined check is not sufficient for React Native, as navigator is defined but navigator.userAgent is not.

![simulator_screenshot_59501023-3208-4B41-9053-9DFAD95900F9](https://github.com/framer/motion/assets/504909/59eda410-7f69-45cf-ad7e-ca2e09bdbfbc)

Uzlopak:
@hirbod 
Update to node 21.1.0

hirbod:
@Uzlopak node is not related to React Native.

Uzlopak:
Then this issue has nothing to do with the problem you reported. So please open a new issue.

hirbod:
The issue is exactly related to it, since its the same code that breaks. RN resolves to the CJS part, but its unrelated to node.
Here is a patch file in case anybody else encounters this.

```patch
diff --git a/node_modules/framer-motion/dist/cjs/index.js b/node_modules/framer-motion/dist/cjs/index.js
index bd8dc68..8c1397d 100644
--- a/node_modules/framer-motion/dist/cjs/index.js
[...]

PR #2375, opened 2023-10-20
Add optional chaining to navigator.userAgent

TerminalStar:
Closes https://github.com/framer/motion/issues/2371

### DIFF
--- a/packages/framer-motion/src/projection/node/create-projection-node.ts
+++ b/packages/framer-motion/src/projection/node/create-projection-node.ts
@@ -2098,7 +2098,7 @@ const defaultLayoutTransition = {
 
 const userAgentContains = (string: string) =>
     typeof navigator !== "undefined" &&
-    navigator.userAgent.toLowerCase().includes(string)
+    navigator.userAgent?.toLowerCase().includes(string)
 
 /**
  * Measured bounding boxes must be rounded in Safari and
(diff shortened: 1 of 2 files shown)


ziishaned (approved):

rmill2016:
Please merge 😭

TerminalStar:
@mattgperry can you help to get this merged in? 🙏

rmill2016:
> @mattgperry can you help to get this merged in? 🙏

Someone else on another thread recommends downgrading Node version to resolve this temporarily, have you tried this yet?

TerminalStar:
@rmill2016 yes, downgrading works but I came across this as we were looking upgrade to v21 to fix a security vulnerability. Node v21.1.0 has [added the `userAgent` property back](https://github.com/nodejs/node/releases/tag/v21.1.0) so we can upgrade to that version without this change.

That said, I would still support it being merged in as it makes the code more resilient to being run in environments with different shapes of `navigator` object - notably this is the only dependency of ours that doesn't somehow check for `userAgent` before accessing it. But given I can bump the node version in my project now, I don't feel strongly about it and the maintainers can do as they wish 🤷‍♀️

mattgperry:
I'm just going to close this out as this is specced so **should** (imo) be treated as a bug from the implementor.

PR #2414, opened 2023-11-24
fix: correctly type reorder group and item generics

victorboucher:
The use of `forwardRef` caused the generic of the `Reorder.Group` and `Reorder.Item` components to always get inferred as unknown. Explicitly typing it like this solves that issue.


mattgperry:
Great contribution, thanks!

PR #2419, opened 2023-11-29
Ensure optimised animations don't run after hydration

mattgperry:
In Safari it can be that the optimised appear animation script runs after hydration. This PR ensures optimised animations won't trigger in this situation.

### DIFF
--- a/packages/framer-motion/src/animation/optimized-appear/types.ts
+++ b/packages/framer-motion/src/animation/optimized-appear/types.ts
@@ -1,17 +1,19 @@
 import { Batcher } from "../../frameloop/types"
 import { MotionValue } from "../../value"
 
+type HandoffFunction = (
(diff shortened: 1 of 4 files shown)


packages/framer-motion/src/animation/optimized-appear/types.ts
```diff
@@ -1,17 +1,19 @@
...
  */
 declare global {
     interface Window {
-        HandoffAppearAnimations?: (
-            storeId: string,
-            valueName: string,
-            value: MotionValue,
-            sync: Batcher
-        ) => number
+        HandoffAppearAnimations: undefined | false | HandoffFunction
```
shuangq, line 17: Shouldn't this still be an optional key?

shuangq (approved):

PR #2422, opened 2023-11-29
Add a function to disable instant transitions

nvh:
This PR adds a function to manually reset the instant animation state back to false. We need this in the preview of FramerStudio to prevent animations from accidentally being disabled.


packages/framer-motion/src/utils/use-instant-transition.ts
```diff
@@ -37,3 +37,7 @@ export function useInstantTransition() {
         })
     }
 }
+
+export function disableInstantTransitions() {
```
nvh, line 41: I thought it made more sense to call this `disableInstantTransitions` that `enableAnimations` because I thought that would be a bit too broad as a public API and might be misinterpreted

mattgperry: Totally agree

mattgperry (approved):

PR #2426, opened 2023-11-30
Fixing handoff when paired with portal

mattgperry:
Currently, if a tree rendered into a portal mounts before the main tree, optimised animations will fail to handoff and will re-run from start.

This moves the point of deleting `HandoffAppearAnimations` from "any effect" to "when we have handed animations off"


packages/framer-motion/src/motion/utils/use-visual-element.ts
```diff
@@ -87,7 +87,6 @@ export function useVisualElement<Instance, RenderState>(
          * so components added after the initial render can animate changes
          * in useEffect vs useLayoutEffect.
          */
-        window.HandoffAppearAnimations = false
```
shuangq, line 90: Let's move/cleanup the comment above as well

shuangq (approved):
Nice one!

PR #2428, opened 2023-11-30
Skip more needless animations

mattgperry:
Adds a further check for when target is provided as a keyframes list

### DIFF
--- a/packages/framer-motion/src/animation/interfaces/visual-element-target.ts
+++ b/packages/framer-motion/src/animation/interfaces/visual-element-target.ts
@@ -92,14 +92,18 @@ export function animateTarget(
                 visualElement.getProps()[optimizedAppearDataAttribute]
 
             if (appearId) {
-                canSkipHandoff = false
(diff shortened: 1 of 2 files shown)


packages/framer-motion/src/animation/interfaces/visual-element-target.ts
```diff
@@ -79,18 +92,22 @@ export function animateTarget(
...
+                if (elapsed) {
+                    canSkipHandoff = false
+                    valueTransition.elapsed = elapsed
+                    ;(valueTransition as Transition).syncStart = true
+                }
             }
         }
 
-        let canSkip = canSkipHandoff && valueTarget === value.get()
+        let canSkip = canSkipHandoff && !hasKeyframesChanged(value, valueTarget)
```
adaam2, line 110:
```suggestion
        let canSkip = canSkipHandoff && !haveKeyframesChanged(value, valueTarget)
```

PR #2431, opened 2023-12-01
Wrap safeToRemove function with useCallback hook

baeharam:
**Why?**

Most React users use [lint for hook](https://github.com/facebook/react/blob/main/packages/eslint-plugin-react-hooks/README.md) which shows error when deps of `useEffect` is missed. So, `safeToRemove` function need to be inserted into `useEffect` when using `useAnimate` hook. Even if framer guide shows example code that is not including `safeToRemove` into deps of `useEffect`, react user want to solve lint error.

**Solution**

Therefore, `safeToRemove` should ensure that it keeps reference of same function in same component. If this is intended implementation, I would appreciate it if you explain the reason.

Thanks for reading and your great library!


mattgperry:
Makes sense, sorry for the delay, and thanks for the PR!

PR #2432, opened 2023-12-01
Optimised appear animation improvements

mattgperry:
This PR makes two improvements to optimised appear animations.

1. Use a single animation to detect document paint ready. This reduces the number of instantiated `ready` pre-animations from `n` values to `1`.
2. Only measure time via `performance.now()` once each for starting optimised animations and handing off optimised animations.
   - `performance.now()` is different every call whereas animations are frame-locked. Normally calling `element.animate()` and later measuring the `startTime`, all animations will have started on the same time. We are manually setting `startTime` so this should be the same across all animations. This will help animations stay in sync when many are started at once.


packages/framer-motion/src/animation/optimized-appear/start.ts
```diff
@@ -24,40 +39,43 @@ export function startOptimizedAppearAnimation(
...
+        )
 
-    appearAnimationStore.set(storeId, {
-        animation: readyAnimation,
-        startTime: null,
-    })
+        appearAnimationStore.set(storeId, {
+            animation: readyAnimation,
+            startTime: null,
+        })
```
shuangq, line 57: Not introduced in this PR, but why do we need to store the dummy animation in the appearAnimationStore?

shuangq (approved):
Code looks good! Left a minor question:

PR #2435, opened 2023-12-04
Fix margin type in InViewOptions interface

Bobokhuja:
Create MarginType for property "margin" in InViewOptions interface


mattgperry (changes requested):
Typo but looks good to me otherwise

Bobokhuja:
> Typo but looks good to me otherwise

Excuse me, can you tell me where exactly? Because I don't really understand English

NunesClement:
> > Typo but looks good to me otherwise
> 
> Excuse me, can you tell me where exactly? Because I don't really understand English

There is just one typo, you wrote: MarinType instead of MarginType

PR #2437, opened 2023-12-05
Improve optimised appear scheduling

mattgperry:
This PR ensures that only the first `animateChanges` runs in `useLayoutEffect` with a new variable, `HandoffComplete`, that avoids the race condition between not allowing `startOptimisedAppear` to run after hydration, and handoff working even if another React tree has been mounted first (like the Made in Framer button)


shuangq (approved):
LGTM!

PR #2440, opened 2023-12-07
Optimised appear animations V2

mattgperry:
This PR changes the handoff strategy of optimised appear animations.

## Before
Screenshot 2023-12-08 at 12 19 36

Currently, Framer runs hardware-accelerated `opacity` and `transform` animations via WAAPI before React hydrates. Then, when React hydrates, Framer Motion transfers `opacity` to a WAAPI animation and `transform` to the individual transform JS animations. This way, these animations can be interrupted independently via gestures etc as usual.

The problem with this approach is the precise timing required to seamlessly transfer WAAPI animations is difficult to get right and the accuracy of reported timings is different across browsers.

Therefore, on big pages animations can appear to jump and, once on the main thread, stutter, during this period of uniquely heavy JS initialisation.

## After

This PR changes the strategy to preserve the optimised WAAPI animations for as long as possible.

Screenshot 2023-12-08 at 12 21 12

When we handoff, we start new JS animation for `opacity` and individual transforms, same as today. The difference is we don't cancel the WAAPI animations. WAAPI/CSS animations override any changes to an element's underlying computed styles so these JS animations aren't visible.

Only when a value needs to perform a different animation do we cancel the WAAPI animation. This way, we preserve the capability to interrupt transforms individually while also running an optimised animation for longer.

Screenshot 2023-12-08 at 12 22 24

Because all transforms are animating via JS, cancelling the WAAPI animation from any single transform will simply "reveal" the underlying animation, which is a tidy way of allowing any one transform to interrupt the unified `transform` animation.

In the majority of cases, the optimised animation will not be interrupted so we will now see an accelerated animation throughout the appear animation.

In the minority of cases where this animation is interrupted and we might have perceived a small jump or stutter, this will be improved because it is less noticeable to handoff to a different animation than an identical one.


packages/framer-motion/src/animation/interfaces/motion-value.ts
```diff
@@ -118,6 +118,7 @@ export const animateMotionValue = (
          * Animate via WAAPI if possible.
          */
         if (
+            !transition.isHandoff &&
```
shuangq, line 121: Let's add a comment here to explain why we want to animate individual transforms when it's a handoff one.

PR #2445, opened 2023-12-11
Add type for margin prop in InViewOptions interface

Bobokhuja:


mattgperry:
Very nice, thanks for the PR!

PR #2447, opened 2023-12-13
Remove support for `transformValues` prop

mattgperry:


nvh (approved):
This is still being used in [FrameWithMotion](https://github.com/framer/FramerStudio/blob/10dc94a5599c0dae4a966daa3af45e5034847a6c/src/library/src/render/presentation/Frame/FrameWithMotion.tsx#L430), so this potentially breaks old components that are still importing `Frame` from "framer" _and_ are using `size`, `radius`, or `shadow` or a color value for one of the props.

We should be a bit carefull with this, and at least make sure that all the instances of `FramerWithMotion` used in our codebase _and_ the components in the insert menu don't use these API's

nvh:
Uses of FrameWithMotion in FramerStudio:
Screenshot 2023-12-21 at 14 18 43

Issue #2339, opened 2023-09-12
[BUG] When using css variables for colors no smooth transition is created


alina-boshkov:
**1. Description**
When using keyframe values for background color, regular color values work as expected and have a smooth transition between them. 
But when updating to use css variables the color switch is instant without smooth transitioning. 

**2. [CodeSandbox](https://codesandbox.io/s/framer-motion-css-vars-issue-zs2yz3?file=/src/App.js)

**3. Steps to reproduce**
1. create a motion element. 
2. create css vars for colors to transition from and to. 
3. add these to the animate. 
4. check the animation.
```
	<motion.div
      animate={{
        // background color switch is smooth
        // background: ["#565464", "#ff0083"],


        // background color switch isn't smooth
        background: ["var(--first-color)", "var(--second-color)"],
        transition: { delay: 1.4 }
      }}
      className="box"
    />
```

**4. Expected behavior**
css variables to have smooth transition the same as regular color values.

**5. Video**
https://github.com/framer/motion/assets/57999214/3eef5d96-056b-4958-9873-c31771fef71a

180107072:
```code
      <motion.div
        initial={{
          backgroundColor: "var(--first-color)"
        }}
        animate={{
          backgroundColor: "var(--second-color)",
          transition: { delay: 1.4 }
        }}
        className="box"
      />

Issue #2309, opened 2023-08-22
[BUG] animate() in an effect is not properly reading CSS Variables


samselikoff:
**Describe the bug**

It seems Framer Motion is having trouble reading CSS Variables when using `animate()` in an effect:

```jsx
useEffect(() => {
  animate(ref.current, {
    // This doesn't work
    opacity: ["var(--opacity-from)", "var(--opacity-to)"]
  });
}, [animate, ref, count]);
```

**CodeSandbox**

https://codesandbox.io/s/framer-motion-css-variable-bug-5vkfws

1. Click button, see message fade
2. Comment out line 12, and uncomment line 15 (uses CSS variables)
3. Click button. Message doesn't fade.

**Expected behavior**

The animation should trigger using the values from the CSS Variables.

rudzky:
any update on this? 😅

Issue #2168, opened 2023-05-31
[BUG] If you use the css variable in the background color to animate it, it may not work at some circumstance.


sukjae:
**Describe the bug**

If you use the css variable in the background color to animate it, it may not work at some circumstance.

**IMPORTANT: Provide a CodeSandbox reproduction of the bug**

https://codesandbox.io/s/framer-motion-background-animatable-ilqf6x?file=/src/App.js

```jsx
<motion.div
  id="first"
  style={{
    width: 100,
    height: 100,
    backgroundColor: "#ff4400",
    "--adaptiveBlue200": "#8b95a1"
  }}
  animate={{
    backgroundColor: "#8b95a1" // working
    // backgroundColor: ["#ff4400", "#8b95a1"] // working
    // backgroundColor: [null, "#8b95a1"] // working
    // backgroundColor: "var(--adaptiveBlue200, none)" // working
    // backgroundColor: [null, "var(--adaptiveBlue200, #fff)"] // error: 'var(--adaptiveBlue200, #fff)' is not an animatable color. Use the equivalent color code instead.
  }}
  transition={{
    duration: 3
  }}
/>
```

**Steps to reproduce**
1. animate backgroundColor with array
2. first item of array should be null
3. second item of array should be css variable

from [this code](https://github.com/framer/motion/blob/main/packages/framer-motion/src/utils/mix-color.ts#L17), only allows hex, rgba, hsla for colorTypes. Perhaps add css variable?

PR #2448, opened 2023-12-14
Defer keyframe resolution

mattgperry:
When keyframes aren't inherently mixable, we make them mixable by measuring the DOM or resolving values from it.

```javscript
animate(element, {
  height: "auto", // What is auto?
  width: ["50%", "100px"], // How do these mix?
  x: [null, 100], // What is the initial keyframe?
  background: "var(--tint)" // What color is this?
})
```

This is currently done synchronously per `animate()` call / `motion` component animation which can lead to layout thrashing when many are performed and/or when mixed in amongst other layout-measuring effects.

Screenshot 2024-02-27 at 10 50 19

## Solution

This PR defers keyframe resolution until the following animation frame. All reads and writes are batched so we have at max 1 style recalculation/ layout reflow.

It also refactors JS and WAAPI animations into a `MainThreadAnimation` and `AcceleratedAnimation` class inheriting from `BaseAnimation` to share as much of this new logic as possible.

Animations still return synchronously-available animation controls like `.play()` / `.time` etc. As much of this is computed or otherwise relies on resolved keyframes, they internally use `this.resolved` which is behind a getter. This getter will synchronously flush all pending animations as a de-opt when accessed. Some methods have already been changed to allow pending operations but this can be improved further. In the absolute worst case, where animations are created/accessed

```
const a = animate()
a.time

const b = animate()
b.time
```

Performance is the same as today. But in the best case, cold start performance is 2.5x faster and unit conversion performance is 6x faster, and much of that work is lifted out of the effect and into the next animation frame.

Screenshot 2024-02-27 at 10 56 39

Bonus:
- Support unit conversion and CSS variables in keyframes

Fixes https://github.com/framer/motion/issues/2168
Fixes https://github.com/framer/motion/issues/2309
Fixes https://github.com/framer/motion/issues/2339


dev/benchmarks/cold-start-anime.html
```diff
@@ -28,15 +28,15 @@
             }
```
mattgperry, line 28: Can ignore the `/benchmarks`

packages/framer-motion/cypress/integration/drag-to-reorder.ts
```diff
@@ -201,7 +201,7 @@ describe("Drag to reorder", () => {
                     const y = step > 0 ? delta : -delta
                     chain = chain
                         .trigger("pointermove", 360, baseY + y, { force: true })
-                        .wait(50)
+                        .wait(100)
```
mattgperry, line 204: Flakey test

packages/framer-motion/src/animation/__tests__/animate-waapi.test.ts
```diff
@@ -12,6 +13,8 @@ describe("animate() with WAAPI", () => {
             { duration: 1, transform: { duration: 2 } }
         )
 
+        await nextFrame()
```
mattgperry, line 16: WAAPI `nextFrame` calls generally added as the WAAPI animation is only initialised after keyframes are resolved. We could add more custom logic here to say if something is WAAPI we can resolve fewer keyframes as it can handle CSS vars etc itself

packages/framer-motion/src/animation/hooks/animation-controls.ts
```diff
@@ -1,13 +1,41 @@
...
 import type { VisualElement } from "../../render/VisualElement"
-import { AnimationControls } from "../types"
+import { AnimationControls, AnimationDefinition } from "../types"
 import { animateVisualElement } from "../interfaces/visual-element"
 
 function stopAnimation(visualElement: VisualElement) {
     visualElement.values.forEach((value) => value.stop())
 }
 
+function setVariants(visualElement: VisualElement, variantLabels: string[]) {
```
mattgperry, line 11: Moved in from `utils/setters` as only used in this file.

packages/framer-motion/src/animation/interfaces/motion-value.ts
```diff
@@ -36,32 +39,10 @@ export const animateMotionValue = (
         let { elapsed = 0 } = transition
         elapsed = elapsed - secondsToMilliseconds(delay)
 
-        const keyframes = getKeyframes(
```
mattgperry, line 39: Moved to `can-animate`

packages/framer-motion/src/render/utils/animation-state.ts
```diff
@@ -212,8 +214,9 @@ export function createAnimationState(
              * Build an object of all the resolved values. We'll use this in the subsequent
              * animateChanges calls to determine whether a value has changed.
              */
+            // TODO Resolve with options
```
adamseckel, line 217: Does this TODO need addressing before merge? Or is it no longer relevant?

packages/framer-motion/src/render/svg/SVGVisualElement.ts
```diff
@@ -33,6 +33,7 @@ export class SVGVisualElement extends DOMVisualElement<
     }
 
     readValueFromInstance(instance: SVGElement, key: string) {
+        // console.log("reading", key, "from", instance)
```
adamseckel, line 36:
```suggestion
```

packages/framer-motion/src/render/html/utils/make-none-animatable.ts
```diff
@@ -0,0 +1,33 @@
...
+
+    if (animatableTemplate && name) {
+        for (const noneIndex of noneKeyframeIndexes) {
+            unresolvedKeyframes[noneIndex] = getAnimatableNone(
+                name,
+                animatableTemplate
+            )
+        }
+    }
+}
```
adamseckel, line 33: I'm having a hard time understanding what this is doing. What is the template, just some keyframe value like a color string or numeric value?

packages/framer-motion/src/render/utils/animation-state.ts
mattgperry: Removed, I think this was an idle thought

packages/framer-motion/src/gestures/hover.ts
```diff
@@ -20,7 +19,7 @@ function addHoverEvent(node: VisualElement<Element>, isActive: boolean) {
         }
 
         if (props[callbackName]) {
-            frame.update(() => props[callbackName](event, info))
+            props[callbackName](event, info)
```
adamseckel, line 22: Is this an unrelated/drive-by change to no longer defer event handlers?

mattgperry: It's actually entwined because without doing this we can add a two-frame delay between interactions and paint (which is counter to the point of this) as we have the frame for the event then the frame to resolve keyframes

adamseckel: Ah!

adamseckel (commented):
I wonder if before merging this, it would be worthwhile to create a FramerStudio branch with this alpha, and see what tests pass/fail as an indicator of if we are missing anything critical?

packages/framer-motion/src/render/dom/DOMKeyframesResolver.ts
```diff
@@ -0,0 +1,183 @@
...
+
+        unresolvedKeyframes[unresolvedKeyframes.length - 1] = positionalValues[
+            name
+        ](
+            element.measureViewportBox(),
+            window.getComputedStyle(element.current)
+        ) as any
+
+        if (name === "height" && this.suspendedScrollY !== undefined) {
+            window.scrollTo(0, this.suspendedScrollY)
```
adamseckel, line 169: Why do we need to scroll the window to 0?

adamseckel (approved):

packages/framer-motion/src/render/dom/DOMKeyframesResolver.ts
mattgperry: We're scrolling it to the value it was before measurement - the measurement can fuck with page scroll

Issue #1160, opened 2021-05-17
[FEATURE] Flag to instantly finish animations


evanpurkhiser:
**Is your feature request related to a problem? Please describe.**

Hi all. We over at @getsentry use framer-motion (it's amazing!). We also do visual regression testing on our PRs, which means we generate screenshots from selenium based acceptance tests, then visually compare the screenshots against the screenshots generated from master.

For the screenshot comparisons, we have to make sure all animations are disabled. For framer motion We've added a [`testableTransition`](https://github.com/getsentry/sentry/blob/b331fd6128e853a222dc99c970cb4e5275db1494/static/app/utils/testableTransition.tsx#L19-L28) utility, which we then wrap all of our `transition` values with. The idea is that it just sets the transition `type` to `false` essentially disabling the animation.

This has generally worked, but we've noticed some places where the screenshots produce minorly  different results between runs. You can see one here https://storage.googleapis.com/sentry-visual-snapshots/getsentry/sentry/6ad776199421d49c927f23c0a62877b96ad348e5/index.html. Maybe it is related to inconsistencies in how the browser renders elements on a gpu accelerated layer?

I was wondering if there was any better way to globally "disable" animations in framer motion.

**Describe the solution you'd like**
`MotionConfig` could have an option to force framer motion to never "animate" anything, and always immediately transition states. This would be nicer than having to remember to always use `testableTransition`.

Open to any ideas. Thanks for this incredible library!

lewisl9029:
I have a _really_ crude temporary workaround for this involving using a combination of `<MotionConfig reducedMotion="always">` to disable all transitions involving movement, and a `* { opacity: 1 !important; }` global style to disable opacity transitions, and so far that seems to have disabled everything I needed.

Though the obvious severe caveat here is that this means opacity styles won't show up in tests, without tacking on another layer of `!important` (which would break opacity transitions in non-test environments).

I'd be happy to submit a PR to implement this properly if the maintainers (cc @mattgperry?) are open to accepting it! I'm thinking the API and implementation could be analogous to `reducedMotion` in `MotionConfig`, except just a `disable` boolean prop that disables all transitions, not just ones that cause movement.

mattgperry:
We would quite like something like this in Framer too so I might be able to take a look in the next couple

jzecca:
Very interested in this.

I need to generate screenshots of my app using Puppeteer, and being able to skip/disable all animations globally would be awesome.

gajus:
[...]

PR #2461, opened 2024-01-02
Skip animations globally

mattgperry:
This allows setting a config option that disables animations globally.

```javascript
import { MotionGlobalConfig } from "framer-motion"

MotionGlobalConfig.skipAnimations = true
```

Fixes https://github.com/framer/motion/issues/1160


Swannerino:
This doesn't disable animations, but rather causes them to complete instantly, which causes issues for components with cyclical animations. Is it possible to have an option that actually disables animations instead?

PR #1569, opened 2022-06-03
Fix AnimatePresence won't unmount fastly changing content

JaeSeoKim:
## Description

`exiting` object is not enclosed for `useRef()` as shown below

```tsx
-   const exiting = new Set<ComponentKey>()
+   const exiting = useRef(new Set<ComponentKey>()).current
```

Because the object `exiting` is a new object each time it is re-rendered, the `onExit` event validates the `Defer re-rendering untill all exiting child have indented left` incorrectly.

```tsx
const onExit = () => {
    allChildren.delete(key)
    exiting.delete(key)

    // Remove this child from the present children
    const removeIndex = presentChildren.current.findIndex(
        (presentChild) => presentChild.key === key
    )
    presentChildren.current.splice(removeIndex, 1)

    // Defer re-rendering until all exiting children have indeed left
    if (!exiting.size) {
        presentChildren.current = filteredChildren

        if (isMounted.current === false) return

        forceRender()
        onExitComplete && onExitComplete()
    }
}
```

---

###### 2022-06-11 update

A bug occurred because the `onExit` function is newly defined each time when a fastly outgoing elements unmount instantly.

To solve this problem, solved it by memoizing `exitingChild`.

```tsx
// If the component was exiting, reuse the previous component to preserve state
let extingChild = exitingChildren.get(key)
if (extingChild) return extingChild

const onExit = () => {
    ...
}

extingChild = (
    <PresenceChild
        key={key}
        isPresent={false}
        onExitComplete={onExit}
        custom={custom}
        presenceAffectsLayout={presenceAffectsLayout}
    >
        {child}
    </PresenceChild>
)
exitingChildren.set(key, extingChild)
return extingChild
```

But when I solved this problem, I saw a different problem.

The rendering array will have a strange change position.

Generally, if a change is made from `["A", "D", "E", "F", "B", "C"]` to `["1", "A", "2", "B", "3", "C"]` during animation, 
it is expected to be `["1" ,"A" ,"D" ,"E" ,"F" ,"2" ,"B" ,"3" ,"C"]`

It does not work as intended and is output in the form `["1", "D", "E", "F", "A", "2", "B", "3", "C"]`.

```tsx
// unnatural position
const insertionIndex = presentKeys.indexOf(key)

const onExit = () => {
  ...
}

// If the current array is smaller than "insertionIndex", it will not be inserted in the intended location
childrenToRender.splice(
    insertionIndex,
    0,
    <PresenceChild
      ...
    </PresenceChild>
)
```

[...]

PR #2477, opened 2024-01-05
Ensure AnimatePresence executes exiting animations in sequence

### LINKED ISSUE #2462: [BUG] `AnimatePresence` does not execute exiting animations in sequence during fast renders
✅

**2. Describe the bug**
Items with an exit animation within the `<AnimatePresence />` component do not reliably exit in sequence during fast renders.

https://codesandbox.io/p/devbox/dreamy-liskov-q35pmx

**4. Steps to reproduce**
1. Click on the 'Add' button a few times in quick succession.
2. Notice that the list items that appear below the button exit in a non-sequential order.

**5. Expected behavior**
The list items should exit in sequential order even during fast renders.

**6. Video or screenshots**

https://github.com/framer/motion/assets/62209188/d73c8cdc-8cdd-4037-adda-d9cd901c613c

Notice that 'hey 6' jumped above 'hey 5' at the 0:04 second mark, while 'hey 7' disappeared before it's supposed to.

regexyl:
This PR fixes #2462. 

It also fixes 3 tests (i.e. "Only renders one child at a time if mode === 'wait'", "Fast animations with wait render the child content correctly", "Fast animations with wait render the child content correctly (strict mode disabled)") where not enough leeway is given for exiting elements to completely exit from the DOM.

# Before

https://github.com/framer/motion/assets/62209188/d73c8cdc-8cdd-4037-adda-d9cd901c613c

# Now

https://github.com/framer/motion/assets/62209188/da6e0868-f36e-4d95-a13f-4486389d8b7f

### DIFF
--- a/packages/framer-motion/src/components/AnimatePresence/index.tsx
+++ b/packages/framer-motion/src/components/AnimatePresence/index.tsx
@@ -96,8 +96,9 @@ export const AnimatePresence: React.FunctionComponent<
     const isMounted = useIsMounted()
 
     // Filter out any children that aren't ReactElements. We can only track ReactElements with a props.key
-    const filteredChildren = onlyElements(children)
-    let childrenToRender = filteredChildren
+    const filteredChildren = useRef(onlyElements(children))
+    filteredChildren.current = onlyElements(children)
+    let childrenToRender = filteredChildren.current
 
     const exitingChildren = useRef(
         new Map<ComponentKey, ReactElement<any> | undefined>()
@@ -105,7 +106,6 @@ export const AnimatePresence: React.FunctionComponent<
 
     // Keep a living record of the children we're actually rendering so we
     // can diff to figure out which are entering and exiting
-    const presentChildren = useRef(childrenToRender)
 
     // A lookup table to quickly reference components by key
     const allChildren = useRef(
@@ -118,9 +118,7 @@ export const AnimatePresence: React.FunctionComponent<
 
     useIsomorphicLayoutEffect(() => {
         isInitialRender.current = false
-
-        updateChildLookup(filteredChildren, allChildren)
-        presentChildren.current = childrenToRender
+        updateChildLookup(filteredChildren.current, allChildren)
     })
 
     useUnmountEffect(() => {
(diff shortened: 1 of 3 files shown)


packages/framer-motion/src/components/AnimatePresence/index.tsx
```diff
@@ -96,17 +96,14 @@ export const AnimatePresence: React.FunctionComponent<
...
+    filteredChildren.current = onlyElements(children)
+    let childrenToRender = filteredChildren.current
 
     const exitingChildren = useRef(
         new Map<ComponentKey, ReactElement<any> | undefined>()
     ).current
 
-    // Keep a living record of the children we're actually rendering so we
-    // can diff to figure out which are entering and exiting
-    const presentChildren = useRef(childrenToRender)
```
regexyl, line 108: Removed `presentChildren` for clarity - `filteredChildren`, `exitingChildren` and `allChildren` are enough to maintain the status quo

regexyl:
@mattgperry I'm not sure why the test under framer-motion-3d is failing - do you have a clue?

JaeSeoKim:
In the past, PRs related to this issue have been closed due to personal circumstances, so it's good to know!

- #1569

regexyl:
Thanks @JaeSeoKim! Just had a quick look-through, and it seems like there are two main problems the PR was trying to solve - (1) bringing `useRef` to `exitingChildren` (also later separately solved through #2113), and (2) the algorithm that decides how exiting, current and entering elements mesh together.

This PR tries to solve a different problem where a quick waterfall of existing elements do not exit in sequence (i.e. elements exit in quick succession, but in different times), instead of how the exiting/current/entering elements should be rendered - though it seems like a pretty interesting problem to solve 🦄

mattgperry:
Thanks for the excellent PR and apologises it's taken a while to get to. I'll merge and publish today.

mattgperry:
Ah apologises. I had to revert this PR because of these lines:

```javascript
    const filteredChildren = useRef(onlyElements(children))
    filteredChildren.current = onlyElements(children)
```

The intention here is to update `filteredChildren` every render. But `filteredChildren` is essentially a mutable here and React renders are theoretically pure. 

Therefore, it's only safe to assign a `ref` during render on the initial render:

```javascript
    const filteredChildren = useRef(null)
    if (!filteredChildren.current)
      filteredChildren.current = onlyElements(children)
```

(Which obviously isn't the intent of this line)

Otherwise what we have is a situation where React could render something with a `children` list that isn't committed, and an Effect or `onExit` being run on a `children` list that isn't the one we think it is.

mattgperry:
Another issue - or maybe it's been fixed in the meantime? - is I ported the tests in this PR to `main` https://github.com/framer/motion/commit/8ee362878b1e18f4ea30df6239f33ce8806361ba and they pass already so I don't know if they capture the behaviour in the linked issue

Issue #2088, opened 2023-04-14
[BUG] useSpring does not follow MotionValue animated with animate()


hatsumatsu:
Using `useSpring` with `useMotionValue`, I expect the spring to follow the MotionValue. However, when animating the MotionValue via `animate()` —f.e. for sequencing animations— the spring is acting weird.

https://codesandbox.io/s/pedantic-greider-2r39t3

In this example we create a MotionValue and a Spring that follows that value

````js
const x = useMotionValue(0);
const xSpring = useSpring(x, {...config});
````

To debug the animation we add two `<motion>` components, one using the MotionValue and the other the Spring in their style prop.

Setting the Motionvalue via `set()` works as expected: The motion component using the MotionValue updates immediately, while the component using the Spring is smoothly animated.

Animating the MotionValue via `animate()` as described in [the docs](https://www.framer.com/motion/animate-function/##animate-a-motionvalue ) animates the first component smoothly while the second one moves unreliably.
This means that currently you can't use `.set()` and `animate()`side-by-side.

Maybe I'm using the API all wrong, but I would expect the Spring to follow the MotionValue smoothly.

mattgperry:
No that's definitely weird, I can take a look

hatsumatsu:
@mattgperry do you have a hint for me how I could work around this bug? Would be so cool to be able to use both declarative `<motion>` elements with injected motion values and imperative `animate()` side by side.

PR #2492, opened 2024-01-17
Passive velocity calculations

mattgperry:
Fixes https://github.com/framer/motion/issues/2088

Replaces the current velocity calculation strategy with a passive approach.

Before, velocity was semi-naive and semi-actively calculated.

By semi-naive, I mean that values were pushed from `current` to `prev` with no concept of when they were produced. The most recent animation frame delta was used to calculate the velocity between the two values.

Take this:

```javascript
const x = motionValue(0)
x.set(1)
x.set(2)
```

The delta of this would likely be calculated as (2 - 1) with a time delta of 16ms.

But as this value was updated within synchronous code, conceptually there is no time delta. The velocity should be 0. (The alternative being using performance.now() for infinitesimally small time deltas - useless for any practical application)

By semi-active, I mean that although velocity itself was only calculated when `getVelocity` was called, consider this code:

```javascript
const x = motionValue(0)

requestAnimationFrame(() => x.set(1))
```

Querying velocity would return the correct velocity. But by calling `set`, we previously called a job (which scheduled another job) to run at the end of the following frame to replace `prev` with `current`, so that after a long enough time the delta would be (1 - 1) and the velocity would be correctly zero.

This active process actually adds a lot of overhead to animations, removing it increases "best-case" (linear easing, numerical interpolation) animation speed by 2x.

Instead we now maintain an `updatedAt` timestamp. `prevFrameValue` now stores the value of `current` at the end of the previous frame along with another timestamp. These timestamps are used to calculated velocity. A time delta larger than 30ms is considered too long and velocity is `0`.

Additionally we have replaced `performance.now()` in some instances with `time.now()`. `time` is a new eventloop-synchronous timer that ensures one time is used for a block of synchronous code. If a value is updated multiple times within the same block of synchronous code, only the latest value is used to calculate velocity.

### DIFF
--- a/packages/framer-motion/src/projection/styles/types.ts
+++ b/packages/framer-motion/src/projection/styles/types.ts
@@ -13,18 +13,3 @@ export interface ScaleCorrectorDefinition {
 export interface ScaleCorrectorMap {
     [key: string]: ScaleCorrectorDefinition
 }
-
-// export type ScaleCorrection = (
-//     latest: string | number,
-//     layoutState: LayoutState,
-//     projection: TargetProjection
-// ) => string | number
-
-// export interface ScaleCorrectionDefinition {
-//     process: ScaleCorrection
-//     applyTo?: string[]
-// }
-
-// export type ScaleCorrectionDefinitionMap = {
-//     [key: string]: ScaleCorrectionDefinition
-// }
--- a/packages/framer-motion/src/value/use-spring.ts
+++ b/packages/framer-motion/src/value/use-spring.ts
@@ -5,8 +5,10 @@ import { useMotionValue } from "./use-motion-value"
 import { MotionConfigContext } from "../context/MotionConfigContext"
 import { SpringOptions } from "../animation/types"
 import { useIsomorphicLayoutEffect } from "../utils/use-isomorphic-effect"
-import { AnimationPlaybackControls } from "../animation/types"
-import { animateValue } from "../animation/animators/js"
+import {
+    MainThreadAnimationControls,
+    animateValue,
+} from "../animation/animators/js"
 import { frameData } from "../frameloop"
 import { millisecondsToSeconds } from "../utils/time-conversion"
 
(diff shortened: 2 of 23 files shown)


packages/framer-motion/src/projection/styles/types.ts
```diff
@@ -13,18 +13,3 @@ export interface ScaleCorrectorDefinition {
 export interface ScaleCorrectorMap {
     [key: string]: ScaleCorrectorDefinition
 }
-
-// export type ScaleCorrection = (
```
mattgperry, line 17: Collateral

packages/framer-motion/src/value/use-spring.ts
```diff
@@ -53,6 +53,8 @@ export function useSpring(
 
             stopAnimation()
 
+            console.log("reading value as ", value.getVelocity())
```
adamseckel: This needs to be removed before merging, right?

packages/framer-motion/src/value/use-velocity.ts
```diff
@@ -15,8 +16,16 @@ import { useMotionValue } from "./use-motion-value"
 export function useVelocity(value: MotionValue<number>): MotionValue<number> {
     const velocity = useMotionValue(value.getVelocity())
 
-    useMotionValueEvent(value, "velocityChange", (newVelocity) => {
-        velocity.set(newVelocity)
+    const updateVelocity = () => {
+        const latest = value.getVelocity()
+        velocity.set(latest)
+
+        if (latest) frame.update(updateVelocity)
```
adamseckel, line 23: This confused me for a second, but I think i understand it now. Am I right that this is intentionally creating a loop so that if the value stops changing, the velocity continues to be updated until its reported as 0?

adamseckel (approved):
Nice improvement!

packages/framer-motion/src/value/use-velocity.ts
mattgperry: Yeah that's right, let me add a comment

PR #2497, opened 2024-01-23
Update CHANGELOG.md to correct date

jacobjuul:
Version 11.0.0 was released in the future


mattgperry:
Thanks for the PR though since fixed

PR #2511, opened 2024-02-07
Optimisations

mattgperry:
- `clamp` - 2x faster
- `mixComplex` - 2x faster
- Prefer `mix` in `animate()` over `interpolate` (faster)


packages/framer-motion/src/animation/animators/js/index.ts
```diff
@@ -92,9 +95,12 @@ export function animateValue<V = number>({
                 `Only two keyframes currently supported with spring and inertia animations. Trying to animate ${keyframes}`
             )
         }
-        mapNumbersToKeyframes = interpolate([0, 100], keyframes, {
-            clamp: false,
-        })
+
+        mapNumbersToKeyframes = pipe(
```
mattgperry, line 99: Changed for better perf

packages/framer-motion/src/value/types/complex/index.ts
```diff
@@ -18,107 +13,95 @@ function test(v: any) {
...
-    const matches = info.tokenised.match(regex)
-
-    if (!matches) return
-
-    info["num" + countKey] = matches.length
-    info.tokenised = info.tokenised.replace(regex, token)
-    info.values.push(...(matches.map(parse) as any))
-}
+const complexRegex =
+    /(var\s*\(\s*--[\w-]+(\s*,\s*(?:(?:[^)(]|\((?:[^)(]+|\([^)(]*\))*\))*)+)?\s*\))|(#[0-9a-f]{3,8}|(rgb|hsl)a?\((-?[\d\.]+%?[,\s]+){2}(-?[\d\.]+%?)\s*[\,\/]?\s*[\d\.]*%?\))|((-)?([\d]*\.?[\d])+)/gi
```
adamseckel, line 38: 🤯

mattgperry: Ha its not as clever as it looks, just three capture groups of preexisting color/number/var() patterns

adamseckel (approved):

PR #2534, opened 2024-02-29
Fix CSS variable regex performance

mattgperry:
Surprising values can cause a catastrophic degredation in performance of this regex

```
var(--token-31a8b72b-4f05-4fb3-b778-63a7fb0d9454, hsl(224, 78%, 54%)) /* {"name":"Midnight Blue"} */
// ~5ms

var(--token-c534b380-e14e-4ddc-9802-ad88d1f94f8e, rgba(255, 255, 255, 0.12)) /* {"name":"Border Night"} */
// ~500ms
```

This PR trims strings of comments and brings execution time reliably down to ~0ms.


tom-james-watson (approved):
LGTM

Though this would break with keys where the comment is at the start - I presume we're ok with that?

mattgperry:
I think we only put comments at the end in Framer, and I only added support for this at all for Framer. I think anyone else can just not try to animate comments 😅

Issue #1727, opened 2022-10-05
[BUG] Content Security Policy blocks inline styles


sylvaindubus:
Hello,

I'm currently working on a project where we have a strict CSP which prevents inline execution of styles. This is problematic here as this library relies on style attributes (`style="opacity: 1; transform: none;"` for instance). Unfortunately, I couldn't find a workaround, like with a `nonce` attribute which we already use for other libraries. 

Thanks for any help.

Regards,

mattgperry:
Unfortunately that functionality is core to all JavaScript animation libraries. I think to fix you'd have to stick to CSS animations and transitions.

ianwensink:
Hi @mattgperry. Is this really the only acceptable answer? Is there no way to have a workaround for this?

Has no one ever used a strict CSP ánd Framer motion (or any other animation library) at the same time?

bricejar:
@mattgperry @ianwensink @sylvaindubus  Hi ! Tell me if I need to open another issue. There might be a way to satisfy a stricter Content Security Policy.  

Globally allowind inline styles on a web app is a security vulnerability. If an attacker finds a way to display html with a malicious inline style in your website they could do a lot of harm. 
With CSP, you can allow dynamic style tags with a randomly generated nonce. So instead of showing the styles inline, you could create a class and generate a <style> to style this class. I assume this could have a performance impact but that would be the best mitigation strategy. 

Unfortunately right now there is no way to hook into the framer-motion machine to prevent inline styles from being added and transform them into a class. Just food for thought !

fdev:
Not all methods of setting inline styles are blocked by `style-src self`. Basically, using `element.style.cssProp = value` from javascript doesn't violate CSP.

See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/style-src for examples of what is and what isn't blocked.

jv4pres:
as @fdev points out, there is no issue once the page is loaded -- Framer Motion is already fully compatible with a strict CSP today *on the client side*, because setting styles via script is supported by all security policies. this issue is limited to server-side rendering; if the server generates an inline `style` attribute, the client will raise a security error.

[...]

PR #2541, opened 2024-03-04
add: content-security-policy nonce support

lochie:
this PR adds support for [content-security-policy style nonces](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/style-src#unsafe_inline_styles)

fixes: https://github.com/framer/motion/issues/1727 https://github.com/framer/motion/discussions/2406

---

**use case:**

when a content security policy (CSP) is set up to require a nonce, all `style` tags within the document should also include the matching generated nonce.

this primarily has issues when using `<AnimatePresence>` as `PopChild` appends a `<style>` tag to the dom.

---
**example:**

if a CSP was set up with the following policy with a generated nonce:
```
Content-Security-Policy: style-src 'nonce-2726c7f26c'
```

all `style` tags within the document should also include the the matching generated nonce:
```html
<style nonce="2726c7f26c">
  .el {
    background: red;
  }
</style>
```

---

**proposed usage:**

given that the `MotionConfig` component is the only context provider in the library, it would be the only place where the nonce could currently be set.

the `nonce` prop can be added to the `MotionConfig` component to allow the `motion` components to use the nonce.

```jsx
import { MotionConfig } from "framer-motion";

<MotionConfig nonce="2726c7f26c">
  <motion.div animate={{ x: 100 }} />
</MotionConfig>
```

the [MotionConfig documentation](https://www.framer.com/motion/motion-config/) should also be updated to include this new option.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ Framer Motion adheres to [Semantic Versioning](http://semver.org/).
 
 Undocumented APIs should be considered internal and may change without warning.
 
+## [11.0.9] 2024-03-05
+
+### Added
+
+-   Added support for Content Security Policy (CSP) nonces.
+
 ## [11.0.8] 2024-02-29
 
 ### Fixed
--- a/dev/examples/MotionConfig-nonce.tsx
+++ b/dev/examples/MotionConfig-nonce.tsx
@@ -1,0 +1,71 @@
+import * as React from "react"
+import { motion, MotionConfig, AnimatePresence } from "framer-motion"
+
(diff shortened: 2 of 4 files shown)


lochie:
hey @mattgperry, this is a relatively simple PR that solves a fairly niche edge-case.
curious to hear your thoughts on this and if there's anywhere else in the codebase that `<style>` elements are implemented  (I could only find it within `PopChild`).

I'm happy to update this PR to accomodate any changes if necessary.

PR #2587, opened 2024-03-22
TypeScript 5

mattgperry:


Coobaha (commented):
👏 Mostly did a quick check and pushed minor fixes while reviewing 👼 

There are some unsound places due to typecasting. Fixing them would require converting many parts to be more static and may not be worth it as it could require significant changes and new bugs. 

- I noticed that the dev command logs errors.

image

- bumped the CI base image to upgrade the Node.js version to v20 so it works with the new syntax.

- bumped jest/eslint deps related to TS as they were broken (at least in my editor)

- I think tests and editor are using different `config` as I had no errors in the  editor,r but tests were showing `tsc` errors

Coobaha (approved):

packages/framer-motion/src/animation/sequence/__tests__/index.test.ts
```diff
@@ -493,9 +494,9 @@ describe("createAnimationsFromSequence", () => {
         const { duration, ease, times } = animations.get(a)!.transition.x
 
         expect(duration).toEqual(2)
-        expect(ease![0]).toEqual("linear")
-        expect(ease![1]).toEqual("linear")
-        expect(typeof ease![2]).toEqual("function")
+        expect((ease as Easing[])[0]).toEqual("linear")
```
o-alexandrov, line 497: @mattgperry 
There is degradation in type-checking.
It’s safer to use non-null assertion than `as`.
https://www.reddit.com/r/typescript/comments/11rqtn3/is_as_any_better_than_to_assert_nonnullishness/

packages/framer-motion/src/animation/sequence/create.ts
```diff
@@ -330,8 +332,11 @@ export function getValueTransition(
     transition: DynamicAnimationOptions & At,
     key: string
 ): DynamicAnimationOptions {
-    return transition[key]
-        ? { ...transition, ...transition[key] }
+    return transition[key as keyof typeof transition]
```
o-alexandrov, line 335: @mattgperry how about depending on:
- return type
- and global per-repo type override for loops, so the key is auto-inferred from the array
  - you should also consider to switch to Object.keys for a slight performance boost 
  - https://dev.to/tmhao2005/typescript-override-typing-for-objectkeys-5dmh

It would ensure cases like this don’t need manual `as` override, removing possibilities of human errors.

If you still want to manually define `key`’s type, then you should do it [once before the for loop](https://stackoverflow.com/a/59656330).

packages/framer-motion-3d/src/render/use-render.ts
```diff
@@ -43,12 +43,14 @@ function useVisualProps(props: ThreeMotionProps & MeshProps) {
         const visualProps: ThreeMotionProps & MeshProps = {}
 
         for (const key in props) {
-            const prop = props[key]
+            const prop = props[key as keyof typeof props]
```
o-alexandrov, line 46: @mattgperry how about depending on:
- return type
- and global per-repo type override for loops, so the key is auto-inferred from the array
  - you should also consider to switch to Object.keys for a slight performance boost 
  - https://dev.to/tmhao2005/typescript-override-typing-for-objectkeys-5dmh

It would ensure cases like this don’t need manual `as` override, removing possibilities of human errors.

If you still want to manually define `key`’s type, then you should do it [once before the for loop](https://stackoverflow.com/a/59656330).

PR #2605, opened 2024-04-03
Fixing layout projection with `motion(Fragment)` nodes

mattgperry:
This PR skips adding `motion(Fragment)` nodes to the layout projection tree.

We added these nodes to allow animating variants via a renderless component rather than being bound to a `div` etc (@adamseckel has more context)

The problem with these nodes is layout animation scale correction is presumed to be performed by a child node, thus these nodes were swallowing the scale corrective calculations.

### DIFF
--- a/packages/framer-motion/src/render/dom/types.ts
+++ b/packages/framer-motion/src/render/dom/types.ts
@@ -21,6 +21,13 @@ export interface DOMVisualElementOptions {
     allowTransformNone?: boolean
 
     /**
+     * If `true`, this element will be included in the projection tree.
+     *
+     * @public
+     */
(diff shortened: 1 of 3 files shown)


packages/framer-motion/src/render/dom/types.ts
```diff
@@ -20,6 +20,13 @@ export interface DOMVisualElementOptions {
      */
     allowTransformNone?: boolean
 
+    /**
+     * If `true`, this element will be included in the projection tree.
```
shuangq, line 24: If `true` or undefined?

shuangq (approved):

PR #2619, opened 2024-04-12
feat: use modern jsx transform

kurtextrem:
This PR updates the jsx transform to the one introduced with React 17: https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html

Starting with React 19, there will be a warning when using the old one (see https://github.com/facebook/react/pull/28781), plus the new transform gives a nice speed-bump with React 19: https://github.com/facebook/react/pull/28781

I've also run `npx react-codemod update-react-imports .` to update / remove imports where not needed. I've split those into two commits so that the config changes and the autofixes can be looked at separately.


mattgperry:
Great, thanks for this!

mattgperry (approved):

PR #2629, opened 2024-04-18
Fixing `motion.div` types for TS5

mattgperry:


iamakulov:
Just for education, why did this break TypeScript 👀

kurtextrem (approved):

kurtextrem:
I assume because of https://www.totaltypescript.com/jsx-element-vs-react-reactnode#whenshouldntwe-usereactreactnode so something is not on TS5.1+?

PR #2635, opened 2024-04-19
Binding `stop()` to animation

mattgperry:
This PR binds `stop` to the animation to fix a pattern where `animation.stop` is passed back from `useEffect`.

```
useEffect(() => {
  const animation = animate()
  return animation.stop
})
```


nvh (approved):
[goodidea](https://app.graphite.dev//settings/meme-library?org=framer)

Issue #2514, opened 2024-02-11
[BUG] layoutid has incorrect positioning when height adjusts


Sebastian-Nielsen:
These two issues touching on the same bug has still not been properly addressed:
1. https://github.com/framer/motion/issues/1972
2. https://github.com/framer/motion/issues/2006

The solution to simply add 
![image](https://github.com/framer/motion/assets/34586104/9fac7d2c-81e8-423f-9355-e6b2f2b8d65f)

wont work for the case when your layoutid is to animate vertically. E.g. imagine a case where you stacked the navigation buttons vertically as opposed to horizontally.

kKaskak:
This indeed does not work, here is a quick sandbox for it: https://codesandbox.io/p/devbox/framer-motion-layoutid-bug-vertical-scroll-v92zzt

Issue #2006, opened 2023-03-13
[BUG] - broken layoutId animation with Next.js page change and scroll


ValentinH:
## Describe the bug

We are using `layoutId` to animate a current link indicator in our navbar (using fixed positioning) in a Next.js app. The animation works as expected during page navigation except when the page is scrolled. In that case, the initial position of the animation seems to be affected by the scroll offset. See this video:

https://user-images.githubusercontent.com/2678610/224677550-be7bfb11-1eba-45df-a190-77c7c3a99444.mov

## Steps to reproduce

1. Go to https://codesandbox.io/p/sandbox/adoring-albattani-qtdt3b?file=%2Fsrc%2FLayout.tsx
2. Scroll the page
3. Click on the "Page 1" link
4. The animation is incorrect

## Expected behavior

The animation should work regardless of the scroll.

## Environment details
- MacOS 13.2.1
- Chromium 110, Firefox 110  and Safari 16.3
- framer-motion: 10.2.4 

## Similar issues
This issue seems similar to https://github.com/framer/motion/issues/1580 but I created a new one because I'm not fully sure that's the same issue. Feel free to tell me to close this one and comment there. 🙏

CRIMSON-CORP:
I have another work around🙂, in case this issue still persists in your work.

The idea is to make the parents of the nav and the main page to have the exact height of the screen, and make the content page or its descendant to overflow, that way, the sticky nav is never scrolled, and hence retains its position on the page so it doesn't fly around when the page is changed, this might seem hacky and all but it is the cleanest solution I was able to think off

checkout this spinoff of your sandbox
https://codesandbox.io/p/sandbox/adoring-albattani-qtdt3b?file=%2Fsrc%2FLayout.tsx

CRIMSON-CORP:
Thank you @ValentinH, I've updated my comment

minardimedia:
Anybody from the maintainers putting attention to this ?. This looks similar to something that can be fixed with parent <montion.div layout layoutRoot > the case is that it does not solve the issue. The <motion.div layoutId={"underline"}> elements keeps using the whole window to make is calculations for the transform so when you scroll and change from page the calculation is wrong. The whole point of layoutRoot in my opinion was to avoid this in the the first place but it does not work.

n1ckoates:
Hi! Does anyone here know if there's a workaround for this when using the app router?

simonlagneaux:
[...]

Issue #1972, opened 2023-02-23
[BUG] layoutId has incorrect positioning when height adjusts.


Looskie:
When underlying children have height adjustments, layoutId seems to not know where to go and results in a weird and inconsistent animation.

**3. IMPORTANT: Provide a CodeSandbox reproduction of the bug**
https://codesandbox.io/p/sandbox/framer-motion-layoutid-issue-uu7kb9?selection=%5B%7B%22endColumn%22%3A13%2C%22endLineNumber%22%3A88%2C%22startColumn%22%3A13%2C%22startLineNumber%22%3A88%7D%5D&file=%2Fsrc%2FApp.tsx

(I even added a button with a "work-around", but the work-around is not the preferred way of fixing it (especially in my use case where I'm rendering pretty hefty components)

**4. Steps to reproduce**

Check sandbox and switch through tabs

**5. Expected behavior**

The underline should stay in its place and not animate off to the ends

**6. Video or screenshots**

what its NOT supposed to do
https://user-images.githubusercontent.com/50378828/220937673-29501ab9-6c78-41d0-9832-3c7218c63206.mov

what its SUPPOSED to do

https://user-images.githubusercontent.com/50378828/220937849-fe5a55ff-54d5-43ce-8672-96dd066d8715.mov

**7. Environment details**
N/a

hexiro:
I'm experiencing this same issue when the `layoutId` components are `position: fixed` and the page scrolls. This could be caused by the same mechanic under the hood, but regardless, it produces an undesired effect with the Y values animating.

hexiro:
workaround: [codesandbox](https://codesandbox.io/p/sandbox/framer-motion-layoutid-issue-forked-q0upoj)

![image](https://user-images.githubusercontent.com/42787085/227646943-2e467f16-0454-4700-b530-fe3e225e71c5.png)

the patch is on line 48 and works by setting `originY` to `0px`, which effectively disables the Y value from shifting during the animation.

Sebastian-Nielsen:
> workaround: [codesandbox](https://codesandbox.io/p/sandbox/framer-motion-layoutid-issue-forked-q0upoj)
> 
> ![image](https://user-images.githubusercontent.com/42787085/227646943-2e467f16-0454-4700-b530-fe3e225e71c5.png)
> 
> the patch is on line 48 and works by setting `originY` to `0px`, which effectively disables the Y value from shifting during the animation.

[...]

PR #2647, opened 2024-05-02
Fix projection of fixed elements when scrolling

koop:
Fixes the failing `position: fixed;` projection tests added in #1768.
- I verified that all other projection tests still passed (I verified this manually because I couldn't get the cypress configuration to run on my local machine).

Fixes issues #2514, #2006, and #1972.

| Before  | After |
| ------------- | ------------- |
| [Sandbox to reproduce the issue](https://codesandbox.io/p/devbox/framer-motion-fixed-projection-bug-yzsrql?file=%2Fsrc%2FApp.tsx)  | [Sandbox to verify fix (with patch applied)](https://codesandbox.io/p/devbox/framer-motion-fixed-projection-fix-with-patch-7gx7pl?file=%2Fsrc%2FApp.tsx%3A9%2C26)  |
|  |  |

### Details

**To handle projecting fixed elements within scrollable pages:** as I understand it, the existing logic accounted for scroll root ancestors within `removeElementScroll` by iterating over the `path`, but this didn't account for the case where the current element was a fixed element, because the `path` does not contain the current node. Instead, we now avoid adding the scroll position in `measurePageBox` when the projection is or is within a fixed element.

**To handle projected fixed elements within scrollable elements:** this was a similar case where we weren't accounting for the current element's scroll information. If the current element is fixed, we can return early from `removeElementScroll`.

**To handle projecting when an element changes between fixed and static positions:** we track whether the position was fixed during the snapshot phase (instead of relying on the most recently measured value) and use that to determine whether the projection considers an element fixed or not. Two considerations here:
- I'm not sure about the nuances of resuming projections and whether or not there would be a more correct way to track this information.
- This does not solve for the case where an element transitions between `position: sticky;` and `position: fixed;`. Fixing this may involve tracking more than just whether or not the element is fixed when the snapshot is taken; I haven't spent enough time with that failing test case.


koop:
Sounds good!

> We don't need to worry so much about position: sticky - stickied elements themselves can flick between "fixed" and "static" modes and there's no good way of tracking this.

That's roughly what I expected (and why I didn't include it here), monitoring sticky position/state is a pain.

lilingxi01:
Thank you for making this PR! This could resolve a lot of transition glitches related to `position: fixed`.

Thanaen:
Is it possible to finance this PR (via [Polar.sh](https://polar.sh/) for example) to support the work needed to merge it? 😄

mattgperry:
Thanks for this excellent PR and apologises for the delay. I'm merging into an intermediate branch to try it out and if all tests are running well I'll publish in next couple days.

PR #2648, opened 2024-05-05
Export `createDomVisualElement`

o-alexandrov:
Currently, it's impossible to create a custom `FeatureBundle` (smaller version of `domAnimation`) because `createDomVisualElement` is not exported.

https://github.com/framer/motion/blob/df6ccaf61a59c0b835234830685222c632139444/packages/framer-motion/src/render/dom/features-animation.ts#L10

By exporting `createDomVisualElement`, users would then be able to create a smaller `FeatureBundle`, for example:

```ts
export default {
  renderer: createDomMotionComponent,
  ...animations, // `animations` variable is already exposed
} satisfies FeatureBundle
```

### DIFF
--- a/packages/framer-motion/src/index.ts
+++ b/packages/framer-motion/src/index.ts
@@ -13,7 +13,11 @@ export * from "./dom-entry"
 export * from "./three-entry"
 
 /**
- * Features
+ * To create a custom FeatureBundle (ex. part of domAnimation), we expose renderer (createDomVisualElement)
+ */
+export { createDomVisualElement } from "./render/dom/create-visual-element"
+/**
+ * Features (groups of features exposed for convenience)
  */
 export { domAnimation } from "./render/dom/features-animation"
 export { domMax } from "./render/dom/features-max"


o-alexandrov:
@mattgperry could you please review the PR?

o-alexandrov:
This PR is literally a one line code change which only exports a single variable that unblocks all use cases with slimmer LazyMotion.

@mattgperry could you please merge?

mattgperry:
Hi @o-alexandrov 

Although it's "literally a one line code change" it is also changing public API. The reason these bundles are exposed as (documentationally-speaking) opaque variables as these can then change under-the-hood at any time without breaking anything. Or in other words, I don't want people making custom bundles without a good API around that.

What I can do instead though is export a minimal bundle that just has animations included - but it's not a big saving over `domAnimations`

mattgperry:
`domMin` will be available in `11.3.17`

o-alexandrov:
You are welcome not to include `domMin`, if not needed.
I no longer use `framer-motion`, as CSS animation with CSS variables handle all cases I previously relied on using `framer-motion`

Issue #2656, opened 2024-05-10
[BUG] Animating display from 'block' to 'none' not working since 11.0.11


dd-jonas:
**1. Read the [FAQs](#faqs) 👇**

**2. Describe the bug**

When defining variants with `display: block` and `display: none`, the display does not return to `none` upon triggering the animation.

**3. IMPORTANT: Provide a CodeSandbox reproduction of the bug**

[Reproduction](https://stackblitz.com/edit/vitejs-vite-mc7x2z?file=src%2FApp.tsx)

**4. Steps to reproduce**

1. Go to the sandbox above
2. Click on the show/hide button to show the overlay
3. Click on the show/hide button to hide the overlay
4. Notice that the opacity has animated, but display is still block (notice that the backdrop is still there, because you can't select nay text on the screen)

**5. Expected behavior**

The overlay should return to `display: none` as was the case before 11.0.11.

**6. Video or screenshots**

/

**7. Environment details**

/

mattgperry:
Yeah this did get inadvertently broken though the previous behaviour wasn't great either (instantly animating to "none" when you probably want that to happen at the end of the animation)

I'll take a look over the next couple days and see if we can do something where animating `display` and `visibility` can be handled more gracefully.

alex-cortenix:
Any progress on that??

PR #2660, opened 2024-05-13
Interpolation between visible and hidden

mattgperry:
Implements https://github.com/framer/motion/issues/2656

This PR adds interpolation of visible <-> hidden.

It adds behaviour that when animating to/from "hidden" or "none" and a value-less string that the non-hidden value is returned for all progress values except `<= 0` (if hidden is the first keyframe) and `>= 1` (if hidden is the last keyframe).

This currently works with `display` and `visibility` styles, and also any detected interpolation in `mix` and `useTransform` between a value-less string and `"none"`/"hidden"`

### DIFF
--- a/dev/examples/Animation-display-visibility.tsx
+++ b/dev/examples/Animation-display-visibility.tsx
@@ -1,0 +1,32 @@
+import { useState } from "react"
+import { MotionConfig, motion } from "framer-motion"
+
+/**
+ * An example of the tween transition type
+ */
+
+const style = {
+    width: 100,
+    height: 100,
+    background: "white",
+}
+
+export const App = () => {
+    const [state, setState] = useState(true)
+
+    return (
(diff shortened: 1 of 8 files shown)


dev/examples/Animation-display-visibility.tsx
```diff
@@ -0,0 +1,32 @@
...
+    return (
+        <MotionConfig transition={{ duration: 1 }}>
+            <motion.div
+                initial={{ display: "block" }}
+                animate={{
+                    display: state ? "block" : "none",
+                    visibility: state ? "visible" : "hidden",
+                    opacity: state ? 1 : 0.2,
+                }}
+                onUpdate={console.log}
```
kaloyanvi, line 26:
```suggestion
                onUpdate={() => console.log("on update")}
```

kaloyanvi (approved):
LGTM 🚀

PR #2661, opened 2024-05-14
Fix jump if optimised appear animation is interrupted

mattgperry:
If a large amount of work is performed between a deferred animation starting and keyframe resolution, there can be a perceptible jump in interrupted optimised animations. This PR fixes that by deferring cancellation until the following frame.


dev/optimized-appear/defer-handoff-block.html
```diff
@@ -0,0 +1,155 @@
...
+                                    showError(
+                                        document.getElementById("box"),
+                                        `Stutter detected`
+                                    )
+                                }
+                            })
+                        })
+                        /**
+                         * By blocking the main thread here, we ensure that
+                         * the keyframes
```
shuangq, line 81: This comment is incomplete...?

packages/framer-motion/src/animation/optimized-appear/handoff.ts
```diff
@@ -34,9 +34,28 @@ export function handoffOptimizedAppearAnimation(
     const cancelAnimation = () => {
         appearAnimationStore.delete(storeId)
 
-        try {
-            animation.cancel()
-        } catch (error) {}
+        if (frame) {
+            /**
+             * If we've been provided the frameloop as an argument, use it to defer
+             * cancellation until keyframes of the subsequent animation have been resolved.
```
shuangq, line 40: Can we also elaborate a bit why we need to defer the cancellation?

shuangq (approved):
Nice ❤️

PR #2672, opened 2024-05-16
Fixing optimised appear jump with `useLayoutEffect` and layout animations

mattgperry:
This PR fixes a jump in optimised appear animations when `useLayoutEffect` triggers a state update leading to a layout animation.

By moving `CancelHandoffAnimations` to before the snapshot is taken, not only do we fix the snapshots but also the cancel is skipped currently in Framer as no layout animations are detected on this mount re-render, leaving the smoother animation in tact.

### DIFF
--- a/packages/framer-motion/src/motion/utils/use-visual-element.ts
+++ b/packages/framer-motion/src/motion/utils/use-visual-element.ts
@@ -63,7 +63,7 @@ export function useVisualElement<Instance, RenderState>(
     useIsomorphicLayoutEffect(() => {
         if (!visualElement) return
 
-        microtask.postRender(visualElement.render)
+        microtask.render(visualElement.render)
 
         /**
          * Ideally this function would always run in a useEffect.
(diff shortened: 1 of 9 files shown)


packages/framer-motion/src/projection/node/create-projection-node.ts
```diff
@@ -577,6 +577,11 @@ export function createProjectionNode<I>({
             if (this.isUpdateBlocked()) return
 
             this.isUpdating = true
+
+            if (window.HandoffCancelAllAnimations) {
+                window.HandoffCancelAllAnimations()
+            }
```
shuangq, line 583: Can we add a comment why this happens before the snapshot?

packages/framer-motion/src/motion/utils/use-visual-element.ts
```diff
@@ -63,7 +63,7 @@ export function useVisualElement<Instance, RenderState>(
     useIsomorphicLayoutEffect(() => {
         if (!visualElement) return
 
-        microtask.postRender(visualElement.render)
+        microtask.render(visualElement.render)
```
shuangq, line 66: What is this change? (and the microtask.preRender change below)

shuangq (approved):

packages/framer-motion/src/motion/utils/use-visual-element.ts
mattgperry: True, not sure these are needed any more. `render` is the "right" place for this one to fire but `preRender` I think was an artefact of when I was trying some things out.

PR #2685, opened 2024-05-28
Reduce optimised appear animation de-opts via layout animations

mattgperry:
Currently, when any layout animation fires during optimised appear animations, all optimised appear animations are cancelled.

Due to difficulties matching timings, especially on heavier sites this can result in a visual stutter.

This PR adds logic to determine whether the node performing a layout animation is, or is the ancestor of, a node performing an optimised appear animation. Only then will optimised appear animations be cancelled.

**Note:** All optimised appear animations are still cancelled under this model, as cancelling a WAAPI animation is a DOM write. Snapshotting elements is a DOM read. So this cancel write might be sandwiched amongst other DOM reads. By cancelling all optimised animations at the same time, we ensure there is only one read/write/read thrash, vs unlimited.


packages/framer-motion/src/projection/node/create-projection-node.ts
```diff
@@ -100,6 +101,25 @@ function resetDistortingTransform(
...
+
+    const { visualElement } = projectionNode.options
+
+    if (!visualElement) {
+        return false
+    } else if (getOptimisedAppearId(visualElement)) {
+        return true
+    } else if (
+        projectionNode.parent &&
+        !projectionNode.hasCheckedOptimisedAppear
```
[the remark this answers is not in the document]
mattgperry: ty - fixed

dev/optimized-appear/defer-handoff-layout-ancestor.html
```diff
@@ -0,0 +1,167 @@
+<!--
+  This test checks that the layout animation firing on an child element
+  cancel the optimized appear animation on an ancestor element.
```
shuangq, line 3: Nice tests ❤️

packages/framer-motion/src/projection/node/create-projection-node.ts
```diff
@@ -100,6 +101,26 @@ function resetDistortingTransform(
...
+    } else if (getOptimisedAppearId(visualElement)) {
+        return true
+    } else if (
+        projectionNode.parent &&
+        !projectionNode.parent.hasCheckedOptimisedAppear
+    ) {
+        return isOptimisedAppearTree(projectionNode.parent)
+    } else {
+        return false
+    }
```
shuangq, line 121: nit: since it's all return statements, maybe we can get rid of the else blocks 
```suggestion
    if (!visualElement) return false  
    if (getOptimisedAppearId(visualElement)) return true
    if (
        projectionNode.parent &&
        !projectionNode.parent.hasCheckedOptimisedAppear
    ) return isOptimisedAppearTree(projectionNode.parent)

    return false
```

shuangq (approved):

PR #2691, opened 2024-05-31
Updating interpolation for invalid colors

mattgperry:
When `mixColor` receives an invalid color, we currently throw an error in development mode. If uncaught and fixed, this can lead to hard throws in production.

This PR replaces the throw with a warning, and changes the interpolation from color to immediate.


adamseckel (approved):
Nice!

Issue #2554, opened 2024-03-13
[BUG] AnimatePresence gets stuck when state changes quickly 


tommhuth:
**2. Describe the bug**
Framer Motion `v11.0.12`: When state changes very quickly `AnimatePresence` gets stuck and out of sync and no longer removes the element but instead always animates it in. Reverting  to version `v10.12.2` (based of other bug report concerning a similar/same issue) does not exhibit the same problem

**3. IMPORTANT: Provide a CodeSandbox reproduction of the bug**
[https://codesandbox.io/p/sandbox/serene-brattain-qwd278](https://codesandbox.io/p/sandbox/serene-brattain-qwd278)

**4. Steps to reproduce** 

1. Conditionally render an element inside AnimatePresence with an exit animation
2. Trigger condition change quickly
3. Animating component gets stuck as rendered, and any changes to the condition will just trigger an reset from initial => animate. The component is no stuck as visible and will never leave the DOM. 

**5. Expected behavior**

The component should not get stuck as mounted

**6. Video or screenshots**

If applicable, add a video or screenshots to help explain the bug.

**7. Environment details**

Chrome 122, Mac Sonoma

justrealmilk:
Your code sandbox is broken

Roon311:
I am encountering the same issue

mattgperry:
Thanks for the sandbox - are you able to set it to public so I can take a look?

Roon311:
https://codesandbox.io/p/github/Roon311/trying/master?workspaceId=e4d7df23-6eed-436c-861f-b723f62062f5

Roon311:
@mattgperry When you use the right arrown to navigate through the animating images, the error shows up

notmedia:
@Roon311 you link is also unavailable 
```
Unable to access this workspace
It is likely that you are not a member of the team that this project belongs to. To be able to access the web editor of this repository and edit the code, you need to be invited to the team.
```

justrealmilk:
Okay now that I know why it breaks I can actually trigger the problem https://codesandbox.io/p/sandbox/modern-wood-h35hc6

Roon311:
@notmedia is it still inaccessible ?

OlegWock:
Here is another reproduction. Double click the button to see it

https://stackblitz.com/edit/hold-to-confirm-framer-motion-fj53pr?file=package.json

I can confirm that framer motion 11.0.8 works correctly with same code

tommhuth:
[...]

PR #2695, opened 2024-06-07
Fix AnimatePresence

wonkyum-kim:
#2554 

This PR fixes an AnimatePresence bug.

A problem seems to occur if we add a component with the same key while the existing exit animation is not finished.

So I tried a few things to fix it.

I think it worked out well and passed all the tests.

The example below is taken from #2554. (https://codesandbox.io/p/devbox/great-brattain-hc9kgz)

before

https://github.com/framer/motion/assets/67191801/c88ff93c-d61b-45ff-92a4-27bb4262dae1

after

https://github.com/framer/motion/assets/67191801/08e847a6-9463-4056-a3f1-a63353757297


mattgperry:
Thanks for the PR! Though closing this out as I think this is actually related to interrupting a WAAPI animation so I'm gonna make an E2E test covering this and then publish a fix.

PR #2699, opened 2024-06-13
Upgrade React Three Fiber 9

mattgperry:


gavan-x:
@mattgperry  does this mean we can just use motion instead of framer motion 3d?

PR #2702, opened 2024-06-17
Require and return a numeric MotionValue in useSpring types

### LINKED ISSUE #2495: [FEATURE] Stricter `useSpring` signature

The return type of `useSpring` is `MotionValue<any>` which forces users to manually cast it to a safer type, or do unsafe things with the values. Also, the signature accepts arbitrary `MotionValue` sources, even though the [docs](https://www.framer.com/motion/use-spring/##create) & [implementation](https://github.com/framer/motion/blob/42bd9fbb7502ab24217fb238760356467d7d9d0d/packages/framer-motion/src/value/use-spring.ts#L82) only allow numeric motion values.

Simple example ([TypeScript Playground)](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgVwM4FMDKYrAHYDmANChgLIQzAR4BqAhgDbLpwC+cAZlBCHAOTd6IdFAC0IStTz84AejlwYATzDpUALjgBGAAwA6bQA59ugFBmAxjVTxUOfATgBeUlgeEAFLoCUZuAGBCnAAegD8Fva4hPoYMJ4ARJwQEAk+8ooAKgAWwKhwqNkQyIwAJnA4EKXIlqz0eHCiPFAWVjbweDQAcsgiuJYUVDQMzKyuaOiD0iMsiQDEnItpbXi2BR4EAEwubtjRBJ6deD19wANSw0ws6cE5eQVFJeWV1bVw9Y1QzUA)

```ts
const spring = useSpring(0)
      // ^? const spring: MotionValue<any>

spring.set("foo") // This should produce an error

const nonNumericMotionValue = useMotionValue("#fff")
useSpring(nonNumericMotionValue) // This should produce an error too
```

1. Return `MotionValue<number>` instead of `MotionValue<any>`
2. Only accept `MotionValue<number>` instead of arbitrary `MotionValue`

This can be accomplished by changing the source type [here](https://github.com/framer/motion/blob/42bd9fbb7502ab24217fb238760356467d7d9d0d/packages/framer-motion/src/value/use-spring.ts#L34) to `MotionValue<number> | number` and the return type will be fixed accordingly.

In this [TypeScript Playground](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgVwM4FMCyEbAgOwDUBDAG2XQBo4YBPMdObXAk8qm+xgZTCmHwBzAPJgW+VHAC+cAGZQIIOAHJ5xEOigBaEDjz5lcAPRHODVAC44ARgAMAOmsAOe7YBQbgCboAxqWJQjLLI+D7iKBi8-EIAFG5wCXCoEMhQPuhWzPpsFAA8+MggAEaaAHxwAD5wBcWalPGJPgSywIIA-FZRAiJi+qhuAJSZeqxkeTUlUKUeTRLwqHzdcAC8EehdsbYDDYnGpgB6bR4L0YL2GDAxAESyEBBXA3twAKJQClBuJ93n6JfWjyY4AA5CBwTTvDxuWaoeD4AhAwqaYA+LKjdgrNaoohjdDXADEskJDyhBBhSUWQgATBi0OsKYIYnD8AiNPwUSNsewAaZXu8gA) I've manually `declare`-ed the desired function signature to demo how it fixes the issues shown earlier.

An alternative for fixing the return type is to manually cast the MotionValue to the safer type:

```ts
const spring = useSpring(0) as MotionValue<number>
```

This is annoying to have to do and opens opportunity for user error.

Willing to make a PR for this if desired.

controversial:
The current types for `useSpring` use `MotionValue<any>`, even though in practice `useSpring` only handles numeric MotionValues.

Since `useSpring` only accepts numeric inputs and only returns a numeric MotionValue, it should be typed with `MotionValue<number>` in order to be as specific as possible, and encourage as much type safety as possible.

It’s common for developers to enable linter rules that protect against unsafe uses of `any` values in TypeScript, since those values are prone to break type safety guarantees. The current types for `useSpring` create such “unsafe” types even though in practice, the value is always a number. See the following example:

Screenshot 2024-06-17 at 12 06 09 PM

Closes #2495

### DIFF
--- a/packages/framer-motion/src/value/use-spring.ts
+++ b/packages/framer-motion/src/value/use-spring.ts
@@ -31,7 +31,7 @@ import {
  * @public
  */
 export function useSpring(
-    source: MotionValue | number,
+    source: MotionValue<number> | number,
     config: SpringOptions = {}
 ) {
     const { isStatic } = useContext(MotionConfigContext)
@@ -90,7 +90,7 @@ export function useSpring(
 
     useIsomorphicLayoutEffect(() => {
         if (isMotionValue(source)) {
-            return source.on("change", (v) => value.set(parseFloat(v)))
+            return source.on("change", (v) => value.set(v))
         }
     }, [value])
 
--- a/packages/framer-motion/src/value/use-spring.ts
+++ b/packages/framer-motion/src/value/use-spring.ts
@@ -11,6 +11,13 @@ import {
     animateValue,
 } from "../animation/animators/MainThreadAnimation"
 
+// typescript wants only strings passed to parseFloat
+function safeParseFloat(v: unknown) {
+    if (typeof v === "number") return v
+    if (v === null || v === undefined) return NaN
+    return parseFloat(v.toString())
+}
+
 /**
  * Creates a `MotionValue` that, when `set`, will use a spring animation to animate to its new state.
  *
@@ -31,14 +38,14 @@ import {
  * @public
  */
 export function useSpring(
-    source: MotionValue<number> | number,
+    source: MotionValue<unknown> | number,
     config: SpringOptions = {}
 ) {
     const { isStatic } = useContext(MotionConfigContext)
     const activeSpringAnimation = useRef<MainThreadAnimation<number> | null>(
         null
     )
-    const value = useMotionValue(isMotionValue(source) ? source.get() : source)
+    const value = useMotionValue(isMotionValue(source) ? safeParseFloat(source.get()) : source)
     const latestValue = useRef<number>(value.get())
     const latestSetter = useRef<(v: number) => void>(() => {})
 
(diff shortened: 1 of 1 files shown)


mattgperry:
I sort of agree. I think the improved return type is great and we should keep that.

For the input `MotionValue` I'm sort of ambivalent. I agree that the example in the linked PR **shouldn't** be accepted but I do feel like as we support spring animations on values like `"100px"` or whatever that intuitively this should be accepted too:

```javascript
const px = useMotionValue("100px")
useSpring(px)
```

Realistically we're gonna need to keep the input sanitised because it is quite potentially used within Framer user code. For now I'd prefer to leave the input types as-is too. But if this could be amended to improve only the return type then I can merge and publish.

controversial:
Thanks for the feedback @mattgperry - I pushed another commit to address this case!

`useSpring` now accepts any `MotionValue<unknown>` as input, and type-safely transforms that input to a number, to return a `MotionValue<number>`

Sorry it took me a little while to get to this

mattgperry:
Thanks for the updates, I'll merge and publish now!

PR #2703, opened 2024-06-18
fix(isColorString): cannot convert undefined or null to object

kurtextrem:
The current line breaks if the `v` is `null` or `undefined`. TS does not catch it, as `Object.prototype.hasOwnProperty.call` takes in args of type `any` (which `v` is). I assume the type is wrong and should be `Exclude<undefined | null, any>` instead (but that's a TS issue)

### DIFF
--- a/packages/framer-motion/src/value/types/color/utils.ts
+++ b/packages/framer-motion/src/value/types/color/utils.ts
@@ -1,5 +1,5 @@
 import { Color, HSLA, RGBA } from "../types"
-import { floatRegex, isString, singleColorRegex } from "../utils"
+import { floatRegex, isNullish, isString, singleColorRegex } from "../utils"
 
 /**
  * Returns true if the provided string is a color, ie rgba(0,0,0,0) or #000,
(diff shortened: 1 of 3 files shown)


packages/framer-motion/src/value/types/color/utils.ts
```diff
@@ -8,7 +8,7 @@ import { floatRegex, isString, singleColorRegex } from "../utils"
 export const isColorString = (type: string, testProp?: string) => (v: any) => {
     return Boolean(
         (isString(v) && singleColorRegex.test(v) && v.startsWith(type)) ||
-            (testProp && Object.prototype.hasOwnProperty.call(v, testProp))
+            (testProp && !isNullish(v) && Object.prototype.hasOwnProperty.call(v, testProp))
```
mattgperry, line 11: This hasn’t been imported yet. Is it possible to add a test for this behaviour?

kurtextrem: done

iRevolutionDev:
Waiting for this fix!!

mattgperry (approved):

PR #2704, opened 2024-06-18
Adding e2e tests to CI

mattgperry:
This PR moves E2E tests to CircleCI, with a check for passing tests on `main` before publish.


framer-motion-e2e.xml
```diff
@@ -0,0 +1,207 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
nvh, line 1: This should not be committed

packages/framer-motion/src/components/AnimatePresence/__tests__/AnimatePresence.test.tsx
```diff
@@ -445,7 +446,7 @@ describe("AnimatePresence", () => {
     })
 
     test("Fast animations with wait render the child content correctly (strict mode disabled)", async () => {
-        const promise = new Promise<boolean>((resolve) => {
+        await new Promise<boolean>((resolve) => {
```
nvh, line 449: In theory if this reslove with false, we would not catch this now.
If it rejects, than the await throws and that will fail the test, right?

nvh (approved):

packages/framer-motion/jest.config.json
```diff
@@ -14,5 +14,15 @@
...
         "jest-watch-typeahead/testname"
+    ],
+    "reporters": [
+        "default",
+        [
+            "jest-junit",
+            {
+                "outputDirectory": "../../test_reports",
+                "outputName": "framer-motion.xml"
+            }
```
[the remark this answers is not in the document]
mattgperry: Yeah those are outputting correctly via `make test-e2e`

nvh (approved):

PR #2707, opened 2024-06-19
Fixing `dragConstraints={ref}` with active layout animations

mattgperry:
Fixes https://github.com/framer/company/issues/29798

Ensures that draggable components are always remeasured without their transform.

### DIFF
--- a/packages/framer-motion/cypress/integration/drag.ts
+++ b/packages/framer-motion/cypress/integration/drag.ts
@@ -234,9 +234,9 @@ describe("Drag", () => {
             .get("[data-testid='draggable']")
             .trigger("pointerdown", 10, 10)
             .trigger("pointermove", 15, 15)
-            .wait(50)
+            .wait(200)
             .trigger("pointermove", 300, 300, { force: true })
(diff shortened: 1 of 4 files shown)


packages/framer-motion/cypress/integration/drag.ts
```diff
@@ -234,9 +234,9 @@ describe("Drag", () => {
             .get("[data-testid='draggable']")
             .trigger("pointerdown", 10, 10)
             .trigger("pointermove", 15, 15)
-            .wait(50)
+            .wait(200)
             .trigger("pointermove", 300, 300, { force: true })
-            .wait(50)
+            .wait(200)
```
nvh, line 239: Why is this required?

nvh (approved):

packages/framer-motion/cypress/integration/drag.ts
mattgperry: The layout animation I added is like 100 - 200ms long so this delay has to be long enough for them to trigger and fuck up the measurements

PR #2710, opened 2024-06-24
fix: window is not defined in nextjs

apades:
My env is next@14.2.3, next-intl@3.13.0, and i use in client component, i don't kown why it trigger this line "documentNode.mount(window)", i have drop to next@13 but it still trigger
微信截图_20240624142023

微信截图_20240624143220


syi0808:
Apart from the issue. I think client component will pre-rendered in server side too.

mattgperry:
I did a similar but more root fix with https://github.com/framer/motion/commit/8f04d53d11bcec144617170c28ef0d03e8e45a2c#diff-dbfc1fdc5de528bff405b87663a1b834418f4f8c0f89658aad624ff504a11e7a - can you confirm this works with `11.3.10`?

apades:
Yes, it is work, thanks for your work

PR #2735, opened 2024-07-18
Perf: Improving performance of calculated transform comparison

mattgperry:
This PR replaces the calculated transform comparison from creating a transform string and comparing the string, to comparing the `Delta` features directly.

This makes the delta comparison 100x faster in Safari and 20x faster in Chrome for an overall 1% per-frame reduction in script time.

There's also a little bit of a drive-by where projection metrics are recorded only if `MotionDebug` is present on `window`.


packages/framer-motion/src/projection/node/create-projection-node.ts
```diff
@@ -1390,32 +1407,35 @@ export function createProjectionNode<I>({
              * layout reprojection or the final bounding box.
              */
             calcBoxDelta(
-                this.projectionDelta,
+                this.projectionDelta!,
```
mattgperry, line 1410: Moving object creation to `createProjectionDeltas()` removed the implicit understanding this object has now been created

adamseckel (approved):

PR #2747, opened 2024-07-31
Optimise `transform` string

mattgperry:
Currently, if any transform is non-default, all non-default transforms get included in the `transform` string:

```
{ x: 0 } => "none"
{ x: 0, scale: 1.2 } => "translateX(0px) scale(1.2)"
```

This PR changes it so these no-op transforms are excluded.

```
{ x: 0 } => "none"
{ x: 0, scale: 1.2 } => "scale(1.2)"
```

This results in up to 30% faster effects in the way we have it configured inside Framer (where transforms might be provided as `MotionValue`s but always stay at default values).

### DIFF
--- a/packages/framer-motion/src/render/html/utils/build-transform.ts
+++ b/packages/framer-motion/src/render/html/utils/build-transform.ts
@@ -3,6 +3,7 @@ import { MotionProps } from "../../../motion/types"
 import { getValueAsType } from "../../dom/value-types/get-as-type"
 import { numberValueTypes } from "../../dom/value-types/number"
 import { ResolvedValues } from "../../types"
+import { HTMLRenderState } from "../types"
 
 const translateAlias = {
     x: "translateX",
@@ -20,7 +21,8 @@ const numTransforms = transformPropOrder.length
  * providing a transformTemplate function.
  */
 export function buildTransform(
-    transform: ResolvedValues,
+    latestValues: ResolvedValues,
+    transform: HTMLRenderState["transform"],
     transformTemplate?: MotionProps["transformTemplate"]
 ) {
     // The transform string we're going to build into.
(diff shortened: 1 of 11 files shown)


packages/framer-motion/src/value/use-will-change/__tests__/will-change.ssr.test.tsx
```diff
@@ -91,7 +91,7 @@ function runTests(render: (components: any) => string) {
         )
     })
 
-    test("will-change correctly not applied when isStatic", () => {
+    test.only("will-change correctly not applied when isStatic", () => {
```
adamseckel, line 94: only!

packages/framer-motion/src/render/html/utils/build-transform.ts
```diff
@@ -18,22 +21,43 @@ const numTransforms = transformPropOrder.length
...
+        }
+
+        if (!valueIsDefault) {
+            transformIsDefault = false
+        }
+
+        if (!valueIsDefault || transformTemplate) {
             const transformName = translateAlias[key] || key
-            transformString += `${transformName}(${transform[key]}) `
+            const valueAsType = getValueAsType(value, numberValueTypes[key])
```
adamseckel, line 55: Is this a just a drive by to use a new shared helper?

packages/framer-motion/src/render/html/utils/build-transform.ts
```diff
@@ -18,22 +21,43 @@ const numTransforms = transformPropOrder.length
...
+            transformString += `${transformName}(${valueAsType}) `
+
+            if (transformTemplate) {
+                transform[key] = valueAsType
+            }
```
adamseckel, line 60: I'm curious why didn't have to do this before?

adamseckel (approved):
Nice one. Don't forget to cleanup the test.only before merging!

packages/framer-motion/src/render/html/utils/build-transform.ts
mattgperry: No this was moved in from `build-styles` so we basically only run it if the transform isn't default or if we need to pass it to `transformTemplate`

PR #2752, opened 2024-08-07
Optimised appear improvements

mattgperry:
This PR adds three improvements to optimised appear animations.

### Improve appear animation check

Currently, to check if appear animations exist we just check for `data-appear-id` on an element. But this can exist in the DOM even if the appear animation exists on a different breakpoint. Now we check the `elementId` in a `Set` that is only added to when an optimised animation actually starts.

### Cancelling only `transform` animations

Previously, any layout animation that is a descendant of an element with, or has itself, an optimised appear animation, would cancel all optimised appear animations to correctly measure the DOM.

Now, only `transform` appear animations trigger this cancellation, and only `transform` appear animations are cancelled.

### Changing API

Recent improvements to SSG versioning mean I can change the API to be backwards incompatible, so I've improved the `window` namespacing with a `Motion` prefix and removed unused vars.


dev/html/src/imports/script-assert.js
```diff
@@ -92,6 +94,26 @@ window.Assert = {
...
+        const computedStyle = window.getComputedStyle(element).transform
+
+        style = style.replace("translateX(", "").replace(")", "")
+
+        const matrixType = computedStyle.includes("3d") ? "3d" : "2d"
+
+        let x = 0
+        const matrixValues = computedStyle
+            .match(/matrix.*\((.+)\)/)[1]
+            .split(", ")
```
shuangq, line 108: I wonder if it would be better to construct a dom matrix with the string to avoid the parsing, although that does come with the cost of constructuring a matrix so not sure if it's a good idea 🤔

dev/html/src/imports/script-assert.js
```diff
@@ -92,6 +94,26 @@ window.Assert = {
...
+        if (matrixType === "2d") {
+            x = parseFloat(matrixValues[4])
+        } else {
+            x = parseFloat(matrixValues[12])
+        }
+        console.log(x, parseFloat(style))
```
shuangq, line 114: This needs to be cleanup?

packages/framer-motion/src/projection/node/create-projection-node.ts
```diff
@@ -108,21 +108,25 @@ function resetDistortingTransform(
...
 
     const { visualElement } = projectionNode.options
 
     if (!visualElement) {
         return false
-    } else if (getOptimisedAppearId(visualElement)) {
+    } else if (
+        window.MotionHasOptimisedTransformAnimation!(
+            getOptimisedAppearId(visualElement)
+        )
```
shuangq, line 122: Is it safer to do 
```suggestion
        window.MotionHasOptimisedTransformAnimation?.(
            getOptimisedAppearId(visualElement)
        )
```

shuangq (approved):

dev/html/src/imports/script-assert.js
mattgperry: It's just a test script so the cost isn't a big problem but likewise it can just be a bit shit

packages/framer-motion/src/projection/node/create-projection-node.ts
mattgperry: So this function is wrapped in a check for these functions - `?.` is safer but in this case I think it's ok to avoid the overhead (?. gets transpiled to quite a noisy series of checks and we want to quickly look up a tree here)

shuangq: > So this function is wrapped in a check for these functions 

ah cool, missed that bit.

packages/framer-motion/src/animation/optimized-appear/store.ts
```diff
@@ -4,3 +4,5 @@ export interface AppearStoreEntry {
 }
 
 export const appearAnimationStore = new Map<string, AppearStoreEntry>()
+
+export const elementsWithAppearAnimations = new Set<string>()
```
kurtextrem, line 8: If the appear effect is done, shouldn't we clear the set to avoid memory leaks? (or use `WeakSet/Map` instead?)

mattgperry: (The reason they're strings vs Element look ups is partly in case of hydration errors and partly because of pre-hydration checks)

kurtextrem: ah yeah makes sense I thought we store elements in here, then I guess it's kinda okay-ish

PR #2755, opened 2024-08-09
Allow animation `startTime` to be set manually

mattgperry:


packages/framer-motion/src/animation/animators/AcceleratedAnimation.ts
```diff
@@ -266,6 +267,17 @@ export class AcceleratedAnimation<
...
 
+    get startTime() {
+        const { resolved } = this
+        if (!resolved) return null
+
+        const { animation } = resolved
+
+        // Coerce to number as TypeScript incorrectly types this
+        // as CSSNumberish
+        return animation.startTime as number
```
adamseckel, line 278: Is this *just* a typescript problem or would a simple `if (!isNumber(animation.startTime)) return null` be more future proof?

adamseckel (approved):

packages/framer-motion/src/animation/animators/AcceleratedAnimation.ts
mattgperry: Web Animations L2 spec does show times as CSSNumberish but it isn't ready yet so could still be changed - I also struggle believing they will make this backwards incompatible

PR #2757, opened 2024-08-09
Selectively cancel parts of tree with layout animation/optimised transform conflicts

mattgperry:
Currently, when any layout animation triggers and it detects it's in a part of a tree running optimised `transform` animations, it will cancel **all** optimised `transform` animations (to prevent style thrashing).

This PR changes the strategy to accept a potential level of style thrashing in return for only disabling optimised animations on the specific shards of tree involved in an animation. This will lead to fewer visual stutters.


dev/html/public/optimized-appear/defer-handoff-layout-sibling-transform.html
```diff
@@ -0,0 +1,145 @@
...
+                                        .getElementById("layout-box")
+                                        .getAnimations().length
+
+                                    if (
+                                        numAnimationsOnA !== 1 ||
+                                        numAnimationsOnB !== 0
+                                    ) {
+                                        showError(
+                                            document.getElementById("box"),
+                                            "Layout animation on B cancelled optimised animation on A"
```
shuangq, line 99: nit: the error can be different, right?
If A !== 1: cancelled by B,
If B !== 0: optimised animation not cancelled?

packages/framer-motion/src/animation/optimized-appear/start.ts
```diff
@@ -65,20 +65,26 @@ export function startOptimizedAppearAnimation(
...
-            })
-
-            window.MotionHandoffCancelAll = undefined
+        window.MotionCancelOptimisedTransform = (elementId: string) => {
+            const animationId = appearStoreId(elementId, "transform")
+            const data = appearAnimationStore.get(animationId)
+
+            if (data) {
+                data.animation.cancel()
+                appearAnimationStore.delete(animationId)
```
shuangq, line 86: This is out of scope, but I'm curious why .delete is not handled by `.cancel()` internally? Is there a call site where you want to cancel but not deleting the entry?

shuangq (approved):

packages/framer-motion/src/animation/optimized-appear/start.ts
mattgperry: `animation` is the WAAPI animation that we're keeping in an external store - so it's more like this function is that wrapper function you describe

dev/html/public/optimized-appear/defer-handoff-layout-sibling-transform.html
mattgperry: yeah good suggestion

PR #2762, opened 2024-08-13
Fix unit conversion with external `MotionValue`s

mattgperry:
This PR ensures that keyframes for all DOM animations, not just those on `MotionValue`s "owned" by a `motion` value, are correctly resolved. Thereby fixing unit conversion on externally-defined `MotionValue`s.

This also ensures that all DOM animations are resolved at the same time, better keeping them in sync.


packages/framer-motion/src/render/dom/DOMKeyframesResolver.ts
```diff
@@ -29,22 +29,16 @@ export class DOMKeyframesResolver<
...
-            true
-        )
+        super(unresolvedKeyframes, onComplete, name, motionValue, element, true)
     }
 
     readKeyframes() {
         const { unresolvedKeyframes, element, name } = this
 
-        if (!element.current) return
+        if (!element || !element.current) return
```
adamseckel, line 41: I'm curious, is there some transpilation reason this is preferable to:
`!element?.current`

adamseckel (approved):

packages/framer-motion/src/render/dom/DOMKeyframesResolver.ts
mattgperry: Yeah `?.` transpiles like this https://babeljs.io/repl#?browsers=defaults%2C%20not%20ie%2011%2C%20not%20ie_mob%2011&build=&builtIns=false&corejs=3.21&spec=false&loose=false&code_lz=LYTw_AdA9gDgLgSygOwIYBtJwBYOQcyA&debug=false&forceAllTransforms=false&modules=false&shippedProposals=false&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Creact%2Cstage-2&prettier=false&targets=&version=7.25.3&externalPlugins=&assumptions=%7B%7D

PR #2767, opened 2024-08-16
Resume cancelled optimised animations where possible

mattgperry:
This PR builds off https://github.com/framer/motion/pull/2757

When cancelling optimised appear animations we can encounter two visual artefacts:
1. "Slippage" in Safari caused by its incorrect timing model on accelerated animations
2. Stutter induced by main thread jank

The prior PR ensures we only cancel WAAPI animations on trees where layout animations take place. This PR resumes those animations immediately after measurement on nodes where we aren't performing layout animations, decreasing the period of time on jank and reducing the number of nodes with visible slippage.


packages/framer-motion/src/projection/node/create-projection-node.ts
```diff
@@ -1972,6 +1974,7 @@ export function createProjectionNode<I>({
 
         clearSnapshot() {
             this.resumeFrom = this.snapshot = undefined
+            this.hasCheckedOptimisedAppear = false
```
shuangq, line 1977: Why are we resetting this?

packages/framer-motion/src/projection/node/create-projection-node.ts
```diff
@@ -726,6 +727,7 @@ export function createProjectionNode<I>({
             frameData.isProcessing = true
             steps.update.process(frameData)
             steps.preRender.process(frameData)
+            window.MotionResumeOptimisedAnimations?.()
```
shuangq, line 730: I think this could use a comment to explain the timing.

shuangq (approved):

packages/framer-motion/src/projection/node/create-projection-node.ts
mattgperry: Because the animations are only suspended we would need to recheck on each subsequent layout animation. I'm gonna take another look at this PR though and see about making a better cleanup function so we know when we can stop doing this.

PR #2772, opened 2024-08-23
Ensure optimised appear animations are cancelled when styles differ

mattgperry:
In Framer we provided two `MotionValue`s for each component:
- via `values`, this value handles `animate`/`whileHover` etc animations
- via `style`, this value is rendered

The value provided to `style` is composed from other `MotionValue`s each handling an effect, for instance scroll transform etc.

If this rendered value falls out of sync with the one being used by Motion for its internal animations, then we're in a situation where the optimised appear animation must also have fallen out of sync and we need to cancel it. This PR implements this.


packages/framer-motion/src/animation/optimized-appear/start.ts
```diff
@@ -65,10 +68,26 @@ export function startOptimizedAppearAnimation(
...
+            /**
+             * Keep a map of elementIds that have started animating. We check
+             * via ID instead of Element because of hydration errors and
+             * pre-hydration checks. We also actively record IDs as they start
+             * animating rather than simply checking for data-appear-id as
+             * this attrbute might be present but not lead to an animation, for
+             * instance if the element's appear animation is on a different
+             * breakpoint.
+             */
+            if (!valueName) {
```
shuangq, line 86: In what cases do we not provide a value name for the optimized animation?

packages/framer-motion/src/render/VisualElement.ts
```diff
@@ -397,7 +397,11 @@ export abstract class VisualElement<
             this.removeFromVariantTree = this.parent.addVariantChild(this)
         }
 
-        this.values.forEach((value, key) => this.bindToMotionValue(key, value))
+        this.values.forEach((value, key) => {
+            if (!this.valueSubscriptions.has(key)) {
+                this.bindToMotionValue(key, value)
+            }
```
shuangq, line 403: I also don't fully understand this change, as `bindToMotionValue` seems to be designed to handle cases where a value subscription is present as well. 🤔

packages/framer-motion/src/animation/optimized-appear/start.ts
```diff
@@ -87,17 +109,39 @@ export function startOptimizedAppearAnimation(
...
+
+            const removeSyncCheck = value.on(
+                "change",
+                (latestValue: string | number) => {
+                    if (externalAnimationValue.get() !== latestValue) {
+                        window.MotionCancelOptimisedAnimation?.(
+                            appearId,
+                            valueName
+                        )
+                        removeSyncCheck()
```
shuangq, line 138: I'm not super familiar with the `value.on` API, why do we need to call `removeSyncCheck` again?

dev/html/public/optimized-appear/defer-handoff-external-values.html
```diff
@@ -0,0 +1,119 @@
...
+                    onAnimationComplete: () => {
+                        const box = document.getElementById("box")
+                        const { left } = box.getBoundingClientRect()
+
+                        if (Math.round(left) !== 300) {
+                            showError(
+                                box,
+                                `optimised animation not cancelled by external value mismatch with rendered style`
+                            )
+                        }
```
shuangq, line 76: The opacity animation should not be canceled, right? Do we also want to check that as well?

shuangq (approved):
Left some questions but looks good overall.

packages/framer-motion/src/animation/optimized-appear/get-appear-id.ts
```diff
@@ -1,8 +1,8 @@
-import { VisualElement } from "../../render/VisualElement"
 import { optimizedAppearDataAttribute } from "./data-id"
+import { WithAppearProps } from "./types"
 
 export function getOptimisedAppearId(
-    visualElement: VisualElement
+    visualElement: WithAppearProps
 ): string | undefined {
-    return visualElement.getProps()[optimizedAppearDataAttribute]
+    return visualElement.props[optimizedAppearDataAttribute]
```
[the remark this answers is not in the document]
mattgperry: Practically it doesn't make a difference but it just made the types subset of `VisualElement` simpler

packages/framer-motion/src/animation/optimized-appear/start.ts
mattgperry: In this instance it will check for *any* optimised animation - for instance when determining if we fire animations on layout effect or effect

packages/framer-motion/src/render/VisualElement.ts
mattgperry: Ah good catch - this is a legacy from fixing the same bug that other check fixed (that I discovered while making this PR)

packages/framer-motion/src/animation/optimized-appear/start.ts
mattgperry: Think of it like 

```
function doThing() {
  if (someCondition) {
    element.removeEventListener("mousestart", doThing)
  }
}

element.addEventListener("mousestart", doThing)
```

dev/html/public/optimized-appear/defer-handoff-external-values.html
mattgperry: Yeah I'll add a check - it should be cancelled by mismatches as well

PR #2774, opened 2024-08-23
Resume optimised animations where possible

mattgperry:
When cancelling optimised appear animations we can encounter two visual artefacts:

"Slippage" in Safari caused by its incorrect timing model on accelerated animations
Stutter induced by main thread jank
The prior PR ensures we only cancel WAAPI animations on trees where layout animations take place. This PR resumes those animations immediately after measurement on nodes where we aren't performing layout animations, decreasing the period of time on jank and reducing the number of nodes with visible slippage.

Replaces https://github.com/framer/motion/pull/2767

### DIFF
--- a/dev/html/public/optimized-appear/defer-handoff-block.html
+++ b/dev/html/public/optimized-appear/defer-handoff-block.html
@@ -36,7 +36,7 @@
             const { matchViewportBox } = window.Assert
             const root = document.getElementById("root")
 
-            const duration = 4
+            const duration = 1
             const x = motionValue(0)
             const xTarget = 500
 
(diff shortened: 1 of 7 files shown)


dev/html/public/optimized-appear/defer-handoff-block.html
```diff
@@ -36,7 +36,7 @@
             const { matchViewportBox } = window.Assert
             const root = document.getElementById("root")
 
-            const duration = 4
+            const duration = 1
```
mattgperry, line 39: This test was too long for errors to be caught

PR #2778, opened 2024-08-29
Fix RSC integration

mattgperry:
This PR ensures the `"use client"` directive is output in our generated ES files.

`motion.div` still needs re-exporting due to limitations with RSC.


kurtextrem:
Does it make sense to make the `motion(...)` a deprecated call and only use the better tree-shakeable variant from now on?

kurtextrem (approved):

PR #2787, opened 2024-09-04
Adding `motion.create()`

mattgperry:
In [this PR](https://github.com/framer/motion/pull/2778) we added a new entrypoint for `motion` components that's compatible with React Server Components.

The issue with `import * as x` is you remove the ability to support `Proxy` or other dynamically-assigned objects so the pattern of supporting `motion(Component)` and `motion.div` breaks.

This PR deprecates `motion(Component)` in favour of `motion.create(Component)`


kurtextrem (approved):
Makes perfect sense to me!

PR #2802, opened 2024-09-14
fix: remove unnecessary dependency from safeToRemove callback

### LINKED ISSUE #2801: [BUG] `safeToRemove` Callback Recreated on Parent Re-render

✅

**2. Describe the bug**

The `usePresence` hook provides a `safeToRemove` callback. However, if the parent re-renders during the exit animation and I add `safeToRemove` to the `useEffect` dependency array, the function is recreated, causing the effect to trigger unnecessarily.

This happens despite #2431 wrapping the callback with the `useCallback` hook.

After looking into it, I found that the `onExitComplete` function in `PresenceContext` is being recreated unnecessarily. Take a look at the PR I'll be submitting.

https://codesandbox.io/p/sandbox/safetoremove-re-create-issue-nzvmfy

**4. Steps to reproduce**

1. Go to [codesandbox](https://codesandbox.io/p/sandbox/safetoremove-re-create-issue-nzvmfy).
2. Run the app and check the console log.

**5. Expected behavior**

The `safeToRemove` function should remain the same, regardless of whether the parent is re-rendered or not.

custardcream98:
`onExitComplete` is recreated every time the parent component rerenders when the `presenceAffectsLayout` prop of `AnimatePresence` is set to `true`, which unnecessarily recreates the `safeToRemove` callback.

I’m considering whether to memoize the `onExitComplete` function that is put into the context in `PresenceChild`, but I’m not sure what side effects this might have.

Please leave a comment.

closes #2801


mattgperry:
Thanks for the PR!

PR #2803, opened 2024-09-14
fix: Improve types to include css variables

dca123:
In the [docs](https://www.framer.com/motion/component/#%23%23animating-css-variables) regarding css variables, it's recommended to cast it as `any`. 

This PR should allow you to add css variables, i.e "--${anthingGoesHere}". If the initial "--" is not there, an error is thrown !

Right now it only allows for strings & numbers as the properties.

Valid
```
{
  "--myColor": "#eeeeee"
}
```

Invalid
```
{
  "myColor": "#eeeeee"
}
```

### DIFF
--- a/packages/framer-motion/src/motion/types.ts
+++ b/packages/framer-motion/src/motion/types.ts
@@ -119,6 +119,10 @@ export type ResolveLayoutTransition = (
     info: RelayoutInfo
 ) => Transition | boolean
 
+type CSSVariables = {
+  [key: `--${string}`]: string;
+};
+
 /**
  * @public
  */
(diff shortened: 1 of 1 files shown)


Theghost256 (approved):

mattgperry:
Great PR idea though to accept I would need a test that mixes `x: 100` with `--custom-var` without type errors. 

It might need extending `type TargetProperties` instead.

mattgperry:
Sorry just closing this out as I'm working on implementing this now.

mattgperry:
PR: https://github.com/framer/motion/pull/2825

dca123:
awww i was just about to hop on and get on that, thanks for the fix @mattgperry  !

PR #2804, opened 2024-09-18
Animate values when child variant definitions change

mattgperry:
Ensure that when the **value** of a variant changes, we animate to the latest value.

```jsx
<motion.div initial="a">
  <motion.div variants={{ a: { x }}} />
</motion.div>
```


packages/framer-motion/src/render/utils/get-variant-context.ts
```diff
@@ -0,0 +1,43 @@
+import { VisualElement } from "../VisualElement"
```
mattgperry, line 1: This is a bit of a drive-by moving the method out into a function. To improve code-splitting I want to prefer functions over methods where possible.

Issue #2763, opened 2024-08-13
[BUG] will-change: opacity changes stacking context of elements


tim-richter:
**2. Describe the bug**

`will-change: opacity` is applied to elements, if `AnimatePresence initial={false}`, and it alters the appearance of svg's (and probably other elements as well)

According to MDN (https://developer.mozilla.org/en-US/docs/Web/CSS/will-change), the will-change property should not be used lightly and prematurely. I think it's probably fine to use in a lot of instances, but especially this paragraph is important for the behavior:

> Be aware, that will-change may actually influence the visual appearance of elements, when used with property values, that create a stacking context (e.g. will-change: opacity), as the stacking context is created up front.

So framer-motion actually introduced a breaking change in 11.3, because the appearance of all elements that use the property might change. I think it would be better to apply `will-change` only optionally, in cases where css animations are cleary slow to the user, and not generally for all elements that animate via css.

**3. IMPORTANT: Provide a CodeSandbox reproduction of the bug**

https://stackblitz.com/edit/vitejs-vite-sgwdxh?file=src%2FButton.tsx

**4. Steps to reproduce**

1. Use `AnimatePresence` with `initial={false}`
2. Use any motion element as child (motion.span for example)
3. Animate the opacity
4. Check out the element in DevTools and see that `will-change: opacity` is set on the element
5. Notice weird things happening (element displays differently, svg elements are blurred etc.)

**5. Expected behavior**

Expected behavior would be to not see any visual regressions when using the combination above

**6. Video or screenshots**

It's probably hard to see via screenshare, but the blurry icon is very visible on the monitor and also screenshot tests:

[Kooha-2024-08-13-15-21-33.webm](https://github.com/user-attachments/assets/7ff7f0df-7330-4e68-a082-bbd253019cf3)

**7. Environment details**

NixOS, Chrome Browser Version 127.0.6533.88
Also happens in firefox

mraballard:
@tim-richter i also noticed this issue affecting `motion.div` in my app (on chrome). i noticed dropdowns are stacking behind the motion.div due to the will-change opacity property. changing to a regular div fixes it.

ValentinH:
We are also facing this issue after updating to the latest version. the `will-change` property used to be `auto` on all our `motion.div` and now it's is `opacity`. Therefore, it creates a new stacking context and all elements that were supposed to go outside of the div are now stuck inside it.

[...]

PR #2811, opened 2024-09-24
Stop removing values from `will-change` once they finish animating

mattgperry:
Due to Safari's subpixel rendering issues, stopping elements from being layers can lead to a visual flickering effect. This PR ensures that once a value has been added to `will-change` it stays there.

### DIFF
--- a/packages/framer-motion/package.json
+++ b/packages/framer-motion/package.json
@@ -98,7 +98,7 @@
     "bundlesize": [
         {
             "path": "./dist/size-rollup-motion.js",
-            "maxSize": "33.95 kB"
+            "maxSize": "33.82 kB"
         },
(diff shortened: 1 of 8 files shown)


shuangq (approved):
Code looks good 👍 I'm curious can we do this only for safari? Or would that have no significant impact?

mattgperry:
@shuangq Koen also raised the point that it might be generally better in terms of performance to not keep swapping things on and off the GPU - but also there's not a good way to do this just for Safari either

bumling:
Will this perhaps fix this https://github.com/framer/motion/issues/2763?

bumling:
> Will this perhaps fix #2763?

It didn't

PR #2812, opened 2024-09-25
Support for `linear()` easing function

mattgperry:
Currently, when a WAAPI animation is defined with an easing function:

```javascript
animate(
  element,
  { opacity: 0 },
  { ease: (p) => p ** p }
) 
```

Or an easing name not supported by WAAPI:

```javascript
animate(
  element,
  { opacity: 0 },
  { ease: "anticipate" }
) 
```

Then Motion will pre-generate keyframes for this animation and run it with `linear` easing.

This can have poor performance characteristics: https://issues.chromium.org/issues/41491098

Likewise for complex values like `clip-path` and `filter` this can have the overhead of number -> string conversion.

This PR adds support for the `linear()` easing function. So when an easing isn't supported by WAAPI, we generate a `linear()` easing function instead of pre-generating keyframes.

### DIFF
--- a/packages/framer-motion/src/animation/animators/AcceleratedAnimation.ts
+++ b/packages/framer-motion/src/animation/animators/AcceleratedAnimation.ts
@@ -1,3 +1,6 @@
+import { anticipate } from "../../easing/anticipate"
+import { backInOut } from "../../easing/back"
+import { circInOut } from "../../easing/circ"
 import { EasingDefinition } from "../../easing/types"
 import { DOMKeyframesResolver } from "../../render/dom/DOMKeyframesResolver"
 import { ResolvedKeyframes } from "../../render/utils/KeyframesResolver"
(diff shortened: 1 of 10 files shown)


packages/framer-motion/src/animation/animators/AcceleratedAnimation.ts
```diff
@@ -159,6 +168,22 @@ export class AcceleratedAnimation<
...
+         */
+        if (
+            typeof ease === "string" &&
+            supportsLinearEasing() &&
+            ease in unsupportedEasingFunctions
+        ) {
+            ease =
+                unsupportedEasingFunctions[
+                    ease as keyof typeof unsupportedEasingFunctions
+                ]
```
adamseckel, line 184: If you made a type guard like:
```typescript
function isUnsupportedEase(key: string): key is keyof unsupportedEasintFunctions {
  return key in unsupportedEasintFunctions
}
```

You could avoid the casts.

adamseckel (approved):

packages/framer-motion/src/animation/animators/waapi/utils/linear.ts
```diff
@@ -0,0 +1,19 @@
+import { EasingFunction } from "../../../../easing/types"
+import { progress } from "../../../../utils/progress"
+
+// Create a linear easing point for every 10 ms
```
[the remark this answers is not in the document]
mattgperry: just somewhere between 8 (120fps) and 16 (60fps) that looks quite good without being too accurate.

packages/framer-motion/src/animation/animators/AcceleratedAnimation.ts
mattgperry: ty will implement

Issue #2816, opened 2024-09-29
[BUG]Blurriness that occurs upon the completion of animation


axuj:
**1. Read the [FAQs](#faqs) 👇**

**2. Describe the bug**

#2811 Currently, there is blurriness that occurs upon animation completion in Chrome and Edge browsers..

**3. IMPORTANT: Provide a CodeSandbox reproduction of the bug**

https://codesandbox.io/p/devbox/dgs6kk?file=%2Fsrc%2FApp.tsx

**4. Steps to reproduce**

1. drag button
2. set willChange='auto'

**5. Expected behavior**

**6. Video or screenshots**

![image](https://github.com/user-attachments/assets/01bdbf16-01cf-4b41-8207-bc04d15fbebf)
**7. Environment details**

edge 129.0.2792.65   
chrome 129.0.6668.60

AnYiEE:
It will also cause the background of the child element with `backdrop-filter: blur` added to not blur properly. You can check [this comment](https://github.com/nextui-org/nextui/issues/3817#issuecomment-2380897479). Similarly, the problem disappears after setting `will-change` to `auto` (that is, using version 11.5.6 and earlier).

mraballard:
[fyi ](https://github.com/framer/motion/issues/2763#issuecomment-2386258633)

PR #2824, opened 2024-10-08
Remove auto `will-change`

mattgperry:
Fixes https://github.com/framer/motion/issues/2816
Fixes https://github.com/framer/motion/issues/2763

Auto `will-change` was an attempt at semi-intelligently applying `will-change` for the right values when animations trigger. 

Due to various rendering issues (the above tickets, https://github.com/framer/motion/pull/2818#issuecomment-2392690623 and things we've seen in Framer) I think the safer thing to do is to remove this and leave it to users.

This PR removes auto `will-change` but allows users to opt-in as before via `useWillChange`.


Theghost256 (commented):
Thank you for letting me help

adamseckel (approved):

robmeede:
Any chance this PR will be part of a release soon?

mattgperry:
@robmeede Probably today or tomorrow but you can still override with `style={{ willChange: "auto" }}`

ynamite:
@mattgperry this fix doesn't work with useAnimate (or rather animate) though, right? Is there a similar fix? Right now I'm setting will change manually after a setTimeout. Fugly, but it works :) 

Out of curiosity, how big of a performance impact does `will-change` have or what was the rationale behind adding auto will-change?

mattgperry:
@ynamite That's right actually, you'd either have to pass it through as an animated value or change the element to a `motion` component. But yeah this will be fixed in the next release.

Interested in what was breaking for you with it?

Ideally this would be something I could take off people's plates. As the performance benefit can differ between Safari/Chrome/devices. And making too many layers can be detrimental. 

I'm removing it now actually **because** users are reporting various problems with it, but then manually adding it back in Framer because we're seeing performance improvements, especially during scroll, when its added.

Personally I don't believe the **measured** performance benefits are that great but the perception was that framer.com was smoother with it and sometimes that's more important. But my feeling now is I can't automate it in a way that's useful for a broad range of people, hence this PR.

ynamite:
Ah I see, thanks. I believe I tried solving it as you suggest, but couldn't get it to work. But maybe I'd done something wrong.

And thanks also for the explanation. In my case I'm handling route transitions with Framer Motion. Some pages have modals (triggered after a click on a button) and these modals have position fixed. It's not a problem when the whole view is transitioned, but if it's only certain components of a view, the modals would then be positioned relative to the parent that is being transitioned (stacking context).

I actually quite like the idea of auto will-change, but at the very least, it should either be configurable (default auto) to turn it off or Framer Motion would have to remove it automatically when a transition ends. Similar to how GSAPs `clear: all` property works. Come to think of it, something like `clear: all` would be awesome.

Issue #2508, opened 2024-02-06
[BUG] AnimatePresence popLayout mode does not work within a shadow dom


disambiguator:
# Description
If we use AnimatePresence inside a shadow DOM, the popLayout features don't work. This is because we're unable to find the temporary styles added, because shadow dom won't apply styles found in document.head.

Code sandbox repro: https://codesandbox.io/p/sandbox/framer-motion-shadow-dom-f7gkk6?file=%2Fsrc%2FApp.js%3A89%2C2

**4. Steps to reproduce**

1. Open the code sandbox link
2. Click the example to run the animation. You'll notice the popLayout behavior is not occurring.
3. Uncomment line 86, skipping the shadow DOM.
4. popLayout starts working again.

# Technical guess of the problem
From what I understand, popLayout works by applying a temporary class which positions an element as absolute during exit transition. This does not work in shadow DOMs due to the way the style is applied. I _think_ [the code here](https://github.com/framer/motion/blob/309286da2258c07ab8da2ef60f2df9a2e9bb37db/packages/framer-motion/src/components/AnimatePresence/PopChild.tsx#L78) is adding the temporary class, but it's hardcoded to using `document.head`, which won't be available to the shadow dom.

cymptom:
That guess at the cause seems likely :( 

If anyone has a workaround for this issue, please share 🙏

ParkerSm1th:
Wanted to bump this issue, currently experiencing it and would love to continue using framer motion in my project.

rdamore2812:
Bump too. Would love any workaround!

tehbelinda:
I made a PR to fix this: https://github.com/framer/motion/pull/2830, works if I pass in the shadowRoot directly.
Would appreciate any feedback!

dylank:
@mattgperry can you help get this merged?

tehbelinda:
If anyone else needs this in the meantime, I made a tag with all the dist files included, so you can reference it directly from your package.json, i.e.

`npm install 'https://gitpkg.vercel.app/tehbelinda/motion/packages/framer-motion?v11.11.10-shadow-dom' --save`

(`yarn add` should work too but I haven't tested that one)

PR #2830, opened 2024-10-11
Add support for AnimatePresence when using shadow DOM

tehbelinda:
By default this was using document.head, but that is not always available to where it's rendered, e.g. when using the shadow DOM. In such cases, this change allows for the shadow root to be passed to the new `parentDOM` prop. This change is backwards compatible since we default to `document.head` when not specified.

Fixes https://github.com/framer/motion/issues/2508


dylank:
Hey @mattgperry, do you know what the process is to get this change merged?

tehbelinda:
hi @mattgperry the pipeline is passing and I just updated the latest conflicts on the CHANGELOG, let me know if you are able to review and approve this or is there someone else I should ping? thanks!

dylank:
@kurtextrem is this something you can help with?

connormdavis:
@SaeWooKKang @mattgperry this is a blocker for me would be huge to get this merged 🙏

kurtextrem:
Hey! I had a brief look, it looks good to me. I've forwarded the PR.

packages/framer-motion/src/components/AnimatePresence/__tests__/AnimatePresence.test.tsx
```diff
@@ -346,6 +346,39 @@ describe("AnimatePresence", () => {
         return await expect(promise).resolves.toBe(3)
     })
 
+    test("Can cycle through multiple components with mode === 'popLayout' and dom", async () => {
```
mattgperry, line 349: I'm not 100% on what this test is trying to achieve.

Is it more expected based on the changes that we should check that `style` blocks are being correctly added to the `testDom` elements?

mattgperry (changes requested):
Thanks for the PR! 

It would be good to duplicate `packages/framer-motion/cypress/integration/animate-presence-pop.ts` and the `animate-presence-pop` and `dev/react/src/tests/animate-presence-pop`, and rerunning all the tests in a situation that uses a new `ShadowRoot`.

Additionally, do we expect other elements to be used other than `ShadowRoot`? It feels to me like `shadowRoot` or `root` might be a more indicative prop name.

packages/framer-motion/src/components/AnimatePresence/__tests__/AnimatePresence.test.tsx
tehbelinda: I was aiming for that originally but there doesn't seem to be a way to validate the style block is there, so I resorted to having something that at least exercised the prop. But now that I've added the cypress tests per the other comment, that covers it more thoroughly so I'll just remove this. Let me know if you think there should still be something here

tehbelinda:
Thanks for taking a look @mattgperry ! I've added the tests as requested and updated the prop to `root`

btw as an alternative to enabling the experimental flag to test the shadow dom in cypress, I did try upgrading cypress to 5.2.0 where the behaviour is enabled by default. The tests still passed, but I was wary of mixing the upgrading of a major version in with this in case there was something else I've missed

Issue #2578, opened 2024-03-20
[BUG] Unable to animate svg text elements


blowery:
**1. Read the [FAQs](#faqs) 👇**

**2. Describe the bug**

Framer Motion is unable to animate svg `text` elements. It appears to push content into a `children` attribute instead of altering the contents of the `text` element.

**3. IMPORTANT: Provide a CodeSandbox reproduction of the bug**

https://codesandbox.io/p/sandbox/framer-motion-animate-content-forked-j3y3qt?file=%2Fsrc%2FApp.tsx%3A20%2C11

**4. Steps to reproduce**

1. Load the sandbox
2. Notice the count does not animate
3. Inspect the text element in dev tools
4. Notice it has a `children` attribute that _is_ updating

**5. Expected behavior**

The text should animate, just like in the doc'd demo https://codesandbox.io/p/sandbox/framer-motion-animate-content-bxx7lt?from-embed=

**6. Video or screenshots**

Just a static count, not much to show.

**7. Environment details**

Chrome 123.0.6312.46 / MacOS, reproable everywhere.

simonkarman:
I'm running into the same issue. The code works for motion.h1 elements, however, using the motion value in a motion.text element doesn't work.

I would be happy to create a PR to resolve the issue, however I would need some pointers on where to look.

Xentox-Phil:
yeah same here 🙌🏼

Xentox-Phil:
@simonkarman i have found the same behaviour with using the value that gets returned by the useMotionTemplate and putting it onto a motion.rect it doesn't do anything could that be the same problem as with the texts?

```typescript

const transform = useMotionTemplate`rotate(${shiftedAngleDeg}, ${x}, ${y})`

 <motion.rect
      transform={transform}
      ...
/>
```

i have solved this with the same solution just putting a ref on it and setting it manually...

in the docs theres also only a style example...
https://motion.dev/docs/react-use-motion-template

simonkarman:
I have created the following PR that fixes the issue I was having. Unfortunately no-one has reviewed the PR yet, so I cannot get it merged. You can find it here: https://github.com/motiondivision/motion/pull/2841

You can play around with that PR and see if that fixes your problem too.

Xentox-Phil:
Alright sure ill give it a try when I have time and if the issue persists ill probably open up a issue.

PR #2841, opened 2024-10-23
SVG elements (like motion.text) now update when given a MotionValue as children, matching HTML element behavior

simonkarman:
Fixes the issue as described in: https://github.com/framer/motion/issues/2578

Feature was introduced in: https://github.com/framer/motion/pull/1877

Changelog:
- SVG elements (like `motion.text`) now update when given a `MotionValue` as children, matching HTML element behavior.


simonkarman:
@mattgperry The issue is now fixed. I'm happy to provide additional testing or troubleshooting if needed before merging.

Quick question about #1877: Was the decision intentional to implement `handleChildMotionValue` in `HTMLVisualElement` rather than `DOMVisualElement`?

packages/framer-motion/src/render/dom/DOMVisualElement.ts
```diff
@@ -37,4 +38,21 @@ export abstract class DOMVisualElement<
...
     KeyframeResolver = DOMKeyframesResolver
+
+    childSubscription?: VoidFunction
+    handleChildMotionValue() {
+        if (this.childSubscription) {
+            this.childSubscription()
+            delete this.childSubscription
+        }
+
+        const { children } = this.props
```
simonkarman, line 49: It might still be worth exploring why a `children` prop is set on the SVG elements.

image

Screenshot from example in Chrome macOS: http://localhost:9990/?example=SVG-Text-MotionValue-Child

simonkarman:
Hi @mattgperry, what is the process of getting this PR merged? The pipeline succeeds and there are no conflicts with the main branch. Let me know if you are able to review and approve this. Or let me know if there is someone else I should reach out to. Thanks!

mattgperry:
LGTM - I think SVG attrs/styles needs looking at specifically so I don't mind putting this out with the `children` thing. Thanks for the PR!

PR #2852, opened 2024-10-31
Fix double `.update()` call on mount

mattgperry:
This PR ensures that the insertion effect responsible for updating props from a `motion` commit to its `VisualElement` only runs when the `VisualElement` is hydrated with an element reference. This cuts an unnecessary `.update()` call on mount.


kurtextrem (approved):
nice!

PR #2864, opened 2024-11-12
Changelog doesn't detail breaking changes that are causing downstream issues

krainboltgreene:
So I'm using the livemotion elixir library but we maintain a custom integration with the motion library and we're noticing quite a few deprecations that happened in the v11 update that aren't reflected in the changelog.

I'm still trying to find the replacement for `createMotionState()` that was used.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -709,6 +709,10 @@ Undocumented APIs should be considered internal and may change without warning.
 
 ## [11.0.0] 2024-01-23
 
+## Breaking
+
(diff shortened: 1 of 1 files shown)


mattgperry:
Yeah this is a good point, I can add these to the changelog - however as an
undocumented API this one probably won’t be listed. You can find an upgrade
guide for documented features here:
https://motion.dev/docs/upgrade-guide

There’s no need to upgrade though if you were using createMotionState as
there’s no direct replacement, and the repo remains at
motiondivision/motionone

PR #2876, opened 2024-11-14
fix: export render html utils

### LINKED ISSUE #2875: [FEATURE] export buildHTMLStyles
I'm currently facing a challenge where I need to convert Motion values into HTML-compatible styles for custom integrations. However, the `buildHTMLStyles` function is not accessible externally, which limits my ability to transform or utilize Motion values effectively in specific cases where direct access to HTML-styled outputs is required.

I would like `buildHTMLStyles` to be exported so that it can be accessed as a utility function. This way, I can transform Motion values into HTML styles outside of the Motion framework, enabling more flexible style manipulation and integration with other parts of my application.

One alternative is to manually replicate the logic of `buildHTMLStyles` in my code, but this is inefficient and error-prone, as updates to the original function in Motion would require repeated adjustments. Another alternative is to rely on inline styles, but this doesn't offer the same flexibility as having the function directly available.

**Additional context**  
Exporting `buildHTMLStyles` as a utility could greatly enhance flexibility for developers looking to use Motion-generated values in broader CSS or HTML-based contexts. This feature would support custom animations and styling needs in a more modular way, potentially enabling easier integration across different environments or frameworks.

productdevbook:
close #2875 

This pull request includes updates to the `packages/framer-motion/src/dom.ts` file to improve the export functionality. The most important changes include the addition of new exports from the `render/html/utils` directory.

Export functionality improvements:

* [`packages/framer-motion/src/dom.ts`](diffhunk://#diff-1b2fd64e4dff7dcd1a5ef6d4236645330f2583528bb37614a818dedde028f993R43-R44): Added `buildHTMLStyles` and `buildTransform` to the list of exports.

### DIFF
--- a/packages/framer-motion/src/dom.ts
+++ b/packages/framer-motion/src/dom.ts
@@ -40,6 +40,8 @@ export { pipe } from "./utils/pipe"
 export { progress } from "./utils/progress"
 export { wrap } from "./utils/wrap"
 export * from "./frameloop"
+export {buildHTMLStyles} from "./render/html/utils/build-styles"
+export {buildTransform} from "./render/html/utils/build-transform"
 
 /**
  * Deprecated


mattgperry:
Hey @productdevbook, I'm closing this as this isn't functionality I want to make public. I have plans here around redesigning this whole part of the pipeline so it should be expected to break at any time.

productdevbook:
Could you make this available temporarily until the redesign is complete, so we can use it in the meantime? Once the redesign is ready, could you introduce it to the community in its updated form? Am I mistaken in thinking this functionality is essential for us right now?

mattgperry:
Its redesigned form will be fundamentally different. I'm not sure if it's essential for you or not, it depends on what you're doing. Personally I don't see it necessary to upgrade from `motion@10`, it works well for your usecase.

PR #2892, opened 2024-11-17
exports DynamicAnimationOptions and DOMKeyframesDefinition

CRBroughton:
Hello there :) Saw a recent video on Motion (thanks Theo) and I've started playing around with it and realised for a Vue or any other framework wrapper, we'll probably need access to some of the underlying types that are currently only exposed via the React implementation.

There's also an open issue with this request (minus a type) - [https://github.com/motiondivision/motion/issues/2870](https://github.com/motiondivision/motion/issues/2870)

I've exported both the `DynamicAnimationOptions` and `DOMKeyframesDefinition` types to help enable this.

If there's anything else I need to do on this PR, please shout. I've run the build locally and all seems fine and my very basic implementation with Vue seems to work with the updated files.


targetlucked69:
Hey nice PR! Would love to see `MotionProps` export too! Maybe create an internal one used by react that has `children` prop, but the default one doesnt have

mattgperry:
Yeah I think the problem with the exclude method is it needs to include React to even define `MotionProps` in the first place. I'm currently doing a refactor of DOM things out of Framer Motion so let me take a look at these types and see what I can do.

PR #2906, opened 2024-11-23
Updating three and @react-three/fiber to latest versions

oconpa:
Upgrading both peerDependencies for three and @react-three/fiber to the latest versions, and removing the hardcoded restriction for @react-three/fiber to be fixed to version 8.2.2


mattgperry:
Thanks for this, I'll try and get it published today

mattgperry:
Btw I ended up reverting these changes due to numerous type errors and failing tests.

PR #2937, opened 2024-12-11
Adding missing types to the dom.ts file

alagar-dev:
From what I can tell, the imports added would fix the typescript import errors. I'm not sure if I 100% avoided the react imports, but this is a general start to fixing the missing types bug for non-react projects.

Please make all the edits you need. Just a suggestion, possibly separate the react only types from the others either with spacing.comments or files. It'll be a headache in the future to keep track of everything. 

Personally I'd put all of my types in a folder, with each specific type having it's own folder. At the base of the types folder I'd have an index.ts consuming all the types. Then I'd place a reactTypes folder in there, consume all of the regular types you'd need from types/index. That way you could import all of the regular types and react only types when needed.

### DIFF
--- a/packages/framer-motion/src/dom.ts
+++ b/packages/framer-motion/src/dom.ts
@@ -46,6 +46,14 @@ export { wrap } from "./utils/wrap"
 export * from "./frameloop"
 
 /**
+ * Types
+ */
+
+export * from "./animation/types"
+export * from "./animation/sequence/types"
+export * from "./projection/geometry/types"
+
+/**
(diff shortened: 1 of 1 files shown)


mattgperry:
Thanks for the PR! Though tests are currently failing because of React types in the output vanilla types.

I'm gonna close this PR and take a look at this on an ad-hoc basis starting with `AnimateSequence` in your issue. As it'll be part of a larger project refactoring out more stuff from `"framer-motion"` into `"motion-dom"`

PR #2944, opened 2024-12-13
Fix a potential memory leak if a render step pauses for a long time.

onnlucky:
Motion can hold on to user memory, potentially for a long time, if a render step doesn't have work to do for a long while.

This caused some trouble for us, where through callbacks and closures, this would hold on to old tree data, after a tree was reloaded from the server. Doubling memory use.

image

This PR simply clears the `thisFrame` set the moment it is no longer needed. Fixing the problem for us.


mattgperry:
Wow a wild @onnlucky appears. Great fix, thanks!

PR #2951, opened 2024-12-19
`transition.out`

mattgperry:
This PR adds a new `out` option to `Transition`. When `true`, this transition will be used when the value animates out of its current state.

```jsx
<motion.div
  initial={{ y: 20 }}
  animate={{ y: 0, transition: { delay: 1 } }}
  whileHover={{ y: 2, transition: { out: true } }} // whileHover -> animate = no delay!
/> 
```

Closes https://github.com/motiondivision/motion/issues/1725


mattgperry:
Closing this not because it's not going ahead but just to clean up the board - it's been hanging round a while and we're yet to verify this fixes the specific case we're making it for in Framer

PR #2974, opened 2025-01-07
chore(.nvmrc): add Node.js version 22 for environment consistency

sukvvon:
* Many people may work with different Node.js versions. Therefore, to unify the development environment, Node.js version 22 is specified in the .nvmrc file to ensure consistency.


mattgperry:
Thanks for the PR!

PR #2987, opened 2025-01-09
Updating R3F and Three to latest version; fixing associated types

kgsherman:
Updates dep + peer dep versions.
There's currently usage of `Object3DNode<any, any>`, which was throwing type errors with the new version and was not actually providing any type safety in many cases. Those have been replaced with `Record<string, any>`, though I don't think that's the long term solution.


mattgperry:
Closing this out as I've deprecated Motion 3D as of `12.1.0`

PR #2997, opened 2025-01-13
Add missing types for framer-motion

Adesh-Pandey:
All the members of `framer-motion/src/animation/types.ts` are not available anymore. This makes sure we export it at the top level.


Adesh-Pandey:
I found this issue while trying to upgrade from 10 to 11.17. 
I build it locally to test it and is working like you would expect.

Adesh-Pandey:
Addresses [Issue](https://github.com/motiondivision/motion/issues/2988). 
Btw I have been using nothing but this library for past 3 years. Thank you for maintaining -what it seems from the commit history- it single handedly.

Issue #3055, opened 2025-02-09
[BUG] `onTap` callback fires when button has disabled prop set to `true`


michTheBrandofficial:
Using the onTap callback on `motion.button` elements works fine all the time. But it shouldn't fire when the disabled prop of the element is set to true

Code reproduction on Stackblitz : [Motion app on stack blitz](https://stackblitz.com/edit/react-rjvyjsuk?file=src%2FApp.js)

**Steps to reproduce:**

1. Go to linked reproduction
2. Open the console to see the logs.
3. Click on the will not fire button - the onClick handler should not fire (expected when the button is disabled)
4. Click on the `will fire` button - the onTap handler fires which is not expected

**Expected behavior**

The onTap callback handler should not fire when a button is disabled. It should act like a onClick handler.

**Environment details**

OS: Windows 11 Pro Build 22000.194
Browser: Tested on Firefox and Zen browser
Browser version: Zen v 1.7.6b

gs18004:
Thank you for your report. I will work on this issue.

michTheBrandofficial:
Thank you @gs18004 for your commit. Will close this now

PR #3056, opened 2025-02-09
fix: fix to ignore press event when button is disabled

gs18004:
#3055

Now, onTap callback does not fire when button has disabled prop set to true.


gs18004:
Test fails, but I think this is not the problem with my code. The problem in the test result is not related to my commit.

michTheBrandofficial (commented):
This works fine man. Thanks for the fix

mattgperry:
Banger PR, thanks!

PR #3075, opened 2025-02-20
fix #2822 - Get transform values from DOM as animation starting point

### LINKED ISSUE #2822: [BUG]  animate() doesn't transition from the initial transform state of element

**2. Describe the bug**

When using the animate() function, it appears to ignore the initial transform state of the element.

https://codesandbox.io/p/sandbox/33k8r2

**4. Steps to reproduce**

1. Create a DOM element with an initial transform of `scale(0.1)`
2. Use `animate(dom, {scale: 1.5})` to attempt animation

**5. Expected behavior**

The element should transition from scale(0.1) to scale(1.5)

**6. Video or screenshots**

**7. Environment details**

chrome: 129.0.6668.89

JeroenReumkens:
Fix #2822 

As discussed on Discord, I ran into a similar issue last week. I took some time to dive into the codebase, and might have found a fix for this. I'm not sure if this is the desired behavior though... So open to discuss approaches. 

It adds a util function that gets the transform matrix via computed style, to extract all transform properties from it. It uses this for starting points of animations where the transform property value is `null`. In the previous implementation we  were extracting all styles from the DOM, except for transform styles. This function adds the logic, while keeping the logic for the default values too. 

Also added an example file that demonstrates the behavior. Before this change the scale would start at `1` instead of `0.1`. 

Happy to discuss further. You know where to find me.

### DIFF
--- a/dev/react/src/examples/Animation-useAnimate-initial-transform.tsx
+++ b/dev/react/src/examples/Animation-useAnimate-initial-transform.tsx
@@ -1,0 +1,34 @@
+import { useAnimate } from "framer-motion"
+
+export const App = () => {
+    const [scope, animate] = useAnimate()
+
+    return (
+        <div className="App" ref={scope}>
+            <div
+                className="four"
+                style={{
+                    width: 50,
+                    height: 50,
+                    opacity: 0.5,
+                    backgroundColor: "hotpink",
+                    transform: "scale(0.1)",
+                }}
+            ></div>
+            <p>Take in original transform</p>
+            <button
+                onClick={() => {
+                    animate([
+                        [
+                            ".four",
+                            { x: 90, scale: 2, opacity: 1 },
+                            { duration: 2 },
+                        ],
+                    ])
+                }}
+            >
+                play
+            </button>
+        </div>
+    )
+}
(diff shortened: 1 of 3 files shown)


mattgperry:
Hey @JeroenReumkens 

Thanks for the PR! I'm a little concerned about the filesize. But after I get to the `AnimatePresence`x`layout` bug I'll take a look, add some tests etc.

JeroenReumkens:
> Hey @JeroenReumkens
> 
> Thanks for the PR! I'm a little concerned about the filesize. But after I get to the `AnimatePresence`x`layout` bug I'll take a look, add some tests etc.

Appreciate it @mattgperry 

Tried to look at the failing test, but I can't consistently reproduce it locally. Sometimes a different test fails. I have a feeling it is due to the settimeout's being used, which aren't long enough for the animation to complete, failing the test. Wondering whether my change added that much complexity that it slowed calculations down so much. 

Also curious about your file size comment. Looking forward to the approach you will take, as it seems this logic wasn't in the codebase pretty much since day 0. 

Let me know if I can be of help anywhere.

mattgperry:
@JeroenReumkens I think this is just a flaky test rather than something to do with this PR!

mattgperry:
Merging this into `fix/read-transform` and will have a play

mattgperry:
This is the branch PR (WIP): https://github.com/motiondivision/motion/pull/3077

PR #3149, opened 2025-04-11
Remove random from PresenceChild

gnoff:
Next.js is experimenting with a mode called dynamicIO that interprets IO which introduces non-determinism as un-prerenderable. Generally this is a fine thing to do but there are sync sources of IO like reading random sources and system clocks which are ambiguous. If you use the output of these reads in the realm of introspection and logging there really isn't an imperative need to interpret them as dynamic. But if they affect the rendered output then they clearly communicate their inherent dynamism (i.e I want to see a random order of items every time I load this page).

During a test migration to dynamicIO we discovered that PresenceChild uses Math.random() to achieve an object identity trick guarantee some rendering behavior in the browser. However since all browser code in React (except effects) run in SSR too this trips up the dynamicIO heuristic and Next.js interprets this PresenceChild-using-component as dynamic when it doesn't need to be.

This change updates the implementation of PresenceChild to achieve the desired goal without reading from a random source.


packages/framer-motion/src/components/AnimatePresence/PresenceChild.tsx
```diff
@@ -34,40 +34,38 @@ export const PresenceChild = ({
...
-            onExitComplete: memoizedOnExitComplete,
+            onExitComplete: (childId: string) => {
+                presenceChildren.set(childId, true)
+
+                for (const isComplete of presenceChildren.values()) {
+                    if (!isComplete) return // can stop searching when any is incomplete
+                }
+
+                onExitComplete && onExitComplete()
+            },
```
gnoff, line 53: I reintegrated this callback to eliminate a hook now that the functions in the context object only get recreated when their dependencies change even when presenceAffectsLayout is true.

Additionally, the register function was not previously afforded this memoization behavior and now is which is a nice plus.

And we are using fewer hooks

packages/framer-motion/src/components/AnimatePresence/PresenceChild.tsx
```diff
@@ -34,40 +34,38 @@ export const PresenceChild = ({
...
             },
-        }),
-        /**
-         * If the presence of a child affects the layout of the components around it,
-         * we want to make a new context value to ensure they get re-rendered
-         * so they can detect that layout change.
-         */
-        presenceAffectsLayout
-            ? [Math.random(), memoizedOnExitComplete]
-            : [isPresent, memoizedOnExitComplete]
```
gnoff, line 69: another alternative to random would be to pass in `Object.create(null)` or `{}`. This though just further adds to GC pressure so I went with the conditional clone instead

mattgperry:
Awesome PR, thanks! I'll try and get it merged and released this week.

PR #3152, opened 2025-04-14
fix: clean up isPressing state after parent is dragged

### LINKED ISSUE #2953: [BUG] `onTap` callback doesn't fire on child component after parent is dragged

**2. Describe the bug**

When a tappable component that is a child of a draggable component is used to initiate a drag gesture, its `onTap` callback no longer fires after the drag gesture ends. I suspect this is because the event is being filtered out by `isValidPressEvent` before `isPressing.delete(element)` is called [here](https://github.com/motiondivision/motion/blob/main/packages/motion-dom/src/gestures/press/index.ts#L70).

https://codesandbox.io/p/sandbox/88xds7

**4. Steps to reproduce**

1. Open CodeSandbox
2. Use blue "Child" div as handle to drag parent div
3. Tap "Child" div after drag gesture ends
4. Observe that "Tapped" message is not logged

**5. Expected behavior**

`onTap` handler should fire when "Child" component is clicked, even if "Child" was previously used to initiate a drag gesture on the parent element.

**6. Video or screenshots**

https://github.com/user-attachments/assets/e5bd6461-dbd6-487b-a5b0-dea02b417ad2

**7. Environment details**

N/A

bporcelli:
Remove element from `isPressing` in `onPointerEnd` always, even if `isValidPressEvent(endEvent)` evaluates to `false`.  

Fixes #2953


mattgperry:
Thanks for the PR! Will publish today

Issue #2949, opened 2024-12-18
[BUG] SVG styles not applying on mount


mattgperry:
Visit https://examples.motion.dev/react/use-transform

The offset isn't applied correctly until the first render.

PR #3154, opened 2025-04-15
Fix the SVG transform-origin during initial mount

Taeyeon-Lim:
Dimension is measured on the client side. Therefore, the origin cannot be measured on the first mount, resulting in a jump. Before the dimension is measured, set transformBox to "fill-box" and center the origin to leave the initial origin to the browser.

This is only valid when manually calculating the last value of transformOrigin.

ref: #2949


mattgperry:
I'm a little concerned about the extra bundle size here, but I have to ask, if this works, is it possible for us to just ditch measuring the SVG elements entirely and use this `fill-box` instead?

Taeyeon-Lim:
Unfortunately, I don't know exactly what context "calcSVGTransformOrigin" was added in, so it's hard to give you a definite answer. "fill-box" is widely used, but I think I can keep "calcSVGTransformOrigin" if it requires a manual accurate calculation.

As a temporary measure, I took the form of leaving the server to the browser. I need to make sure most browsers guarantee the same behavior of the "fill-box" and "transform-origin: 50% 50%" combination. What I've checked are firefox and chrome browsers. 

For now, I think it's possible. However, I think someone might want to use "view-box". I think we might consider whether to give this option to the user or not.

mattgperry:
Ah ok - well currently as it stands the intended behaviour of all this measurement is to force transforms to apply to SVG elements the same way they do HTML elements. There's no option to have the default behaviour. So by setting `originX: 0.1` we want an origin point thats 10% into the width of the SVG element itself, rather than the parent view box. In other words, if this behaviour is attainable by setting `transform-box: fill-box` we could replace all that logic with just that one style in `useSVGProps`

mattgperry:
Having a quick play, this is what we're trying to do with all the measurements. This would be a big win for performance and filesize.

Taeyeon-Lim:
I'll try it.

Taeyeon-Lim:
Hi, I have a question. The origin doesn't apply to HTML elements (not SVG). Wouldn't it be better if origin could apply to HTML elements as well?...or is it not necessary?

### example
```js 
// HTML element origin
<motion.div style={{ originX : 0.1 }}  /> 
```
- current : `"50% 50%"`
- better(?) : `"10% 50%"`

Taeyeon-Lim:
oh..., I understand. Thank you.

packages/framer-motion/src/components/Reorder/__tests__/server.ssr.test.tsx
```diff
@@ -13,7 +13,7 @@ describe("Reorder", () => {
         const staticMarkup = renderToStaticMarkup(<Component />)
         const string = renderToString(<Component />)
 
-        const expectedMarkup = `<article><main draggable="false" style="z-index:unset;transform:none;-webkit-touch-callout:none;-webkit-user-select:none;user-select:none;touch-action:pan-x"></main></article>`
```
mattgperry, line 16: This should only be the case if `originX` etc or `transformOrigin` have been explicitly set.

packages/framer-motion/src/motion/__tests__/ssr.test.tsx
```diff
@@ -146,7 +146,7 @@ function runTests(render: (components: any) => string) {
         )
 
         expect(div).toBe(
-            '<div style="transform:translateX(100px) translateY(200px)"></div>'
```
mattgperry, line 149: Likewise we definitely don't want to set this if it wasn't explicitly set by the user - unless this is an SVG with a `transform`

packages/framer-motion/src/motion/__tests__/ssr.test.tsx
```diff
@@ -204,7 +204,7 @@ function runTests(render: (components: any) => string) {
         )
 
         expect(rect).toBe(
-            '<rect mask="" class="test" style="background:#fff"></rect>'
```
mattgperry, line 207: How come this one doesn't have `transform-box:fill-box`?

packages/framer-motion/src/render/html/utils/build-styles.ts
```diff
@@ -69,12 +69,12 @@ export function buildHTMLStyles(
      * Build a transformOrigin style. Uses the same defaults as the browser for
      * undefined origins.
      */
-    if (hasTransformOrigin) {
+    if (hasTransformOrigin || style.transform) {
```
mattgperry, line 72: If you move this logic to the `svg/utils/build-styles` file this will remove the style from HTML elements

mattgperry:
I left some comments, but this looks amazing. I think if we can get `transform-origin` from appearing in HTML elements then we're set.

Taeyeon-Lim:
I'm sorry for the late response and thank you for the comment. Well, I got the default behavior wrong. What you said meant it only changed when you were given `transform-origin` or `transform` by the user...
I will remove the `style.transform` of `html/utils/build-styles` and change the `transform-origin` of SVG to operate when there is only `transform-origin` from html. If I do this, unlike HTML, if there is only `transform`, `transform-origin` will act like `0, 0` instead of `50% 50%`, right? Or, should I include this case as well?

mattgperry:
No rush! Yes good to default to 50% for SVG with transforms but to ensure
HTML only gets an origin if one is defined

Taeyeon-Lim:
I've understood your intentions clearly, and I'll take my time to ensure it's done well. don't worry

mattgperry (approved):

mattgperry:
Thanks for your work on this @Taeyeon-Lim! There's just the conflicts to resolve and we can publish this - I'd do it myself but I'm getting a message that they're too complex to resolve in the web editor.

Taeyeon-Lim:
I see, I'll try to solve it

Taeyeon-Lim:
The conflict is resolved, but the content has gotten a little messy. Is it better to do a new PR?

mattgperry:
@Taeyeon-Lim No it's perfect, thanks for all your work on this!

PR #3181, opened 2025-05-02
Fix React 19 `StrictMode`

mattgperry:
Fixes https://github.com/motiondivision/motion/issues/3180


kevva:
Seems like `12.9.5` failed to publish.

PR #3196, opened 2025-05-07
fix: Layout animation jumps when interrupted by another layout animation

### LINKED ISSUE #3195: [BUG] Layout animation jumps when interrupted by another layout animation
**2. Describe the bug**
When a layout animation is in progress and another layout animation is triggered, the element jumps to an incorrect position before starting the new animation.

This issue affects official examples like number-formatting and radix-tabs, breaking their smooth transition animations.

**6. Video or screenshots**

https://github.com/user-attachments/assets/b3dcc216-02a0-4742-b699-804becf74d8c

https://github.com/user-attachments/assets/9b610610-7501-445a-9765-27a676a1d129

rick-hup:
fix #3195 
- When an animation is interrupted, stop() immediately executes a tick(),forcing the calculation of the current frame's state.
-  This state calculation is based on the current time point, not on the continuity of the animation.  
-  When a new animation starts, it begins from this forcibly calculated state rather than from the smooth transition point of the previous animation.

### DIFF
--- a/packages/motion-dom/src/animation/JSAnimation.ts
+++ b/packages/motion-dom/src/animation/JSAnimation.ts
@@ -435,7 +435,7 @@ export class JSAnimation<T extends number | string>
     stop = () => {
         const { motionValue } = this.options
         if (motionValue && motionValue.updatedAt !== time.now()) {
-            this.tick(time.now())
+            this.tick(this.holdTime!,true)
         }
 
         this.isStopped = true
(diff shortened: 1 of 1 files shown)


mattgperry:
There's some local tests you can run surrounding this - try `yarn test-playwright`

mattgperry:
Ah the test doesn't capture this fully so I'll update but if you run `yarn dev` and visit http://localhost:8000/playwright/animate/animate-stop.html you can see the problem this is attempting to solve (syncing the position of the bottom two boxes)

mattgperry:
Just pointed this branch to a branch that contains a test for syncing `.stop()`

mattgperry:
I've made it so the layout animations don't perform this syncing behaviour. Which isn't ideal but does the trick for now. I didn't have any luck capturing this behaviour in an isolated test though

rick-hup:
After examining the animate-stop demo, when JSAnimation is stopped, the element immediately jumps to its final animation state.

PR #3223, opened 2025-05-18
fix: layout animation interrupt jump

rick-hup:
- Removed the `sync` parameter from the `stop` method in `JSAnimation` for cleaner usage.
- Moved `setAnimationOrigin` call to after starting the animation in `createProjectionNode` to prevent layout jumps caused by stopping previous animations.

### DIFF
--- a/packages/framer-motion/src/motion/__tests__/component.test.tsx
+++ b/packages/framer-motion/src/motion/__tests__/component.test.tsx
@@ -308,4 +308,40 @@ describe("motion component rendering and styles", () => {
         const { container } = render(<Test />)
         expect(container.firstChild).toBeTruthy()
     })
+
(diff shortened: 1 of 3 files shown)


rick-hup:
https://github.com/user-attachments/assets/de7b3e51-355e-4c1d-9971-c89bf2d2ebca

rick-hup:
I previously misunderstood the issue. Here's what actually caused the layout animation jump:
- setAnimationOrigin initializes a new first frame for the layout animation
- In startAnimation, stopping the previous layout animation triggers an update callback based on the previous animation, causing the layout to jump
- Solution: Moved setAnimationOrigin to run after startAnimation to prevent this issue"

mattgperry (approved):

PR #3234, opened 2025-05-27
fix: remove deprecated add listener

tenthyoung:
# fix: remove deprecated add listener

https://github.com/motiondivision/motion/issues/2823


mattgperry:
Thanks for the PR!

PR #3245, opened 2025-06-03
perf(renderHTML): Use `for..in` over `Object.assign`

kurtextrem:
The microbenchmark seems to hint to better performance over `Object.assign`: https://esbench.com/bench/683f2196545f8900a4de2d41


mattgperry:
Thanks for the PR! I’ll take a look tomorrow. hopefully this loop is going away soon in favour of https://motion.dev/docs/style-effect which is much more atomic in its updates, but this would be nice in the meantime

kurtextrem:
Related: I assume we could make the getProjectionStyles call more efficient by creating a function that works similar, but instead of returning the styles, it would apply on element.styles directly. That way we could skip some object allocations etc.

PR #3247, opened 2025-06-04
perf: Make getProjectionStyles write instead of return fresh obj

kurtextrem:
Stack on https://github.com/motiondivision/motion/pull/3245.

Idea here: https://github.com/motiondivision/motion/pull/3245#issuecomment-2936468341.
This way we save an object allocation, plus can skip the `for..in` loop, as we make `getProjectionStyles` write to the target node styles directly.

The types are a bit hacky here, but that was the least invasive fix I could think of (without duplicating all the logic).


mattgperry:
Merging into a branch to take a look at these tests

PR #3250, opened 2025-06-08
Add support for defining a timeline position relative to the start of the most recently added animation (e.g. `<0.5`, `<-1`)

piotrjoniec:
This is something I'm really missing ever since switching from GSAP.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ Motion adheres to [Semantic Versioning](http://semver.org/).
 
 Undocumented APIs should be considered internal and may change without warning.
 
+## [12.17.0] 2025-06-08
+
+### Added
+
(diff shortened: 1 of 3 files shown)


mattgperry:
Thanks for the PR! I’m away from a laptop atm so I can’t double check this
but I could swear it already works like this via the at option?

piotrjoniec:
These are the existing options:

```ts
    if (typeof next === "number") {
        return next
    } else if (next.startsWith("-") || next.startsWith("+")) {
        return Math.max(0, current + parseFloat(next))
    } else if (next === "<") {
        return prev
    } else {
        return labels.get(next) ?? current
    }
```

So you can only define time relative to the end of the previous animation - but not the start.

mattgperry:
Ohhh I see. That's great, nice addition!

PR #3252, opened 2025-06-09
refactor: extract children motion-value logic from visual-element

rick-hup:
- Moved logic related to handling children motion-value out of visual-element, as children is a React-specific concept
- Keeps visual-element framework-agnostic and cleaner
- Children motion-value handling will now be managed in React-specific layer


mattgperry:
It's funny as although children is React-specific, `.textContent` is still DOM-specific. Makes me wonder if this belongs in `VisualElement` some way. 

Regardess, I'm trying to reduce the number of effects overall so I don't love adding a new one for something that will almost never be used. (I suppose the same applies to having this feature in there at all)

To think about it from the Framer/Figma codegen perspective, it's an extra effect being run in a component that can sometimes be included hundreds of times.

Perhaps it's possible to take this hook and put it in an optional `MotionValueContent` child within the `motion` component?

```javascript
isMotionValue(children) ? <MotionValueContent ref={ref}>{children}</MotionValueContent> : children
```

Nice thing about this PR is I recently published the `MotionNodeOptions` type in `"motion-dom"` which is `MotionProps` without `style` and `children`. This PR would get use closer to moving `VisualElement` from `MotionProps` to `MotionNodeOptions`

rick-hup:
I think the MotionValueContent component is a good idea. Right now, children can only render a single Motion value. 
with the MotionValueContent component, users can have more flexibility in how MotionValue is rendered through children.

Motion Vue already provides a similar component for rendering Motion values. Although, in that case, the main reason for adding it was because Vue cannot pass objects value to slots..

PR #3261, opened 2025-06-13
fix: #3260 keep sub-pixel `width/height` when snapshotting exit layout to eliminate 1-px shift

### LINKED ISSUE #3260: [BUG] Layout shift caused by popLayout setting width without sub-pixel precision
**2. Describe the bug**

With `AnimatePresence` set to `mode="popLayout"`, the exiting element’s geometry is captured and reapplied via inline styles.

If the element’s computed width or height contains sub-pixel values (e.g. `402.4 px`), those dimensions are **rounded to the nearest integer** before the element is promoted to `position:absolute`.

The 1-pixel discrepancy introduces an immediate layout shift that is visible just before the exit transition starts.

https://codesandbox.io/p/sandbox/yz3wqg

**4. Steps to reproduce**

1. Open the sandbox. The red-bordered container is rendered at `402.4 px` wide.  
2. Press the space bar (or click the button) to trigger the exit animation.  
3. Watch the container jump by ±1 px in the first frame of the exit.

**5. Expected behavior**

`AnimatePresence` should preserve the exact sub-pixel dimensions it reads from `getBoundingClientRect()` when it applies `width`/`height` inline during `popLayout`, eliminating the layout shift.

**6. Video or screenshots**

Issue in real-world practice:

Note that the subpixel precision loss caused by the activation of AnimationPresence has led to text wrapping layout shift.

https://github.com/user-attachments/assets/7ad77c2b-1a0a-49d8-b70e-5e0a7cde1d98

Issue in provided sandbox:

https://github.com/user-attachments/assets/0786372a-eb1a-48bb-83b2-7cd43b9ef7aa

**7. Environment details**

```
  System:
    OS: macOS 26.0
    CPU: (12) arm64 Apple M3 Pro
    Memory: 239.89 MB / 36.00 GB
    Shell: 5.9 - /bin/zsh
  Browsers:
    Chrome: 137.0.7151.104
    Chrome Canary: 139.0.7235.0
    Safari: 19.0
    Safari Technology Preview: 19.0
```

reekystive:
This PR fixes #3260.

This is just a preliminary modification and has not undergone very thorough testing. Any suggestions and changes are very welcome.

## The Issue

When `AnimatePresence` runs in `mode="popLayout"`, it captures the exiting node’s geometry inside  
`PopChildMeasure.getSnapshotBeforeUpdate`.

The current implementation relies on `offsetWidth/offsetHeight`, which are always rounded to integers, so an element that is **400.4 px** wide becomes **400 px** the instant it is promoted to `position:absolute`.  That 1-px delta produces a visible layout shift on the very first frame of the exit animation.

This PR replaces those integer measurements with `getBoundingClientRect()` so that the stored `width`/`height` retain full floating-point precision. The change is strictly internal: public API and type signatures remain unchanged.

### DIFF
--- a/packages/framer-motion/src/components/AnimatePresence/PopChild.tsx
+++ b/packages/framer-motion/src/components/AnimatePresence/PopChild.tsx
@@ -34,16 +34,17 @@ class PopChildMeasure extends React.Component<MeasureProps> {
         const element = this.props.childRef.current
         if (element && prevProps.isPresent && !this.props.isPresent) {
             const parent = element.offsetParent
-            const parentWidth = isHTMLElement(parent)
-                ? parent.offsetWidth || 0
-                : 0
+            const parentRect = isHTMLElement(parent)
+                ? parent.getBoundingClientRect()
+                : { width: 0 }
 
+            const rect = element.getBoundingClientRect()
             const size = this.props.sizeRef.current!
-            size.height = element.offsetHeight || 0
-            size.width = element.offsetWidth || 0
+            size.height = rect.height
+            size.width = rect.width
             size.top = element.offsetTop
             size.left = element.offsetLeft
-            size.right = parentWidth - size.width - size.left
+            size.right = parentRect.width - size.width - size.left
         }
 
         return null


mattgperry:
The problem with this approach is `getBoundingClientRect` is vulnerable to `transform` on the element itself or on its ancestors.

PR #3270, opened 2025-06-23
Fixing animation types

mattgperry:
Restores support for `translateX`, `originX` etc to `Transition` and `Target` types.

Also adds `Spring` and `Tween` types as deprecated alias for `SpringOptions` and `KeyframeOptions` respectively

Fixes https://github.com/motiondivision/motion/issues/3264


JacobArrow:
@mattgperry does this also fix [these Typescript errors](https://github.com/motiondivision/motion/issues/3264#issuecomment-2991004431)?

mattgperry:
@JacobArrow I don't think that the linked is incorrect as none of those values are valid `Transition` values

PR #3272, opened 2025-06-23
Don't suspend layout animations on iOS when scrolling

raunofreiberg:
This PR resolves https://github.com/motiondivision/motion/issues/3267.

Namely, as I understood from our chat with @mattgperry layout animations suspend on window `resize` events. Unfortunately iOS triggers a `resize` event when the address bar expands or collapses based on scroll position. 

The fix could be to compare window width between events and suspend only if there is an actual change to the window width.

> Note: comparing height will not yield the result we want; height actually changes as a side effect of the address bar movement


mattgperry:
Thanks for the PR! Super busy week so I'll take a look next week, lgtm at a glance.

PR #3280, opened 2025-06-30
feat: drag & pan session configurable distance threshold

KingSora:
Fixes: https://github.com/motiondivision/motion/issues/3279

Adding an option called `distanceThreshold` to the `DragControlOptions` interface to make it configurable.

If this PR is generally accepted I can also add the missing tests & Changelog infos.


mattgperry:
Thanks for the PR!

PR #3281, opened 2025-06-30
feat: add cancel() and stop() functions to useDragControls

KingSora:
Fixes: https://github.com/motiondivision/motion/issues/2721

This exposes a `cancel()` and `stop()` function in the `DragControls` class.
Those functions can be used to cancel or stop drag animations at any point in time.

If this PR is being accepted I can add missing tests and docs & Changelog infos.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,40 @@
+## Structure
+
+This is a monorepo. Development and test environments live in `/dev`, packages live in `/packages`.
+
+Packages are:
+
+-   `framer-motion`: For React-specific code. More of this should be refactored to `motion-dom` where possible.
+-   `motion`: Re-export of `framer-motion`.
+-   `motion-dom`: Vanilla/JS animation library.
+-   `motion-utils`: Pure functions and easing functions.
+
+## Tests
+
+There are three types of test suites in Motion.
+
+1. Jest (unit) tests
+2. Cypress e2e tests
+3. Playwright e2e tests
+
+### Jest
+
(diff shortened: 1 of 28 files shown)


mattgperry:
Hey @KingSora thanks for the PR. 

Is there ever a case when you'd want to manually pass event/info to cancel/stop?

KingSora:
Hey @mattgperry :)

Thanks for taking a look! For my personal usecase both params aren't really needed, but I thought it might be a good addition in case I'm missing something.

From my side this feature would also completely work without them.

mattgperry:
Ok what I'm going to do is merge it into a branch and have a play with this - thanks again!

mattgperry:
I reverted the changes to the logic to `cancel` and `stop` mainly because they seem apart from the scope of simply adding `cancel()` and `stop()` methods, I'm a little concerned no tests flagged the behavioural change.

KingSora:
@mattgperry thanks for merging the feature :)

I've noticed that you removed the `onDragEnd()` call from the `cancel` function in your PR, this will have the side effect that the `onDragEnd` handler will not be called when the `cancel()` function was called. (Could be intentional though)

mattgperry:
Yeah well this was the existing behaviour so I'm hesitant to change it. I think a better API could be a success flag passed via `onDragEnd` but would need to see how this is being used in Framer etc before changing anything.

KingSora:
@mattgperry I see.. before the changes the `cancel()` function was only called in the `stop()` function and it was private, so I figured it shouldn't be a breaking change to move the `onDragEnd` callback into it.. what might have changed as a side effect is the order (e.g. animation starts before the callback vs after the callback)

PR #3318, opened 2025-07-17
fix: eliminate redundant removeChild

rortan134:
Eliminate redundant removeChild call for the injected style element, seems to have slipped in on a previous refactor


mattgperry:
Thanks @rortan134 !

PR #3319, opened 2025-07-18
Fix: Drag offset persists after drop in the same row

JonAbrams:
Supposedly fixes https://github.com/motiondivision/motion/issues/3315, I still need to properly test it.


mattgperry:
Thanks for the PR - I'm just closing this out for now pending a test. Or alternatively if I get to this issue soonish I can take a shot too.

PR #3327, opened 2025-07-22
Fix: move "use client" directive from barrel files to client components

### LINKED ISSUE #3211: [BUG] Latest version incompatible with NextJS 15 build?
Hello everyone and thanks for such a nice library! :) 

Anyway, my NextJS build is failing due to an error when using latest version `12.11.1`:

```js
./node_modules/framer-motion/dist/es/index.mjs
Error: It's currently unsupported to use "export *" in a client boundary. Please use named exports instead.
    at Object.transformSource (.../node_modules/next/dist/build/webpack/loaders/next-flight-loader/index.js:104:53)

Import trace for requested module:
./node_modules/framer-motion/dist/es/index.mjs
```

Any clue what could cause this?

My versions:

```json
"next": "15.3.2",
"react": "19.1.0",
"react-dom": "19.1.0",
```

Motion is at @12.11.1

Thanks

healtheloper:
Fix #3211

## Problem
Placing the `"use client"` directive in barrel files effectively turns the entire module into a client boundary. This causes a Next.js build error because the `export * from` is not supported within files marked as client boundaries.

## Solution
- Removed the `"use client"` directive from all barrel files.
- Added the `"use client"` directive directly to the top of individual client components and hooks where client-side APIs (like `useState`, `useEffect`) are used.

## Testing
- Next env build: Confirmed that the `export *` build error is resolved by running next build.
- RSC Compatibility Test: To ensure compatibility with React Server Components, components were imported directly from motion/react-client and validated to function correctly in an RSC environment.
- Client Component Test: For standard motion components (e.g., <m.div>), verified they render and animate correctly after their parent component was explicitly marked with the 'use client' directive.

## Potential Impact
Components that were previously working due to being within the client boundary may now require explicit `"use client"` directives. Additional files may need to be updated if they were missed during this change.

Thanks for such a nice library!


mattgperry:
Thanks for the PR! I’ll check it out tomorrow. I think you’ll make a lot of people very happy 😅

mattgperry:
@healtheloper The existing use-client test is failing on this branch. You should be able to run it locally with `make test-nextjs`. This test just runs the Next environment in `dev/next` and checks Motion components are rendered:

```
describe("use client", () => {
    it("Correctly renders components", () => {
        cy.visit("/")
            .wait(100)
            .get("#test")
            .should("exist")
            .get("#m-test")
            .should("exist")
    })
})

```

Edit: Oh I see this is because those elements have been removed.

dev/next/app/motion.tsx
```diff
@@ -0,0 +1,17 @@
+"use client"
```
mattgperry, line 1: This should work without `"use client"` if we're importing from `motion/react-client`.

Otherwise you could just do `import { motion } from "motion/react"` like normal

mattgperry: @healtheloper I think you're right, I'm happy for you to remove `motion.create` from the client export. Can I just double check then that you're happy with this PR as it is, other than that?

healtheloper: @mattgperry Yes, thanks for confirming. I'll go ahead and remove `motion.create` from the client export.

healtheloper:
@mattgperry done :)

chris-erickson:
What's the status on this?

Have we figured out what is going on? No one involved, motion, sanity, next wants to touch this in every issue you can surface online, but it's clearly a weird and insidious problem. I am seeing it in a Server Component in a NextJS app. I'm not sure how to keep making server side queries to Sanity but keep the compiler happy. If this PR fixes it.. can we ship this?

mattgperry (approved):

mattgperry:
I've merged this into https://github.com/motiondivision/motion/pull/3353 to try and get an alpha published though I'm getting 

```
next-env:build: node:internal/modules/cjs/loader:1368
next-env:build:   throw err;
next-env:build:   ^
next-env:build: 
next-env:build: Error: Cannot find module '/home/circleci/repo/dev/next/node_modules/next/dist/bin/next'
next-env:build:     at Function._resolveFilename (node:internal/modules/cjs/loader:1365:15)
next-env:build:     at defaultResolveImpl (node:internal/modules/cjs/loader:1021:19)
next-env:build:     at resolveForCJSWithHooks (node:internal/modules/cjs/loader:1026:22)
next-env:build:     at Function._load (node:internal/modules/cjs/loader:1175:37)
next-env:build:     at TracingChannel.traceSync (node:diagnostics_channel:322:14)
next-env:build:     at wrapModuleLoad (node:internal/modules/cjs/loader:235:24)
next-env:build:     at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:171:5)
next-env:build:     at node:internal/main/run_main_module:36:49 {
next-env:build:   code: 'MODULE_NOT_FOUND',
next-env:build:   requireStack: []
next-env:build: }
next-env:build: 
next-env:build: Node.js v22.18.0
next-env:build: ERROR: command finished with error: command (/home/circleci/repo/dev/next) /tmp/xfs-05128fc8/yarn run build exited (1)
```

On that branch. I assume that's because that branch also bumps the version number, which is breaking a CircleCI cache in some way

PR #3338, opened 2025-07-29
Exporting APIs for view transitions

mattgperry:


Fefedu973:
Hey !
Is this the first step into a better support of ViewTransition ?
Is it usable right now ?
Thanks

image

mattgperry:
Yeah I’m prodding around but I’m not really happy with the results at the moment.

Fefedu973:
i'm looking everywhere but can't manage to do it,
Can we use motion with layouts to do cross page animation where an elemnt morph from the old page to the new one and animating independently other elements (like the ios app store example but cross-page instead of in a single page app) and we get to control how each element transition with motion animations etc (like a combinaison of view transition and motion). Is it smth possible to do rn or do your pull request will allow things like that (like the page effect you from framer motion you showed with cross-animation (this means animations where we can at the same time rendered the old and the new page with the element trnasitioning between then instead of transition out from one page to transtiion in to the other with completely separated sequances and that happens sequancially instead of in parallel (example: cross fade animation, morph of object between page transition, examples shown in framer motion)
Thanks !

PR #3344, opened 2025-08-11
Fix: Compose child refs in `PopChild` to preserve external refs

### LINKED ISSUE #2604: ref prop does not work on div directly inside AnimatePresence with mode="popLayout" [BUG]

**2. Describe the bug**
ref prop does not work on the div directly inside AnimatePresence with mode="popLayout"
The ref always comes back undefined.

rortan134:
Fixes #2604 

Hook is from https://github.com/radix-ui/primitives/blob/main/packages/react/compose-refs/src/compose-refs.tsx
Doesn't seem like we have any existing utility for merging refs in the codebase. Let me know if you'd like to do this some other way.


mattgperry:
Thanks for the PR! Will release today

PR #3349, opened 2025-08-18
fix: Support oklch CSS colors

sebastianwd:
Fixes https://github.com/motiondivision/motion/issues/3058

oklch() is a new way to define CSS colors. In oklch(L C H) or oklch(L C H / a), each item corresponds as follows:

    L is perceived lightness (0-1). “Perceived” means that it has consistent lightness for our eyes, unlike L in hsl().
    C is chroma, from gray to the most saturated color.
    H is the hue angle (0-360).
    a is opacity (0-1 or 0-100%).

It's the default color format for the TailwindCSS v4 and shadcn themes.

Added a oklch to rgba converter based on the implementation in the [culori](https://github.com/Evercoder/culori/blob/493ef9f4f4e4cd39a5fcd335d49440d3b7ae59f9/src/oklch/definition.js) library


sebastianwd:
Tests pass locally, not sure what's wrong with ci.
image

mattgperry:
Hey @sebastianwd 

Thanks for your work on this issue though I have to close it unmerged. The primary reason is that this adds a good chunk of bundlesize for a feature that I'm not convinced most people will use.

Earlier in the year I refactored the way keyframes were resolved specifically to add this functionality. If we detect that oklch colors are being animated then we can pass the animation to `NativeAnimation` and let the browser interpolate it. The nice thing about this approach is it won't add any extra bundlesize and it'll expand infinitely - for instance adding support for other color formats including named colors.

This doesn't help you out in an immediate sense though. For that, I did also introduce a short-term fix for Framer that you could also use and that's the ability to add a global value mixer. This would enhance all animations, plus the `useTransform` function.

You can use it essentially like this:

```js
import { MotionGlobalConfig, mix } from "motion/react"

MotionGlobalConfig.mix = (a, b) => {
  if (isOklch(a) || isOklch(b)) {
    return mixOklch(a, b)
  }

  return mix(a, b)
}
```

PR #3363, opened 2025-09-10
Fix incorrect type of Reorder.Item component

### LINKED ISSUE #3359: [BUG] Incorrect type of Reorder.Item component
**1. Describe the bug**

Typescript no longer catches type errors on `Reorder.Item`. 

**2. IMPORTANT: Provide a CodeSandbox reproduction of the bug**

https://stackblitz.com/edit/vitejs-vite-6fwrx6bw?file=src%2FApp.tsx

**3. Steps to reproduce**

Use `Reorder.Item` on `motion@12.23.12`. Provide any props that don't make sense (e.g. `className={1}`) and see that no TS error is reported.

**4. Expected behavior**

Typescript should report errors like e.g. with `motion.div`.

m2na7:
Fixes #3359

## Problem

There was an issue in the `Reorder.Item` component where using invalid props (e.g., onClick="error") did not trigger TypeScript errors, resulting in reduced type safety compared to motion.div.

## Solution

I improved the Props interface by making it generic with a `Tag` parameter to support dynamic HTML element types.
Additionally, I enhanced type inference accuracy by using `HTMLMotionProps<Tag>` and `DOMMotionComponents`, and modified the forwardRef type to use specific HTML element types instead of `any`, thereby improving type safety by eliminating the use of `any` types.

### DIFF
--- a/packages/framer-motion/src/components/Reorder/Item.tsx
+++ b/packages/framer-motion/src/components/Reorder/Item.tsx
@@ -3,22 +3,23 @@
 import { isMotionValue } from "motion-dom"
 import { invariant } from "motion-utils"
 import * as React from "react"
-import { forwardRef, FunctionComponent, useContext } from "react"
+import { forwardRef, useContext } from "react"
 import { ReorderContext } from "../../context/ReorderContext"
 import { motion } from "../../render/components/motion/proxy"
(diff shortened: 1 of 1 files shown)


mattgperry:
The tests were failing for this so replacing with https://github.com/motiondivision/motion/pull/3368

PR #3366, opened 2025-09-16
   fix: update useMotionRef to properly handle externalRef changes

### LINKED ISSUE #3361: [BUG] Updating the ref to a motion element does not work
When updating a `ref` on a `motion` component (e.g., `m.div`), the old ref is not cleared and the new ref is not set. This results in stale references when swapping between ref objects after mount.

https://codesandbox.io/p/sandbox/dxzgcp

**4. Steps to reproduce**

1. Mount the component above.
2. Check the console logs.
3. Observe that `ref1.current` is still populated, and `ref2.current` is `null`.

**5. Expected behavior**

* Before swap: `ref1.current` should point to the `div`, `ref2.current` should be `null`.
* After swap: `ref1.current` should be `null`, `ref2.current` should point to the `div`.

**6. Video or screenshots**
Image

**7. Environment details**

* motion version: \12.23.12
* React version: \19.1.1
* Browser: \Chrome

AmitSingh-5600:
## Description
   Fixes issue where motion components don't update refs when externalRef changes.

   ## Root Cause
   The `useMotionRef` function wasn't including `externalRef` in the `useCallback` dependency array, causing the ref callback to not update when the external ref changes.

   ## Changes
   - Add `externalRef` to the dependency array in `useMotionRef`
   - Add test case to verify ref changes work correctly
   - Update CHANGELOG.md with the fix

   ## Testing
   - Added comprehensive test that reproduces the exact issue from #3361
   - Test verifies that refs are properly cleared and updated when externalRef changes
   - All existing tests continue to pass

   ## Files Changed
   - `packages/framer-motion/src/motion/utils/use-motion-ref.ts`
   - `packages/framer-motion/src/motion/__tests__/ref-change.test.tsx`
   - `CHANGELOG.md`

   Fixes #3361


mattgperry:
Thanks for the PR!

PR #3381, opened 2025-09-24
Fixing React 19 cleanup when external ref returns cleanup function

mattgperry:
Fixes https://github.com/motiondivision/motion/issues/3360


tsotimus:
Thanks @mattgperry - this is blocking our nextjs upgrade!

Issue #2023, opened 2023-03-16
[BUG] AnimatePresence doesn't update with the latest state if the state changes fast.


kyung-lee-official:
**Describe the bug**

`AnimatePresence` doesn't update with the latest state if the state changes fast.

**Provide a CodeSandbox reproduction of the bug**

```tsx
import { AnimatePresence, motion } from "framer-motion";
import { useState } from "react";

const ComponentWrapper = (props: any) => {
	const { status } = props;
	return (
		<motion.div
			initial={{ opacity: 1 }}
			animate={{ opacity: 1 }}
			exit={{ opacity: 0 }}
		>
			<h1>{status}</h1>
		</motion.div>
	);
};

function ApBug() {
	const [status, setStatus] = useState<number>(1);
	return (
		<div>
			<AnimatePresence mode="wait">
				<ComponentWrapper key={status} status={status} />
			</AnimatePresence>
			<h1>status: {status}</h1>
			<button
				onClick={() => {
					setStatus(status + 1);
				}}
			>
				+1
			</button>
		</div>
	);
}

export default ApBug;
```

[CodeSandbox](https://codesandbox.io/p/sandbox/vite-animate-presence-zd5h44?file=%2Fsrc%2FApp.tsx&selection=%5B%7B%22endColumn%22%3A22%2C%22endLineNumber%22%3A37%2C%22startColumn%22%3A22%2C%22startLineNumber%22%3A37%7D%5D)

**Steps to reproduce**

If the `+1` button was clicked slowly, the animation will update with the state. While if the button was clicked swiftly, the animation stops updating with the state.

**Expected behavior**

The animation should update with the state.

**Video or screenshots**

If applicable, add a video or screenshots to help explain the bug.

https://user-images.githubusercontent.com/89364151/225664502-ac68c6cb-1979-418d-bde1-3710385d8049.mp4

**Environment details**

OS: 
Edition	Windows 11 Pro
Version	22H2
Installed on	‎11/‎10/‎2022
OS build	22621.1413
Experience	Windows Feature Experience Pack 1000.22639.1000.0

Browser:
Chrome 
Version 111.0.5563.65 (Official Build) (64-bit)

npm package versions:

```json
{
  "name": "animation-sandbox",
  "private": true,
  "version": "0.0.0",
  "type": "module",
  "scripts": {
    "dev": "vite",
    "build": "tsc && vite build",
    "preview": "vite preview"
  },
  "dependencies": {
    "framer-motion": "^10.4.0",
[...]

PR #3384, opened 2025-09-26
 Bug Fix: Prevent Duplicate Exit Animations in AnimatePresence

seal7766:
### 📋 Problem

AnimatePresence component can process the same exit animation multiple times when receiving rapid sequential DOM events. This occurs when:

1. User clicks a dismissable overlay (first `pointerdown` event)
2. The overlay's dismissal propagates another event immediately
3. Both events trigger state updates that call the same component's exit handler
4. The component attempts to animate out twice, causing visual glitches

This is particularly common with UI libraries like **Radix UI**, where dismissable layers can fire multiple events in quick succession.

### ✨ Solution

Track components currently processing exit animations to prevent duplicate handling:

```typescript
// Add tracking for components in exit process
const exitingComponents = useRef(new Set<ComponentKey>())

const onExit = () => {
    // Skip if already processing this component's exit
    if (exitingComponents.current.has(key)) {
        return
    }

    exitingComponents.current.add(key)

    // Handle edge cases and perform exit...
    // Clean up tracking state when done
}
```

### 🔧 How It Works

1. When `onExit` is called, check if the component is already being processed
2. If **yes**, skip the duplicate call
3. If **no**, mark it as processing and continue with exit animation
4. Clean up the tracking state when:
   - The component finishes exiting
   - The component re-enters before exit completes
   - All exit animations complete

### ✅ Testing

- ✅ **All existing tests pass**
- ✅ **Manually verified** with multiple UI libraries (Radix UI, Arco Design)
- ✅ **No performance impact** (O(1) Set operations)
- ✅ **Works correctly** in React StrictMode

### 🔗 Related Issues

- #2554 - AnimatePresence exit animations triggered multiple times
- #2023 - Duplicate animations with rapid state changes

---

**Impact**: Low risk, high value fix for production applications using AnimatePresence with modern UI libraries.

### DIFF
--- a/packages/framer-motion/src/components/AnimatePresence/index.tsx
+++ b/packages/framer-motion/src/components/AnimatePresence/index.tsx
@@ -87,6 +87,12 @@ export const AnimatePresence = ({
     const exitComplete = useConstant(() => new Map<ComponentKey, boolean>())
 
     /**
+     * Track components that are currently in the process of exiting.
+     * This prevents duplicate exit handling during rapid state changes.
+     */
+    const exitingComponents = useRef(new Set<ComponentKey>())
+
+    /**
      * Save children to render as React state. To ensure this component is concurrent-safe,
      * we check for exiting children via an effect.
      */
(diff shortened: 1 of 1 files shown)


mattgperry:
Thanks for the PR! Is it possible to get a test that fails on `main` and passes on this branch?

seal7766:
@mattgperry 
 I spent considerable time trying to create a failing test for the main branch:

  Unfortunately, I couldn't reproduce the exact bug in tests because it requires:
  - Multiple library versions creating separate React contexts
  - Specific DOM event bubbling that jsdom doesn't fully replicate
  - A race condition that only occurs with real event propagation

  While I couldn't provide a failing test, the fix is straightforward defensive
  programming that prevents a real issue we've encountered and fixed in production.

 Would you thoughts on this approach?

mattgperry:
@claude Please review

mattgperry:
Thanks for this fix! It's been replaced by https://github.com/motiondivision/motion/pull/3454 which also includes some failing tests.

Issue #3392, opened 2025-10-07
[BUG] TypeError: motionMediaQuery.addEventListener is not a function


kmertozen:
TypeError: motionMediaQuery.addEventListener is not a function. (In 'motionMediaQuery.addEventListener("change", setReducedMotionPreferences)'

Environment

Device: iPhone (iOS 13.7)
Browser: Chrome (on iOS 13.7, uses WKWebView / Safari engine)

A cross-compatible implementation would be:

```
if (motionMediaQuery.addEventListener) {
  motionMediaQuery.addEventListener("change", handler);
} else  {
  motionMediaQuery.addListener(handler);
}
```

PR #3393, opened 2025-10-07
Add legacy matchMedia.addListener support for older Safari versions

kmertozen:
Fixes [#3392](https://github.com/motiondivision/motion/issues/3392)


mattgperry:
Thanks for the PR! Though closing as we don't support Safari 13

PR #3400, opened 2025-11-01
fix: Ensure useMotionValueEvent animationComplete fires for useSpring

rick-hup:


mattgperry:
Hey thanks for taking a look! Would this perhaps work better inside MotionValue.attach() ? Or is it a bit messy>

rick-hup:
@mattgperry Thanks! Good call, I've refactored and pushed the changes.

PR #3414, opened 2025-11-27
Support iOS 13 by using MediaQueryList#addListener as a fallback

ambar:
Fixed https://github.com/motiondivision/motion/issues/2823 see https://caniuse.com/mdn-api_mediaquerylistevent


mattgperry:
Thanks for the PR! However, we don't support iOS 13 so closing this without merging. A recommended alternative would be to polyfill the existing implementation if you need iOS 13 support

Issue #3417, opened 2025-12-03
[BUG] ReferenceError: process is not defined in native ES module usage


stefanpearson:
When importing `motion` as a native ES module in the browser (without any bundler/build), it throws:

```
Uncaught ReferenceError: process is not defined
    at errors.mjs:5:1
```

### Reproduction

CodeSandbox and similar online editors always define process, so a true reproduction requires running in a local browser with pure ES modules. See the minimal HTML example below.

```html
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="UTF-8">
  <title>Motion Utils process bug</title>
</head>
<body>
  <script type="module">
    import  "./node_modules/motion-utils/dist/es/errors.mjs";
  </script>
</body>
</html>
```

### Details
Version: 12.23.24

srivastava-ayush:
Its probably referencing node only globals // as process doesnt exist in native browsers

PR #3419, opened 2025-12-03
fix: guard process usage for native browser ESM

srivastava-ayush:
### Problem
Importing `motion-utils` directly in the browser as a native ES module throws
`ReferenceError: process is not defined`.

This happens because `process.env.NODE_ENV` is accessed at module evaluation
time without guarding for non-Node environments. The issue is masked when usin
bundlers that inject a `process` polyfill, but fails in pure browser ESM

### Solution
Guard access to `process` before reading `process.env.NODE_ENV` so the code can
run safely in browsers while preserving existing behavior for Node and bundler
environments.

### Result
-  Native browser ES module imports no longer throw
-  No behavior change for production builds
-  No reliance on bundler polyfill

Fixes : #3417


stefanpearson:
Thanks!

stefanpearson:
I've noticed a few other 'unsafe' references to `process.env` across various files in the `motion` packages. I'm assuming they will need updating too? https://github.com/search?q=repo%3Amotiondivision%2Fmotion%20process.env&type=code

srivastava-ayush:
I dont think so as most of them are node only files so it should not be a problem , but yeah there are 'unsafe' references to process.env inside /src  like Hooks, utils, render logic, that should be guarded as they ll be run in browser.

mattgperry:
Thanks for the PR - I realised I commited something similar today so closing without merge

Issue #3340, opened 2025-07-31
[BUG] scroll progress is negative in reverse directions


li-kai:
Repro: https://codesandbox.io/p/sandbox/y76dz6

In the following scenarios, the scroll progress value is negative:
1. `flex: column-reverse`
2. `flex: row-reverse`
3. `writing-mode: vertical-rl`

**4. Steps to reproduce**

1. Go to https://codesandbox.io/p/sandbox/y76dz6
2. Scroll all 3 boxes
3. See scroll value is negative

**5. Expected behavior**

Doc says the value should be between (0, 1).

**6. Video or screenshots**

<img width="1074" height="1097" alt="Image" src="https://github.com/user-attachments/assets/a89d30ed-afd3-4a8e-ae36-67bafb9a7d67" />

**7. Environment details**

I was able to repro in Chrome, Safari and Firefox.

PR #3421, opened 2025-12-04
fix: Correct scroll progress in reverse layouts

wakront:
### Problem:
  In reverse layouts such as flex-direction: *-reverse, writing-mode: vertical-rl, or direction: rtl, the scroll progress (progress) was incorrectly calculated as a negative value, ranging from 0 to -1, instead of being correctly mapped between 0 and 1. Additionally, due to floating-point precision issues in browsers, the progress sometimes did not reach exactly 1, ending instead with values like 0.999....

### Solution:
  Within the updateAxisInfo function, getComputedStyle is used to check the element's flexDirection, writingMode, and direction properties to determine if the layout is reversed. If the layout is reversed, the axis.current value is rounded using Math.round() and its sign is inverted before being passed to the progress function. By using `Math.round()` to convert the `axis.current` value to an integer, floating-point precision issues caused by browser sub-pixel rendering are mitigated, and the accuracy of the `progress` calculation is enhanced. This ensures that the progress is correctly calculated within the 0 to 1 range.

### Result:
* Scroll progress is correctly calculated within the 0 to 1 range for reverse layouts (e.g., flex-direction: column-reverse, row-reverse, writing-mode: vertical-rl, direction: rtl).

Fixes: #3340

### DIFF
--- a/packages/framer-motion/cypress/integration/scroll-reverse.ts
+++ b/packages/framer-motion/cypress/integration/scroll-reverse.ts
@@ -29,7 +29,7 @@ const calculateExpectedProgress = (
 
 describe("Scroll progress in reverse-direction containers", () => {
     beforeEach(() => {
-        cy.visit("?test=scroll-reverse")
+        cy.visit("?test=scroll-reverse").wait(1000)
         cy.get("#scroller-1").should("be.visible")
     })
 


Heeesang:
Hi! When you get a moment, I’d love your thoughts on this PR.
If the implementation needs to move in a different direction, I’m totally open to revising it. @mattgperry

mattgperry:
@Heeesang @wakront This looks great though I'm not too knowledgable on this problem. What would be awesome are some tests in `dev/react/src/tests` and `framer-motion/cypress/integration`

wakront:
Thanks for the feedback! I added the requested tests under dev/react/src/tests and framer-motion/cypress/integration as suggested. I haven’t used Cypress extensively yet, so the setup might not be perfect, but I followed the patterns in the codebase and aligned the scenarios with the intended behavior. If there's anything that could be improved or refined, I’d be happy to adjust it.

The tests all pass locally, but the CI environment is failing on the run. 
Can you check locally and help me? @mattgperry

wakront:
I checked the behavior and the tests are failing only in the Electron environment. They pass without issues in Chrome. Is there a recommended approach for handling this difference or adjusting the tests for Electron? @mattgperry I'm sorry you might be busy. Please leave a comment when you have time.

wakront:
@mattgperry Just to follow up with a bit more context.

From what I’ve observed, the failing behavior only occurs in the Electron environment.
The same tests pass consistently in Chrome.

Would you consider Electron to be within the supported testing scope for this project?
If not, one option could be to conditionally skip or adjust these tests when running under Electron.
Alternatively, I’m happy to explore an Electron-specific fix if that’s preferred.

Let me know which direction makes the most sense.

wakront:
@ mattgperry quick ping on this 🙂

Given that the failure only reproduces in Electron CI and not in Chrome,
would you prefer:

1) skipping or relaxing these tests under Electron, or
2) adjusting the implementation to explicitly support Electron behavior?

Happy to take either approach — just want to align before making changes.

wakront:
@mattgperry I’m planning to proceed with the following unless you see any issues:

- Treat Chrome as the primary supported environment for these scroll tests
- Skip or relax the failing assertions when running under Electron CI, since the behavior appears to be Electron-specific and does not reproduce in Chrome

This keeps the test coverage for reverse layouts intact while avoiding CI instability from Electron quirks.

If you’d prefer explicit Electron support instead, I can investigate that path as well — but my recommendation is to scope these tests to Chrome for now.

Let me know if you’d like this handled differently.

Issue #3428, opened 2025-12-18
[BUG] React 19 refs


NoelDeMartin:
## Describe the bug

When using a component with [React 19 refs](https://react.dev/reference/react/useImperativeHandle) inside a popLayout animation, I get the following error:

> Accessing element.ref was removed in React 19. ref is now a regular prop. It will be removed from the JSX Element type in a future release.

## CodeSandbox reproduction

https://codesandbox.io/p/devbox/x358rj

TLDR, it's a vanilla Next.js app with motion and I only modified the home page:

```tsx
"use client";

import { AnimatePresence, motion } from "motion/react";
import { Ref, useImperativeHandle, useRef, useState } from "react";

interface FirstHandle {}

function First({ ref }: { ref: Ref<FirstHandle> }) {
  useImperativeHandle(ref, () => ({}));

  return <motion.span>First</motion.span>;
}
function Second() {
  return <span>Second</span>;
}

export default function Home() {
  const [showFirst, setShowFirst] = useState(true);
  const firstRef = useRef<FirstHandle>(null);

  return (
    <div>
      <button onClick={() => setShowFirst(!showFirst)}>Toggle</button>

      <br />
      <AnimatePresence mode="popLayout">
        {showFirst ? <First ref={firstRef} /> : <Second />}
      </AnimatePresence>
    </div>
  );
}
```

## Steps to reproduce

Open the code sandbox, and you'll see the error.

## Expected behavior

It shouldn't throw an error.

## Further details

I have pinpointed the issue to this part of the code in the `PopChild` component, although it may be an issue in other parts of the code as well:

https://github.com/motiondivision/motion/blob/611c7e80a94d24c1b04e99733c683f9fe7d84c3a/packages/framer-motion/src/components/AnimatePresence/PopChild.tsx#L75-L78

PR #3429, opened 2025-12-22
fix: Refactor ref handling in PopChild component

srivastava-ayush:
In React 19, ref is no longer a special property on the element, Its just a normal prop so we have to do  **children.props.ref** not  **children.ref**

fixes: #3428


mattgperry:
Thanks for the PR! How does this work with React 18?

srivastava-ayush:
> Thanks for the PR! How does this work with React 18?


Thanks for reviewing, earlier fix won't work with React 18. I've **updated** the PR to be backwards compatible by adding **nullish coalescing**.

NoelDeMartin:
@constayush Maybe it should read from `props?.ref` before trying to do `.ref`? I'm pretty sure the issue I opened would still get triggered with this code, because when reading `.ref` you don't get undefined, it throws an error.

Issue #3413, opened 2025-11-26
[BUG] useTransform doesn't work well with padding-block using numbers and other logical values


Calvein:
**1. Read the [FAQs](#faqs) 👇**
✅

**2. Describe the bug**

use

**3. IMPORTANT: Provide a CodeSandbox reproduction of the bug**

I hope stackblitz is fine: https://stackblitz.com/edit/vitejs-vite-4d7vhxcr?file=src%2FApp.tsx&terminal=dev

**4. Steps to reproduce**

1. Scroll the example
2. The 2 elements are very similar, one using `useTransform` with numbers and one with strings
3. When scrolling, the one using number jumps from 100 to 0 and never goes back to 100

**5. Expected behavior**

It should work like any other css values accepting pixels, e.g: padding

mattgperry:
Fixed between this version and now.

PR #3430, opened 2025-12-27
fix: support numeric values for CSS logical properties in useTransform

reeseo3o:
## Issue
- #3413

## Summary

`useTransform` did not correctly convert numeric values to pixel units when used with CSS logical properties (e.g. `paddingBlock`, `paddingInline`, `marginBlock`, `marginInline`), causing styles to update incorrectly.

This PR fixes the missing px conversion so CSS logical properties are handled consistently with other CSS properties.

## Root Cause

CSS logical properties were missing from the internal px conversion mappings (`numberValueTypes` and `pxValues`).

As a result, numeric values returned from `useTransform` were written to the DOM without units and treated as invalid CSS.
Because CSS allows unitless `0`, intermediate values were ignored, making updates appear to jump from `0` to the final value.

## Solution

CSS logical properties were added to both px conversion paths.

### Updated files

- `packages/motion-dom/src/value/types/maps/number.ts`
- `packages/motion-dom/src/animation/waapi/utils/px-values.ts`

Numeric values returned from `useTransform` are now converted to `"px"` before being applied to the DOM.

### As-is
- Numeric values not converted to `px`
- Intermediate values ignored
- Values appear to jump during scroll

https://github.com/user-attachments/assets/74f26186-02c5-4d67-bec1-e3d213b7b579

### To-be
- Numeric values converted to pixel strings
- Intermediate values applied correctly
- Smooth updates during scroll

https://github.com/user-attachments/assets/c88c8a84-833b-4e35-a739-1091e9439614


shinseongsu:
Great PR! While you're at it, would it be possible to add support for the `inset` shorthand and its related logical properties?

Since they share the same shorthand logic as `margin` and `padding`, including them would ensure much better consistency. Specifically, I suggest adding the following properties to `px-values.ts` and `number.ts`

```
inset
insetBlock
insetBlockStart
insetBlockEnd
insetInline
insetInlineStart
insetInlineEnd
```

###  As-is
Current Issue Currently, when using CSS variables within these shorthand properties, they are not parsed correctly, as shown in the screenshots below

https://github.com/user-attachments/assets/cd48f00d-5810-4025-8cd1-c8c9b7053a85

###  To-be 
Expected Behavior By adding these to px-values.ts and number.ts, Framer Motion should be able to parse them correctly

https://github.com/user-attachments/assets/2c1a1d90-e63c-4a75-be57-b60bffb88725

I also noticed that #3413 is very similar to this. Since you're already "under the hood" and working on this part of the code, maybe we could bundle them together? It’d be awesome to have all these layout shorthands supported at once—it would make the CSS variable support so much more robust!

What do you think about tossing these into your PR?

reeseo3o:
> Great PR! While you're at it, would it be possible to add support for the `inset` shorthand and its related logical properties?
> 
> Since they share the same shorthand logic as `margin` and `padding`, including them would ensure much better consistency. Specifically, I suggest adding the following properties to `px-values.ts` and `number.ts`
> 
> ```
> inset
> insetBlock
> insetBlockStart
> insetBlockEnd
> insetInline
> insetInlineStart
> insetInlineEnd
> ```
> 
> ### As-is
> Current Issue Currently, when using CSS variables within these shorthand properties, they are not parsed correctly, as shown in the screenshots below
> 
>  2025-12-28.3.22.24.mov 
> ### To-be
> Expected Behavior By adding these to px-values.ts and number.ts, Framer Motion should be able to parse them correctly
> 
>  2025-12-28.3.24.57.mov 
> I also noticed that #3413 is very similar to this. Since you're already "under the hood" and working on this part of the code, maybe we could bundle them together? It’d be awesome to have all these layout shorthands supported at once—it would make the CSS variable support so much more robust!
> 
> What do you think about tossing these into your PR?

Good catch! I agree this should be supported as well.

Since the `inset*` properties share the same shorthand logic and root cause,
I'll add them to the px conversion mappings in this PR for consistency.

Thanks for sharing the examples!

PR #3451, opened 2026-01-06
Add MotionValueState infrastructure for atomic DOM updates

mattgperry:
## Summary

- Introduces `MotionValueState` class hierarchy to enable per-value DOM rendering instead of rebuilding all styles every frame
- Adds `HTMLMotionValueState` for atomic HTML style updates (transform, CSS variables, regular styles)
- Adds `SVGMotionValueState` for atomic SVG attribute updates
- Adds `createMotionValueState()` abstract method to `VisualElement` with implementations in HTML, SVG, Object, and State visual elements
- Enhances base `MotionValueState` with `forEach()`, `getStatic()`, and `prevValue` tracking

## Technical Notes

The state infrastructure is in place but values are currently bound via `bindToMotionValue` to maintain backward-compatible `onUpdate` behavior. This prepares the codebase for future atomic update optimizations where:

- Transform props share a computed `transform` MotionValue
- Each style property gets its own atomic render callback
- DOM writes happen per-value instead of full style rebuilds

## Test plan

- [x] All existing tests pass (680 tests)
- [x] CSS variables tests pass
- [x] Variant stagger order tests pass
- [x] onUpdate callback tests pass
- [x] Keyframe transition tests pass

### DIFF
--- a/dev/html/public/benchmarks/transform-rebuild-overhead.html
+++ b/dev/html/public/benchmarks/transform-rebuild-overhead.html
@@ -1,0 +1,175 @@
+<html>
+    <!--
+      Transform Rebuild Overhead Benchmark
+
+      This benchmark measures the overhead of rebuilding transforms when
+      animating non-transform properties. In the old renderer, animating
+      backgroundColor would trigger a full transform string rebuild.
+      With atomic updates, only backgroundColor would be updated.
+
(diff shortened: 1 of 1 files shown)


mattgperry:
Closing for now as although interesting, not quite as large of an improvement for such a big change

PR #3486, opened 2026-01-19
Allow drag within rotated parents

### LINKED ISSUE #3320: [BUG] Draggable direction is incorrect when parent container is rotated
> [!NOTE]
> **Motion for Vue issues:** Please open in the [Motion for Vue repo](https://github.com/motiondivision/motion-vue).

**2. Describe the bug**

When a draggable is in a rotated parent container, the draggable direction is wrong.

https://playcode.io/2465463

**4. Steps to reproduce**

1. Go to https://playcode.io/2465463
2. Drag the white square
3. Observe

**5. Expected behavior**

The white square should move with the mouse. Instead it moves opposite to the mouse.

**6. Video or screenshots**

https://github.com/user-attachments/assets/e8b31e1f-3cd2-498c-bbd0-13167cdf6895

**7. Environment details**

I don't think this is applicable, but Chrome on Mac.

mattgperry:
When a draggable element is inside a rotated parent container, the drag offset needs to be transformed from screen space to the element's local coordinate space.

This fix:
- Adds applyParentRotation() method that computes cumulative rotation from all ancestor motion elements in the projection path
- Applies inverse 2D rotation to both drag offset and velocity
- Uses minimal bundle impact: only Math.cos/sin, no new dependencies
- Leverages existing projection node data (path, latestValues)

Fixes #3320

### DIFF
--- a/packages/framer-motion/src/gestures/drag/VisualElementDragControls.ts
+++ b/packages/framer-motion/src/gestures/drag/VisualElementDragControls.ts
@@ -2,6 +2,7 @@ import {
     addValueToWillChange,
     animateMotionValue,
     calcLength,
+    cancelFrame,
     convertBoundingBoxToBox,
     convertBoxToBoundingBox,
     createBox,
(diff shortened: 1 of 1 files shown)


mattgperry:
Closing as although it works well for static rotations I feel like a more comprehensive approach would take into rotate animations also and it's not quite there.

PR #3508, opened 2026-01-23
feat: scroll page when Reorder container does not have overflow `auto` or `scroll`

### LINKED ISSUE #3469: [BUG] Reorder does not work as expected in a scrollable page
> [!NOTE]
> **Motion for Vue issues:** Please open in the [Motion for Vue repo](https://github.com/motiondivision/motion-vue).

**2. Describe the bug**

Give a clear and concise description of what the bug is.

I was looking forward to using the [Reorder](https://motion.dev/docs/react-reorder) feature with [Auto-scroll lists](https://motion.dev/docs/react-reorder#auto-scroll-lists), however, upon using a slightly different version than the contrived example it doesn't work.

I believe this [example from Motion.dev docs](https://examples.motion.dev/react/todo-list) should be sufficient.

**4. Steps to reproduce**

1. Go to [Reorder's To-do List example](https://examples.motion.dev/react/todo-list)
2. Attempt to reorder items within scroll boundaries (Succeed)
3. Open `Dev Tools`
4. Select `<body>` element
5. Set `height` value to `200vh`
6. Once again, attempt to reorder items within scroll boundaries (Failure)

**5. Expected behavior**

I'd expect the reorder feature to work within a scrollable page.

**6. Video or screenshots**

_Note: had to reduce file size by compressing the video, sorry if it looks awful_

https://github.com/user-attachments/assets/b8972c68-8096-4ad7-a7aa-81cd81d0d62e

**7. Environment details**

If applicable, let us know which OS, browser, browser version etc you're using.

Browser: Brave 1.85.118 (Official Build) (arm64) — Chromium: 143.0.7499.169
OS: macOS Sonoma 14.5

### React Server Components `"use client"` error

If you're importing `motion` or `m` into a React Server Component environment, ensure you're importing from `motion/react-client` instead of `motion/react`.

```javascript
import * as motion from "motion/react-client"
import * as m from "framer-motion/react-m"
```

### Motion for React won't install

Different versions of Motion for React are compatible with different versions of React.

React 19: `framer-motion@12.0.0-alpha.0` or higher
React 18: `framer-motion@7.0.0` to `framer-motion@11.x`, or `motion`
React 17: `framer-motion@6.x` or lower

### `height: "auto"` is jumping

Animating to/from `auto` requires measuring the DOM. There's no perfect way to do this and if you have also applied padding to the same element, these measurements might be wrong.

The recommended solution is to move padding to a child element. See [this issue](https://github.com/motiondivision/motion/issues/368) for the full discussion.

### Preact isn't working

Motion for React isn't compatible with Preact.

### `AnimatePresence` isn't working

Have all of its immediate children got a unique `key` prop that **remains the same for that component every render**?

```jsx
// Bad: The index could be given to a different component if the order of items changes
<AnimatePresence>
    {items.map((item, index) => (
        <Component key={index} />
    ))}
</AnimatePresence>
```

```jsx
// Good: The item ID is unique to each component
<AnimatePresence>
    {items.map((item, index) => (
        <Component key={item.id} />
    ))}
</AnimatePresence>
```

Is the `AnimatePresence` correctly outside of the controlling conditional? `AnimatePresence` must be rendered whenever you expect an `exit` animation to run - it can't do so if it's unmounted!

```jsx
// Bad: AnimatePresence is unmounted - exit animations won't run
{
    isVisible && (
        <AnimatePresence>
            <Component />
        </AnimatePresence>
    )
}
```

```jsx
// Good: Only the children are unmounted - exit animations will run
<AnimatePresence>{isVisible && <Component />}</AnimatePresence>
```

lgaspari:
Fixes #3469 

# Summary

- When closest scrollable `element` is either `document.body` or `document.documentElement`, then:
  - Max scroll is determined by window's `innerWidth` or `innerHeight` based on `axis`
  - Current scroll is determined by window's `scrollX` or `scrollY` based on `axis`
  - Scroll is performed on `window` invoking `scrollBy` method

# Test plan

- [x] Cypress tests for `reorder-auto-scroll` pass
- [x] Manual testing with scrollable container
- [x] Verify existing Reorder functionality unchanged

# Evidence

https://github.com/user-attachments/assets/0ebb05d0-f6bb-4cb2-a0f3-54221aca3ba8

### DIFF
--- a/dev/react/src/tests/reorder-auto-scroll-container.tsx
+++ b/dev/react/src/tests/reorder-auto-scroll-container.tsx
@@ -40,12 +40,14 @@ export const App = () => {
             {/* Spacer to make page scrollable */}
             <div style={{ height: "300px", background: "#222" }}>
                 <p style={{ color: "#fff", padding: "20px" }}>
-                    Scroll down to see the reorder list. The page is scrollable.
+                    Scroll down to see the reorder list. The container and the page are both scrollable.
                 </p>
             </div>
             <div
                 data-testid="scroll-container"
                 style={{
+                    height: "300px",
+                    overflow: "auto",
                     margin: "0 auto",
                     width: "300px",
                     background: "#444",
--- a/packages/framer-motion/cypress/integration/reorder-auto-scroll.ts
+++ b/packages/framer-motion/cypress/integration/reorder-auto-scroll.ts
@@ -74,7 +74,7 @@ describe("Reorder auto-scroll", () => {
 
     describe("with scrollable container inside scrollable page", () => {
         it("Auto-scrolls container after page has been scrolled", () => {
-            cy.visit("?test=reorder-auto-scroll-page")
+            cy.visit("?test=reorder-auto-scroll-container")
                 .wait(200)
                 // Scroll the page down so the container is in view
                 .window()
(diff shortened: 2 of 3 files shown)


lgaspari:
Hey @mattgperry, sorry for the direct ping. When you have a moment, would you be able to take a look at this PR?

It's my first time contributing, so I'd really appreciate any feedback in case I missed something. I'm happy to make adjustments if needed to get this merged. LMK!

Thanks a lot for your time!
Lean

mattgperry:
Thanks for the PR! Next week is maintenance week so I’ll be taking a good look then!

mattgperry:
Awesome contribution! Thanks for the PR

Issue #3315, opened 2025-07-17
[BUG] If element's layout position shifts horizontally post-drag, the drag offset remains


JonAbrams:
**1. Read the [FAQs](#faqs) 👇**

**2. Describe the bug**

If the result of dragging and dropping a `motion.div` is that its position is moved horizontally, the drag transform sticks around, causing the element to end up much farther horizontally.

**3. IMPORTANT: Provide a CodeSandbox reproduction of the bug**.

https://codesandbox.io/p/sandbox/6nmt2v

**4. Steps to reproduce**

In the codesandbox:

1. Drag a tile onto a tile in another row. 
2. If the tile is dragged to another row, they swap, as expected.
3. Drag a tile onto another tile in the same row.
4. The dropped tile flys off horizontally, ending up in the wrong place

**5. Expected behavior**

I expect the dropped element to be rendered as intended, with no transform applied.

**6. Video or screenshots**

<img width="393" height="387" alt="Image" src="https://github.com/user-attachments/assets/65847cb5-8286-4481-86fb-5c3fbf739d1e" />

Notice that 8 and 14 swapped ok, but 1 is offset to the right by the same distance it was dragged from its origin.

**7. Environment details**

Happens in Chrome and Safari

Things I've tried, but didn't work for me:
- Instead of setting `top` and `left`, I've tried `x` and `y`.
- Applying `transform` with the `transformTemplate`

WaqarNiaxi:
Hi, I’ve reviewed this issue and would like to work on it. Could you please assign it to me?

PR #3513, opened 2026-01-29
fix: prevent drag overshoot when layout changes with dragSnapToOrigin and layoutId #3315

WaqarNiaxi:
Fixed a bug where elements using both `dragSnapToOrigin` and `layoutId` would overshoot their target position when the layout changes after dragging ends.

## Problem

When an element with `dragSnapToOrigin` and `layoutId` completes a drag gesture, two animations occur:
1. `layoutId` animates the CSS position to the new layout position
2. `dragSnapToOrigin` animates the transform back to 0

If the layout changes while the `dragSnapToOrigin` animation is still running, the transform offset combined with the new CSS position causes the element to overshoot and appear in the wrong location.

## Solution

Added logic to detect when layout changes occur after dragging ends. When this happens:
1. Stop all drag animations immediately
2. Reset all motion values (x/y) to 0
3. Allow the `layoutId` animation to position the element correctly

This ensures the element lands precisely at its target position without overshoot.

## Related Issue

Fixes the drag and drop test case where swapping tiles would cause tiles to appear offset.

## Testing

- Created a test case in `/dev/next/app/drag-test.tsx` demonstrating the fix
- The test swaps tiles in a grid and verifies the drag transform doesn't interfere with layout positioning
- All existing drag tests continue to pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ Motion adheres to [Semantic Versioning](http://semver.org/).
 
 Undocumented APIs should be considered internal and may change without warning.
 
-## [Unreleased]
+## [12.29.3] 2026-01-29
 
 ### Fixed
 


mattgperry:
Can you open an issue with a failing codesandbox?

PR #3516, opened 2026-01-30
Fix: fix simple SPA layout animations

MrVoshel:
Avoid stale shared layout nodes during SPA navigations and add an HTML regression fixture for disconnected resumeFrom cases.

I encountered this problem on my SvelteKit app, but I suppose it affects other frameworks that approach navigation in a similar way as well. Svelte provides an onDestroy utility, but I wasn't able to get it to clean old nodes after the first navigation, only a full page refresh "fixed" it.

I suppose the additional checks shouldn't cause a perf regression since it happens only when adding or promoting, but I'm not very familiar with the codebase.

**_Please check carefully the logic and the test since I've narrowed down the issue thanks to Opus and GPT Codex_**, worked fine on my codebase, but haven't done anything extensive or stress tests.

Before the fix:

https://github.com/user-attachments/assets/acad3645-6ee0-4447-a41d-8b2ba5ebcb85

After the fix:

https://github.com/user-attachments/assets/beaddb25-c1a3-431c-b378-5b883f071877

### DIFF
--- a/dev/html/public/animate-layout/shared-element-spa-repeat.html
+++ b/dev/html/public/animate-layout/shared-element-spa-repeat.html
@@ -1,7 +1,8 @@
 <html>
     <!--
-     This test simulates stale shared layout nodes across SPA navigations.
-     It verifies resumeFrom is not set to a disconnected instance.
+     Test for stale shared layout nodes across SPA navigations.
+     Verifies that resumeFrom is not set to disconnected instances and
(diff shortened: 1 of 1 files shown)


MrVoshel:
Clueless on why it fails on CircleCI 
image

EDIT: It was probably because of their infra issues

mattgperry:
Closed by https://github.com/motiondivision/motion/pull/3543

PR #3538, opened 2026-02-04
Add `transition.inherit` to allow transition composition

mattgperry:
## Summary

- Adds `inherit: true` to `Transition`


lochie:
yessss lets goooooo 🎉

PR #3548, opened 2026-02-08
docs: fix README contributing guide link

pavan-sh:
The README contribution link still referenced the old motiondivision/motion repo and master branch.
This updates it to a relative CONTRIBUTING.md link so it stays correct regardless of default branch/org changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ A one-time payment, lifetime-updates membership:
 
 ## 💎 Contribute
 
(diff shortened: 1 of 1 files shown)


mattgperry:
This link wouldn't work outside the Github context but the readme does appear elsewhere and there's no guarantee that the link will work there.

pavan-sh:
> This link wouldn't work outside the Github context but the readme does appear elsewhere and there's no guarantee that the link will work there.

Got it — thanks for the clarification. Makes sense to keep the link working in non‑GitHub contexts as well.

PR #3551, opened 2026-02-09
Fix Motion package README download shields

pavan-sh:
The Motion package README showed npm downloads and jsDelivr hit badges for framer-motion.
This updates both shields to correctly point at the motion package, and fixes the jsDelivr logo parameter.


mattgperry:
Closing this out as the `framer-motion` package actually more represents usage as not everyone has updated to `motion` import.

PR #3554, opened 2026-02-11
Fix: SPA navigation (once again)

MrVoshel:
Follow up on #3516 

Unfortunately #3543 was on the wrong track, nothing improved with client side navigations. Pruning worked correctly there, but didn't fix the animation skip shown in my original PR.

I also tried to simplify the overall code, made the test more accurate, and did some further debugging that made me find a leak where the flat tree nodes were growing in size (since pruning was scoped to the members of add() only).

Here the animations are triggered correctly, pruning is basically the same + unmount()

I tried setting the new test runs to 100 and checking the js heap size.
Now
image

Without unmount (pruning only stack.ts' members)
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,28 @@ Motion adheres to [Semantic Versioning](http://semver.org/).
 
 Undocumented APIs should be considered internal and may change without warning.
 
-## [12.34.0] 2026-02-09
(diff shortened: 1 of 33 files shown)


MrVoshel:
test-react-19 🤷
image

mattgperry:
Thanks for your work on this!

PR #3570, opened 2026-03-02
Add ViewTimeline support for scroll-linked animations

mattgperry:
## Summary

- Use native `ViewTimeline` API when available to drive element-in-viewport scroll animations on the compositor thread, matching the existing `ScrollTimeline` progressive enhancement pattern
- Map Motion's 4 preset offsets (`Enter`, `Exit`, `Any`, `All`/default) to ViewTimeline named ranges (`entry`, `exit`, `cover`, `contain`); string and custom `ProgressIntersection` offsets fall back to JS polling as before
- Add `supportsViewTimeline` feature detection, `rangeStart`/`rangeEnd` support in `TimelineWithFallback` and `NativeAnimation.attachTimeline()`

## Test plan

- [x] Unit tests: 8 tests for offset-to-range mapping (all presets, default, string offsets, non-preset arrays)
- [x] Updated `use-scroll` tests: 6 tests covering ViewTimeline acceleration with target, no acceleration when unsupported, no acceleration with string offsets
- [ ] E2E: new `scroll-view-timeline` Cypress spec verifying accelerate flag for default/preset/string offsets
- [ ] E2E: existing `scroll-target-transform` and `scroll-accelerate` specs pass unchanged
- [ ] Manual: Chrome 115+ with `element.getAnimations()` showing ViewTimeline-driven WAAPI animation

### DIFF
--- a/dev/react/src/tests/scroll-image-reveal.tsx
+++ b/dev/react/src/tests/scroll-image-reveal.tsx
@@ -1,0 +1,107 @@
+import { motion, useScroll, useTransform } from "framer-motion"
+import * as React from "react"
+import { useRef } from "react"
+
+const colors = ["#e63946", "#457b9d", "#2a9d8f", "#e9c46a", "#f4a261", "#264653"]
+
+function RevealImage({
+    color,
+    index,
+    aspectRatio,
+}: {
+    color: string
+    index: number
+    aspectRatio: string
+}) {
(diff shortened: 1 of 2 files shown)


bakura10:
Awesome! For many of our offsets we have values in px/percentage. Those are working when using view timeline in CSS, is there a reason they can't work with ViewTimeline as well?

PR #3627, opened 2026-03-10
Add layout="x" and layout="y" axis constraints

### LINKED ISSUE #1972: [BUG] layoutId has incorrect positioning when height adjusts.

### LINKED ISSUE #3103: [FEATURE] A way to control transition when using layoutId
I'm frustrated that the indicators are animating vertically when other elements are resizing when using `layoutId`. I have a tab with indicator similar to the [example](https://motion.dev/docs/react-layout-animations#shared-layout-animations) in the doc.

A way to limit x or y animation when using `layoutId`

- Tried adding `layout` to the wrapper which isn't ideal since that will animate the entire component vertically.

**Additional context**
You can replicate what I'm frustrated with by adding a div on top of the Tab in the example and changing its height on click without animation.

Div will change its height instantly while the indicator will float outside of the Tab and animate to its position

mattgperry:
## Summary

- Adds `layout="x"` and `layout="y"` as new values for the `layout` prop
- When `layout="x"`, only the x-axis layout change is animated — y-axis changes apply instantly
- When `layout="y"`, only the y-axis layout change is animated — x-axis changes apply instantly
- Works with both `layout` and `layoutId`

Fixes #3103
Fixes #1972

## How it works

In the projection system's `notifyLayoutUpdate`, when `animationType` is `"x"` or `"y"`, the snapshot for the non-animated axis is set to match the current layout. This eliminates the delta on that axis so no animation occurs — the element snaps instantly to its final position on that axis while smoothly animating on the other.

## Test plan

- [x] Added Cypress E2E test for `layout="x"` — verifies x-axis animates while y-axis snaps
- [x] Added Cypress E2E test for `layout="y"` — verifies y-axis animates while x-axis snaps
- [x] All existing layout tests still pass
- [x] Tests pass on both React 18 and React 19
- [x] Full build passes
- [x] All unit tests pass (92 suites, 766 tests)

### DIFF
--- a/packages/motion-dom/src/projection/node/create-projection-node.ts
+++ b/packages/motion-dom/src/projection/node/create-projection-node.ts
@@ -30,7 +30,7 @@ import { mixNumber } from "../../utils/mix/number"
 import { MotionValue, motionValue } from "../../value"
 import { resolveMotionValue } from "../../value/utils/resolve-motion-value"
 import { mixValues } from "../animation/mix-values"
-import { copyAxisDeltaInto, copyBoxInto } from "../geometry/copy"
+import { copyAxisDeltaInto, copyAxisInto, copyBoxInto } from "../geometry/copy"
 import {
     applyBoxDelta,
     applyTreeDeltas,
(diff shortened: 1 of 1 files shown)


looskie:
amazing ship, thank you ❤️❤️🫡

PR #3628, opened 2026-03-11
fix transform motion svg-attribute conflicts with style-transform

### LINKED ISSUE #3082: [BUG] MotionValues cannot be used in `motion.g` element's transform attribute

**2. Describe the bug**

Unlike other SVG attributes (such as a circle's `cx` attribute) motion values cannot be used with a `<motion.g>` element's `transform` attribute.
 
See [this sandbox](https://codesandbox.io/p/sandbox/transformtemplate-bug-forked-fj2w66?file=%2Fsrc%2FApp.tsx%3A20%2C20)

**4. Steps to reproduce**

1. Open the sandbox. Notice that the fill attribute applied to the `<g>` element is applied and propagated to the rect, but that the `rect` does not move.
1. Open the console and notice that the motionValue is being updated, so that's not the problem.
2. Right click on the rect and open the inspector.  Look at the `<g>` element that is the parent of the `<rect>` element and notice that it's transform attribute is set to `[object Object]` instead of the value of the motionValue (`"transform(50 50)"`).

**5. Expected behavior**

`<motion.g>` elements should be able to use motionValues for all of their attributes, including the `transform` attribute.

**6. Video or screenshots**

See the sandbox above

**7. Environment details**

Chrome and Explorer for Mac

malo0n:
This pull request addresses a bug where SVG attributes like `transform` could incorrectly render as `[object Object]` when using MotionValues, and ensures that SVG-animated properties are handled correctly in the DOM. It introduces detection for SVGAnimated* properties, updates attribute handling logic, and adds tests to prevent regressions. The changes improve compatibility with SVG and prevent incorrect rendering of MotionValue objects.

fixes #3082 

**SVG attribute handling improvements:**

* Added a utility function `isSVGAnimatedProperty` to detect SVGAnimated* properties (e.g., `transform`), ensuring they are set via `setAttribute` instead of as properties, which prevents rendering issues.
* Updated `addSVGValue` and `canSetAsProperty` logic to use `isSVGAnimatedProperty`, ensuring SVG-animated properties are handled as attributes, not as DOM properties. [[1]](diffhunk://#diff-97504548e78ac28040def25e282974791b22278282dc26f512080374370523abR1) [[2]](diffhunk://#diff-966d84997f3857bb8a3d29493d20f15d52414984ccb122ad1cbef6ef53acbc24R6-R15) [[3]](diffhunk://#diff-97504548e78ac28040def25e282974791b22278282dc26f512080374370523abR53-R59)
* Improved the handling of `transform` in `buildSVGAttrs` to ensure user-provided SVG transform attributes override motion-synthesized transforms and prevent leaking MotionValue objects into the DOM.

**Testing and reliability:**

* Added tests to verify that MotionValues used as SVG attributes render correctly and do not appear as `[object Object]`, and that animated MotionValues update SVG attributes as expected.

**Minor fixes:**

* Improved regex in `convertAttrKey` for better Unicode handling.

**It's my first contribution in open-source project ever, so please let me know if smth wrong! I've tried to follow all recommendations**

### DIFF
--- a/dev/html/public/playwright/effects/svg-transform.html
+++ b/dev/html/public/playwright/effects/svg-transform.html
@@ -1,0 +1,30 @@
+<html>
+    <head>
+        <style>
+            body {
+                margin: 0;
+                font-family: sans-serif;
+            }
+        </style>
+    </head>
+    <body>
+        <svg width="200" height="200" viewBox="0 0 200 200">
+            <rect
+                id="rect"
+                x="50"
+                y="50"
+                width="100"
+                height="100"
+                fill="#0077ff"
+            />
+        </svg>
+        <script type="module" src="/src/inc.js"></script>
+        <script type="module">
+            const { motionValue, svgEffect } = window.Motion
+
+            // SVG-syntax transform (no units) — must go through setAttribute,
+            // not element.style.transform, so the value is preserved correctly.
+            svgEffect("#rect", { transform: motionValue("rotate(45)") })
+        </script>
+    </body>
+</html>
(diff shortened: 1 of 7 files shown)


mattgperry:
Thanks for the PR! Though handled this through https://github.com/motiondivision/motion/pull/3629

PR #3633, opened 2026-03-12
Add string offset to ViewTimeline

bakura10:
Hello,

This PR adds support for `start start`, `start end`, `end start`, `end end` string offsets to be compatible with ViewTimeline.

### DIFF
--- a/packages/framer-motion/src/render/dom/scroll/utils/__tests__/offset-to-range.test.ts
+++ b/packages/framer-motion/src/render/dom/scroll/utils/__tests__/offset-to-range.test.ts
@@ -56,9 +56,54 @@ describe("offsetToViewTimelineRange", () => {
         })
     })
 
-    it("returns undefined for string offsets", () => {
+    it("returns Enter preset for string offsets", () => {
         expect(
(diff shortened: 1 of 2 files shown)


mattgperry:
Thanks for the PR! Can you point me to an example that'd be simplified/made possible with this? I'm open to adding I'm just not 100% on what we'd get over current offsets.

bakura10:
Sorry @mattgperry I'm not sure to undesrtand. Right now, it seems that Motion will use native `ViewTimeline` (instead of registering a scroll listener) when you are using offset as number, eg. `offset: [[0, 1], [1, 0]]`

However in my case I used `start end`, `end start`... as it is mentally simpler to read (start of target meets end of container). This PR just allows to also use native ViewTimeline for people like me who were using the string name. I wanted to also add support for center but it is a bit more complicated, but with this PR I have about half of my scroll timelines that will use ViewTimeline now.

Or did I misunderstood the ViewTimeline benefits completely? :D

mattgperry:
Sorry you're right - I was confused about what this was doing. thanks again!

PR #3646, opened 2026-03-14
Add rangeStart/rangeEnd support for scroll()

### LINKED ISSUE #3001: [FEATURE] rangeStart and the rangeEnd support

### Issue: Need for `rangeStart`/`rangeEnd` Attributes Despite `offset`

The CSS attributes [`animation-range-end`](https://developer.mozilla.org/en-US/docs/Web/CSS/animation-range-end) and [`animation-range-start`](https://developer.mozilla.org/en-US/docs/Web/CSS/animation-range-start) have corresponding `rangeStart`/`rangeEnd` attributes in the [Web Animations API](https://developer.mozilla.org/en-US/docs/Web/API/Element/animate#rangeend).

#### Question:
Why are `rangeStart` and `rangeEnd` needed if we already have `offset`?  

The issue arises with entry animations.

---

#### Example with Native CSS:

In this example, a `:hover` animation works after the scroll-linked animation completes:

```css
@keyframes animate-in-range {
  from {
    opacity: 0;
  }
  to {
    opacity: 1;
  }
}

.box-right {
  animation: linear animate-in-range;
  animation-timeline: scroll();
  animation-range: 0% 20%;
}

.box-right:hover {
  opacity: 0.5; /* After 20% this works!!!! */
}

/* animated element has anim.effect.getComputedTiming().progress === null */
```

When using Motion Dev, the behavior is different:

```javascript
const animation = animate(
  box,
  {
    opacity: [0, 1],
  },
  {
  }
);

const scrollAnimation = scroll(animation, { offset: ['0%', '20%'] });

// animated element has anim.effect.getComputedTiming().progress === 1 even after 20%
```

With native CSS, the animation’s progress is null after the defined range. 
However, when using Motion Dev, the animation’s progress remains 1 even after the range (e.g., 20%). 
This creates a discrepancy in expected behavior, especially for entry animations like :hover.

Support for the corresponding `rangeStart`/`rangeEnd` attributes in the [Web Animations API](https://developer.mozilla.org/en-US/docs/Web/API/Element/animate#rangeend).

mattgperry:
## Summary

- Adds `rangeStart` and `rangeEnd` options to `scroll()`, forwarding them to the native WAAPI `Animation` object when using `ScrollTimeline`
- When user-provided range is set, `fill` is automatically changed to `"auto"` so the animation becomes inactive outside the range (matching native CSS/WAAPI behavior)
- Existing ViewTimeline offset-to-range mapping continues to use `fill: "both"` (no behavioral change)

## Problem

Native CSS and WAAPI support `animation-range-start`/`animation-range-end` (and the corresponding `rangeStart`/`rangeEnd` on `Element.animate()`). After the range ends, the animation becomes inactive (`progress === null`), allowing other CSS like `:hover` to take effect.

Motion's `scroll()` only supported `offset`, which clamps progress to 0/1 outside the range. This kept the animation's styles applied even after the range, preventing `:hover` and other CSS from overriding them.

## Fix

- Added `rangeStart`/`rangeEnd` to `ScrollOptions` type
- Added `fill` to `TimelineWithFallback` interface
- `attach-animation.ts` forwards user-provided range values with `fill: "auto"` to `attachTimeline()`
- `NativeAnimation.attachTimeline()` applies the `fill` override via `updateTiming()`

**Note:** This feature requires native `ScrollTimeline` support (Chrome 115+). In browsers without it, `rangeStart`/`rangeEnd` are silently ignored and the JS fallback path is used.

Fixes #3001

## Test plan

- [x] Cypress E2E test (`scroll-range.ts`) verifies animation is inactive after `rangeEnd` in Chrome (native ScrollTimeline)
- [x] Test adapts for Electron (no native ScrollTimeline) by checking `window.ScrollTimeline`
- [x] Passes React 18 and React 19
- [x] All 774 unit tests pass
- [x] Build succeeds

### DIFF
--- a/dev/react/src/tests/scroll-range.tsx
+++ b/dev/react/src/tests/scroll-range.tsx
@@ -1,0 +1,37 @@
+import { animate, scroll } from "framer-motion"
+import * as React from "react"
+import { useEffect } from "react"
+
+export const App = () => {
+    useEffect(() => {
+        const box = document.getElementById("box")!
+
+        const animation = animate(box, {
+            opacity: [0, 0.5],
+        })
+
+        const stop = scroll(animation, {
+            rangeStart: "0%",
+            rangeEnd: "20%",
+        })
+
+        return () => stop()
+    }, [])
+
+    return (
+        <>
+            <div style={{ height: "500vh" }} />
+            <div
+                id="box"
+                style={{
+                    position: "fixed",
+                    top: 0,
+                    left: 0,
+                    width: 100,
+                    height: 100,
+                    backgroundColor: "red",
+                }}
+            />
+        </>
+    )
+}
--- a/packages/framer-motion/cypress/integration/scroll-range.ts
+++ b/packages/framer-motion/cypress/integration/scroll-range.ts
@@ -1,0 +1,29 @@
+describe("scroll() rangeStart/rangeEnd", () => {
+    it("Animation is inactive after rangeEnd", () => {
+        cy.visit("?test=scroll-range")
+            .wait(100)
+            .viewport(100, 400)
+
+        // Scroll to ~50% (well past the 20% rangeEnd)
+        cy.scrollTo(0, 800)
+            .wait(200)
+            .get("#box")
+            .then(([$element]: any) => {
+                const opacity = parseFloat(
+                    getComputedStyle($element).opacity
+                )
+
+                if ((window as any).ScrollTimeline) {
+                    // With native ScrollTimeline + rangeStart/rangeEnd + fill: auto,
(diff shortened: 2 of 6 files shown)


bakura10:
Nice, this one would actually be better than the attempt to map Motion ranges to native one. Why this has been closed? :)

If we could both provide rangeStart, rangeEnd and offset (for the JS fallback) this would allows to have best of both worlds. I'm having some perf issue on Safari on pages with ton of scroll driven animations. I've done some manual tests with ViewTimeline and this effectively remove all the perf issues.

PR #3659, opened 2026-03-18
fix: revert string offset ViewTimeline acceleration

### LINKED ISSUE #3658: `useScroll` with `"start"/"end"` string offsets broken with `position: sticky` inside transformed parent since 12.37.0
## Describe the bug

Since `12.37.0` ("Support for hardware accelerating `"start"` and `"end"` offsets in `scroll` and `useScroll`"), `useScroll` with string offsets like `['start start', 'end end']` no longer works correctly when the target element contains a `position: sticky` child and is nested inside a parent `motion.div` with a `scale` transform.

The scroll progress starts at a non-zero value (content is already partially revealed on entry) and doesn't reach `1.0` before the section scrolls out of view (the reveal animation never completes).

## CodeSandbox reproduction

https://codesandbox.io/p/sandbox/5nflw4

## Steps to reproduce

1. Wrap a `useScroll` + `position: sticky` scroll-reveal component inside a `motion.div` that applies a `scale` transform
2. Use `offset: ['start start', 'end end']` on the inner `useScroll`
3. Scroll to the section
4. Observe: content is already partially revealed on entry, and the animation doesn't complete before scrolling past

## Expected behavior

`scrollYProgress` should range from `0` to `1` across the full scroll distance of the target element, regardless of CSS transforms on ancestor elements. This worked correctly in `12.36.0` and earlier.

## Environment details

- **Works:** `motion@12.36.0` / `framer-motion@12.36.0`
- **Breaks:** `motion@12.37.0` / `framer-motion@12.37.0` and later (tested up to `12.38.0`)
- **Breaking change:** "Support for hardware accelerating `"start"` and `"end"` offsets in `scroll` and `useScroll`" (12.37.0 changelog)
- **OS:** Linux
- **Browser:** Chrome (latest)
- **React:** 19.1.0
- **Next.js:** 16.2.0

PanagiotisKaraliolios:
## Problem

Since v12.37.0 (commit `6bae74e` — "Add string offset to ViewTimeline"), `useScroll` with string-based offsets such as `['start start', 'end end']` produces incorrect animation behaviour.

When a scroll-driven animation target uses `position: sticky` inside a `height: 200vh` container with a transformed parent, the ViewTimeline `"contain"` range does not produce identical results to the JS-based scroll tracking that was used before 12.37.0. Specifically:

- Some of the animation is already applied when the element first enters the viewport
- The animation does not reach completion when scrolling past

### Reproduction

See issue #3658 for a CodeSandbox reproduction and a video demonstrating the bug.

## Root Cause

The `parseStringOffset()` function added in `6bae74e` converts string offsets like `"start start"` to `ProgressIntersection` arrays, which enables `offsetToViewTimelineRange()` to return a ViewTimeline range (e.g. `"contain 0%" / "contain 100%"`). This causes `canAccelerateScroll()` in `use-scroll.ts` to return `true`, switching from JS main-thread scroll tracking to hardware-accelerated WAAPI animation.

The WAAPI ViewTimeline animation with `"contain"` range behaves differently from the JS-based scroll tracking for this DOM structure, producing the visual regression.

Notably, the existing Cypress E2E test in `scroll-view-timeline.ts` still asserts **"Does NOT accelerate target with string offset"**, which was never updated after 12.37.0 — confirming the acceleration of string offsets was not fully tested.

## Fix

This PR removes `parseStringOffset()` and the `stringToProgress` mapping from `offset-to-range.ts`, and makes `normaliseOffset()` return `undefined` for any non-array offset item. This restores the pre-12.37.0 behaviour where string offsets always fall back to JS main-thread scroll tracking.

Array-based `ProgressIntersection` offsets (the original preset format) continue to be accelerated as before — this change only affects string offsets.

### Changes

- **`offset-to-range.ts`**: Removed `parseStringOffset()`, `stringToProgress`, and string handling in `normaliseOffset()` (−20 lines)
- **`offset-to-range.test.ts`**: Updated string offset tests to expect `undefined` instead of range objects (+5/−58 lines)

All 37 existing unit tests pass (offset-to-range, use-scroll, use-transform).

Fixes #3658

### DIFF
--- a/packages/framer-motion/src/render/dom/scroll/utils/__tests__/offset-to-range.test.ts
+++ b/packages/framer-motion/src/render/dom/scroll/utils/__tests__/offset-to-range.test.ts
@@ -56,52 +56,19 @@ describe("offsetToViewTimelineRange", () => {
         })
     })
 
-    it("returns Enter preset for string offsets", () => {
+    it("returns undefined for string offsets (not accelerated)", () => {
         expect(
             offsetToViewTimelineRange(["start end", "end end"])
-        ).toEqual({
-            rangeStart: "entry 0%",
-            rangeEnd: "entry 100%",
-        })
-    })
-
-    it("returns Exit preset for string offsets", () => {
+        ).toBeUndefined()
         expect(
             offsetToViewTimelineRange(["start start", "end start"])
-        ).toEqual({
-            rangeStart: "exit 0%",
-            rangeEnd: "exit 100%",
-        })
-    })
-
-    it("returns Any preset for string offsets", () => {
+        ).toBeUndefined()
         expect(
             offsetToViewTimelineRange(["end start", "start end"])
-        ).toEqual({
-            rangeStart: "cover 0%",
-            rangeEnd: "cover 100%",
-        })
-    })
-
-    it("returns All preset for string offsets", () => {
+        ).toBeUndefined()
         expect(
             offsetToViewTimelineRange(["start start", "end end"])
-        ).toEqual({
-            rangeStart: "contain 0%",
-            rangeEnd: "contain 100%",
-        })
-    })
-
-    it("returns All preset for string offsets", () => {
-        expect(
-            offsetToViewTimelineRange(["start start", "end end"])
-        ).toEqual({
-            rangeStart: "contain 0%",
-            rangeEnd: "contain 100%",
-        })
-    })
-
-    it("returns undefined for other string offsets", () => {
+        ).toBeUndefined()
         expect(
             offsetToViewTimelineRange(["start center", "end start"])
         ).toBeUndefined()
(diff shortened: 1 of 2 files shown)


PanagiotisKaraliolios:
The CI failure in `drag-nested.ts` ("Nested drag with constraints and animation -- Child layout") is **unrelated to this PR**.

This PR only modifies `offset-to-range.ts` and its unit test — scroll offset utilities with zero overlap with the drag/constraints/layout system.

The failing Cypress test is timing-sensitive: it drags an element past constraints, releases, waits a hardcoded 2 seconds for the spring animation to settle, then asserts exact pixel positions. This is a known flaky pattern on slower CI runners.

mattgperry:
Just closing this out as I'd prefer an actual fix - this doesn't address `[0,0],[1,1]` which would still essentially be broken.

PR #3662, opened 2026-03-23
Fix AnimatePresence: apply object-form initial on re-entry

lennondotw:
## Summary

When a child re-enters `AnimatePresence` after its exit animation completed, object-form `initial` values (e.g., `initial={{ opacity: 0.5 }}`) are not applied. The component animates from the exit end value instead of jumping to the initial value first.

## Root cause

The re-entry logic added in 6a8d3abb9 and simplified in 3da9d0983 only handles string variant labels:

```typescript
if (typeof initial === "string") {
```

Object-form initial values are skipped, so `resolveVariant` is never called for them.

## Fix

Extend the condition to also handle object-form initial values:

```typescript
if (
    typeof initial === "string" ||
    (typeof initial === "object" &&
        initial !== null &&
        !Array.isArray(initial))
)
```

`resolveVariant` already supports both string and object forms — it returns objects as-is. The `null` and `Array.isArray` guards handle the JS `typeof null === "object"` quirk and array variant labels respectively.

## Test plan

- [x] Added unit test: "Re-entering child with object-form initial resets to initial values when exit was complete"
- [x] Updated CHANGELOG.md


lennondotw:
Hi @mattgperry, friendly ping — this is a small one-line fix for object-form `initial` not being applied on AnimatePresence re-entry. CI is all green. Would appreciate a look when you have a moment!

lennondotw:
Rebased on latest `main` (v12.38.0). The original PR was based on older code before 6a8d3abb9 and 3da9d0983 refactored the re-entry logic — those two commits fixed the string variant case and introduced `resolveVariant`, but the `typeof initial === "string"` guard still skips object-form initial values.

This PR now applies cleanly on top of the current code — just extends the type guard on line 28 of `exit.ts` to also handle `typeof initial === "object"` (with `null` and array guards). `resolveVariant` already returns objects as-is, so no other changes needed.

@mattgperry would appreciate a look when you get a chance — it's a small extension to the fix you landed in 6a8d3abb9.

mattgperry:
Thanks for the PR!

PR #3680, opened 2026-04-16
fix: useAnimate respects MotionConfig.skipAnimations

### LINKED ISSUE #3679: useAnimate does not respect MotionConfig.skipAnimations
## Description

The imperative `useAnimate` hook does not respect `MotionConfig`'s `skipAnimations` prop. It only checks `reducedMotion` via `useReducedMotionConfig()`, while declarative animations (`initial`/`animate`/`exit`) respect both `skipAnimations` and `reducedMotion` through the `VisualElement` class.

This means that even with `<MotionConfig skipAnimations={true}>`, `useAnimate` still creates WAAPI animations — they may run with reduced timing, but they still register as running animations in the browser.

## Reproduction

```tsx
<MotionConfig skipAnimations={true}>
  <MyComponent />
</MotionConfig>

function MyComponent() {
  const [scope, animate] = useAnimate();

  useEffect(() => {
    // This animation still fires via WAAPI despite skipAnimations={true}
    animate(scope.current, { opacity: [0, 1] }, { duration: 0.3 });
  }, []);

  return <div ref={scope}>Hello</div>;
}
```

## Impact

This is particularly problematic for e2e testing with Playwright. Playwright's `toHaveScreenshot()` calls `element.getAnimations()` to check element stability before capturing. WebKit reports the zero-duration WAAPI animations created by `useAnimate` as still running, causing the stability check to time out indefinitely — even though `skipAnimations` is true and the app has explicitly opted out of all animations.

Declarative animations don't have this problem because they go through `VisualElement`, which checks `this.skipAnimationsConfig`.

## Expected behaviour

`useAnimate`'s returned `animate` function should be a no-op (or resolve immediately without creating WAAPI animations) when `MotionConfig.skipAnimations` is `true`, consistent with how declarative animations behave.

## Workaround

We created a `useSafeAnimate` wrapper that reads `skipAnimations` from `MotionConfigContext` and returns a no-op animate function when it's true:

```tsx
export function useSafeAnimate<T extends Element = HTMLDivElement>() {
  const [scope, animate] = useAnimate<T>();
  const { skipAnimations } = useContext(MotionConfigContext);
  const shouldReduceMotion = useReducedMotionConfig();
  const shouldSkip = !!skipAnimations || !!shouldReduceMotion;

  // return no-op animate when shouldSkip is true
  // ...
}
```

## Relevant source

In [`use-animate.ts`](https://github.com/motiondivision/motion/blob/e84202434978a0648ff9728f817f10e887edcb85/packages/framer-motion/src/animation/hooks/use-animate.ts#L10), the hook calls `useReducedMotionConfig()` but does not read `skipAnimations` from `MotionConfigContext`.

jthrilly:
## Summary

`useAnimate` only checked `reducedMotion` via `useReducedMotionConfig()`, ignoring `MotionConfig`'s `skipAnimations` prop. This meant WAAPI animations were still created even when `skipAnimations` was `true`.

When `skipAnimations` is `true`, the returned `animate` function now resolves immediately via an empty `GroupAnimationWithThen` without creating any WAAPI animations — consistent with how declarative animations behave.

## Motivation

This is problematic for e2e testing tools like Playwright, which call `element.getAnimations()` to check element stability before taking screenshots. WebKit reports zero-duration WAAPI animations as running, causing stability checks to time out indefinitely — even though the app has explicitly opted out of animations via `<MotionConfig skipAnimations>`.

The declarative animation path (`initial`/`animate`/`exit` props) correctly respects `skipAnimations` via `VisualElement.shouldSkipAnimations`, but the imperative `useAnimate` path was missing this check.

## Changes

- `use-animate.ts`: Read `skipAnimations` from `MotionConfigContext`. When `true`, return a no-op animate function that creates no WAAPI animations.
- Added test: verify `useAnimate` skips animations when wrapped in `<MotionConfig skipAnimations>`.

Fixes #3679

### DIFF
--- a/packages/framer-motion/src/animation/hooks/__tests__/use-animate.test.tsx
+++ b/packages/framer-motion/src/animation/hooks/__tests__/use-animate.test.tsx
@@ -1,6 +1,7 @@
 import "@testing-library/jest-dom"
 import { render } from "@testing-library/react"
 import { useEffect } from "react"
+import { MotionConfig } from "../../../components/MotionConfig"
 import { useAnimate } from "../use-animate"
 
 describe("useAnimate", () => {
@@ -114,5 +115,40 @@ describe("useAnimate", () => {
         })
 
         expect(frameCount).toEqual(3)
+    })
+
+    test("Skips animations when MotionConfig skipAnimations is true", () => {
+        return new Promise<void>((resolve) => {
+            const Component = () => {
+                const [scope, animate] = useAnimate()
+
+                useEffect(() => {
+                    const animation = animate(
+                        scope.current,
+                        { opacity: 0.5 },
+                        { duration: 1 }
+                    )
+
+                    // Animation should not be tracked in scope
+                    expect(scope.animations.length).toBe(0)
+
+                    animation.then(() => {
+                        // Element style should not be changed
+                        expect(scope.current).not.toHaveStyle(
+                            "opacity: 0.5;"
+                        )
+                        resolve()
+                    })
+                })
+
+                return <div ref={scope} />
+            }
+
+            render(
+                <MotionConfig skipAnimations>
+                    <Component />
+                </MotionConfig>
+            )
+        })
     })
 })
--- a/packages/framer-motion/src/animation/hooks/use-animate.ts
+++ b/packages/framer-motion/src/animation/hooks/use-animate.ts
@@ -1,10 +1,11 @@
 "use client"
 
-import { useMemo } from "react"
-import { AnimationScope } from "motion-dom"
+import { useContext, useMemo } from "react"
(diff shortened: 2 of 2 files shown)


jthrilly:
Yeah, I see that this should probably transition immediately to any end state rather than discarding animation entirely.

mattgperry:
Superseded by #3683, which lands the same fix but applies the final keyframe value instantly (matching the contract of `MotionGlobalConfig.skipAnimations` and declarative `<MotionConfig skipAnimations>`) instead of discarding target values via a no-op animate. Closing this PR — #3679 will be closed by the new PR's merge. Thanks for the report and the diagnosis @jthrilly.

PR #3757, opened 2026-06-12
Fix NaN in spring-animated SVG polygon points (#2791)

### LINKED ISSUE #2791: [BUG] NaN in point animation when using spring animations
**Describe the bug**

When animating a `<polygon>` point list with a "spring" animation, at various points in the animation, the computed pointlist contains NaNs and throws console errors.

image

https://codesandbox.io/p/sandbox/pensive-khayyam-5mwd29?file=%2Fsrc%2FApp.js%3A9%2C68&workspaceId=5c31fde4-86cd-4353-99cd-cbe570ccc4aa

**Steps to reproduce**

1. Define an SVG with a polygon element.
2. Animate the points of the element, and use a spring animation.
3. Observe that you get a bunch of complaints about NaNs in your error console.

When the animation type is not "spring", the NaN errors are not thrown (though one "expected number, 'undefined'" message is thrown per polygon element).

**Expected behavior**

When an SVGPointList would produce NaNs, it should not incorrectly attempt to set the pointlist with NaNs on the element.

**Environment details**

Reproduced in Brave v 1.69.153 (Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/128.0.0.0 Safari/537.36)

mattgperry:
## Bug

Animating an SVG `<polygon>`'s `points` with `type: "spring"` intermittently writes an invalid point list and floods the console:

```
<polygon> attribute points: Expected number, "NaN,NaN NaN,NaN …".
```

(With non-spring types the NaNs don't appear.) Reported in #2791.

## Cause

The all-`NaN` point list (with an otherwise correct comma/space template) can only occur if the spring's progress output is `NaN`. Tracing the spring generator:

- For non-numeric keyframes (a `points` string), `JSAnimation` animates a `0 → 100` progress spring and maps it back through the complex-value mixer. The **mixer is not at fault** — it produces finite output even with spring overshoot (progress `> 1` / `< 0`) and mismatched point counts.
- `getSpringOptions()` resolves physics parameters with `{ ...defaults, ...options }`. An explicit **`stiffness: undefined`** — e.g. an optional prop forwarded into `transition` — *clobbers* the default, and `0`/negative values pass straight through. Those values then divide and feed `Math.sqrt()` during resolution (`dampingRatio = damping / (2 * Math.sqrt(stiffness * mass))`, `undampedAngularFreq = sqrt(stiffness / mass)`), yielding `NaN` spring values for every frame.

A `NaN` *velocity*, by contrast, is harmless — it's sanitised by `velocity || 0`.

## Fix

Guard `stiffness` and `mass` to a positive, finite value in `getSpringOptions()`, falling back to the defaults otherwise. The check runs once at resolution time, not in the per-frame hot path.

```ts
if (!(springOptions.stiffness > 0)) springOptions.stiffness = springDefaults.stiffness
if (!(springOptions.mass > 0)) springOptions.mass = springDefaults.mass
```

This catches `0`, negative, `NaN`, and `undefined` while leaving valid values untouched.

## Tests

Failing-first (verified to fail on the unfixed code, pass after the fix):

- `packages/motion-dom/.../generators/__tests__/spring.test.ts` — `spring()` with `stiffness`/`mass` of `0` or explicit `undefined` no longer returns `NaN`.
- `packages/motion-dom/.../__tests__/JSAnimation.test.ts` — a spring over a polygon `points` string never emits `"NaN"`. Before the fix this produced exactly `"NaN,NaN NaN,NaN NaN,NaN"`, matching the screenshot in the issue.

The unit tests reproduce the precise symptom and are the regression gate; the root cause is environment-independent spring math, so no Cypress E2E is required. Full `motion-dom` (476) and `framer-motion` client (799) suites pass.

## Uncertainty

The original CodeSandbox reproduction is no longer reachable, so the exact `transition` the reporter used couldn't be confirmed. The fix targets the only input class that reproduces the reported `"NaN,NaN …"` output — a non-positive/undefined spring physics parameter — and the precise symptom is reproduced and gated by the new tests. The separate `Expected number, "undefined"` message noted in the issue is a distinct, smaller symptom (an unresolved origin keyframe rendering as `undefined`) and is intentionally left out of scope to keep this fix minimal.

Fixes #2791

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ Undocumented APIs should be considered internal and may change without warning.
 
 -   `AnimatePresence`: Exiting children no longer interleave with entering children, which could reorder and remount children present in both renders.
 -   `motion`: Throw error when passing a custom `motion` component an incorrect `ref` type.
+-   `spring`: Invalid `stiffness`, `damping` or `mass` (`0`, negative, non-finite, or an explicit `undefined` from a forwarded prop) no longer resolve to `NaN` animation values, which corrupted values like an SVG `polygon`'s `points` and left the animation running forever.
 
 ## [12.42.2] 2026-07-01
 
--- a/packages/motion-dom/src/animation/__tests__/JSAnimation.test.ts
+++ b/packages/motion-dom/src/animation/__tests__/JSAnimation.test.ts
@@ -1422,4 +1422,22 @@ describe("JSAnimation", () => {
         expect(animation.sample(1000).value).toBe("90%")
         expect(animation.sample(1999).value).toBe("90%")
     })
+
+    // https://github.com/motiondivision/motion/issues/2791
+    test("Spring over SVG polygon points never produces NaN", () => {
+        // An invalid spring physics value (here an explicit `undefined`
+        // stiffness, as forwarded from an optional prop) previously emitted
+        // NaN progress, which the complex-value mixer turned into an invalid
+        // "NaN,NaN NaN,NaN" point list written to the <polygon> element.
+        const animation = animateValue({
+            keyframes: ["150,5 75,200 225,200", "150,5 50,180 250,180"],
+            type: "spring",
+            stiffness: undefined,
+            autoplay: false,
+        })
+
+        for (let t = 0; t <= 2000; t += 50) {
+            expect(animation.sample(t).value).not.toContain("NaN")
+        }
+    })
 })
--- a/packages/motion-dom/src/animation/__tests__/JSAnimation.test.ts
+++ b/packages/motion-dom/src/animation/__tests__/JSAnimation.test.ts
@@ -1429,15 +1429,24 @@ describe("JSAnimation", () => {
         // stiffness, as forwarded from an optional prop) previously emitted
         // NaN progress, which the complex-value mixer turned into an invalid
         // "NaN,NaN NaN,NaN" point list written to the <polygon> element.
+        const target = "150,5 50,180 250,180"
         const animation = animateValue({
-            keyframes: ["150,5 75,200 225,200", "150,5 50,180 250,180"],
+            keyframes: ["150,5 75,200 225,200", target],
             type: "spring",
             stiffness: undefined,
             autoplay: false,
         })
 
         for (let t = 0; t <= 2000; t += 50) {
-            expect(animation.sample(t).value).not.toContain("NaN")
+            const coords = animation.sample(t).value.match(/-?[\d.]+/g)!
+            // Every coordinate must be a finite number
+            expect(
+                coords.every((coord) => Number.isFinite(Number(coord)))
+            ).toBe(true)
         }
+
+        // ...and the spring must still settle on the target, so this doesn't
+        // pass for any non-NaN corruption of the point list
+        expect(animation.sample(2000).value).toBe(target)
     })
 })
--- a/packages/motion-dom/src/animation/generators/__tests__/spring.test.ts
+++ b/packages/motion-dom/src/animation/generators/__tests__/spring.test.ts
@@ -285,3 +285,36 @@ describe("toString", () => {
         )
     })
 })
+
+// https://github.com/motiondivision/motion/issues/2791
+describe("spring NaN guards", () => {
+    const sample = (options: ValueAnimationOptions<number>) => {
+        const generator = spring(options)
+        return [0, 100, 300, 600, 1000].map((t) => generator.next(t).value)
+    }
+
+    test("stiffness of 0 does not produce NaN", () => {
+        const values = sample({ keyframes: [0, 100], stiffness: 0 })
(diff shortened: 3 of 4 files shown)


mattgperry:
Addressed code review feedback and rebased onto `main` (was 98 commits behind).

### Fixed

The original guard was incomplete — it ran too late and covered the wrong set of values:

- **`damping` was omitted.** `spring({ keyframes: [0, 100], damping: undefined })` still returned `[NaN, NaN, NaN, NaN, NaN]`, i.e. the reported bug reproduced in full through a third prop. It also never sets `done`, so the frameloop spins indefinitely.
- **`!(x > 0)` let `Infinity` through**, despite the comment claiming to guard non-finite values. `stiffness: Infinity` still resolved to NaN. Now uses `Number.isFinite`.
- **The guard ran after the physics-vs-duration branch decision.** `isSpringType` tests `!== undefined`, so `stiffness: 0` counted as "physics specified" and silently discarded a provided `duration` — `{ duration: 500, stiffness: 0 }` produced output byte-identical to a bare default spring. Physics are now resolved *before* the branch decision.
- **Stiffness was replaced without its paired damping.** Duration resolution derives `damping` from `stiffness`; overwriting one alone yields `dampingRatio: 0`, an undamped spring. The degenerate cases are now replaced as a pair.
- **`if (options.visualDuration)`** → `!== undefined`.
- **Invalid physics now warns** (`spring-invalid-physics`) instead of silently substituting defaults.

`damping: 0` is deliberately still honoured — it's a valid, perpetually oscillating spring, and only `stiffness`/`mass` are divisors.

### Tests

The NaN-guard suite now tables all three physics keys against every way each can be invalid (`0`, negative, `NaN`, `±Infinity`, explicit `undefined`), plus the duration-not-discarded cases and the warning. The `JSAnimation` polygon test asserted `.not.toContain("NaN")`, which passes for any non-NaN corruption; it now checks every coordinate is finite *and* that the spring settles on the target.

Verified these are real gates: reverting `spring.ts` to the previous commit fails 9 tests (the `damping`, `Infinity`, discarded-`duration` and warning cases); reverting to `main` fails 21.

### Other findings, not changed

- **`restSpeed ||=` / `restDelta ||=`** were flagged as the same falsy-zero bug class. They aren't — springs approach their target asymptotically, so `restSpeed: 0` means the spring can only rest once the exponential underflows to exactly zero, which is ~149s for a default spring and would just hit the generator duration cap. The `||=` is protective here, so I left it.
- **`inertia`'s `bounceStiffness`** now falls back to spring's default of 100 rather than inertia's 500 when explicitly passed `0`. Honouring inertia's default would mean inventing a policy for a value the caller explicitly set to zero, so I left the invalid-input path pointing at spring's own default.

### Checks

- `motion-dom`: 517 passed
- `framer-motion`: 802 passed
- Bundle size: +~130 B gzip. The size gates that currently fail (`size-rollup-motion`, `-m`, `-animate`, `-scroll`, `-waapi-animate`, `style-effect`) fail identically on `main` at the same numbers — pre-existing, not from this change. The two gates with headroom still pass.

MILLERMARRU (commented):
The spring physics logic itself is solid. Tracing through `resolvePhysics`, the object-spread ordering was the real bug: the old code spread `...options` after the property defaults, and in JS an explicitly present `stiffness: undefined` key (which is exactly what happens when a component forwards an optional prop that wasn't set) survives a spread and overwrites the default rather than falling through to it. Moving the physics resolution to compute `stiffness ?? springDefaults.stiffness` etc. after the spread, and doing it for all three physics keys together before deciding between physics- and duration-based resolution, closes both the original NaN report and the more subtle case the tests call out, where `stiffness: 0` used to be treated as "physics was specified" and silently discarded a valid `duration`/`visualDuration`.

One thing that looks like it'll break the build though: `spring.ts` imports `warning` from `motion-utils`, but that package's public entry only re-exports `warn-once.ts`, which exports `warnOnce` and `hasWarned`, no `warning`. I checked the full file listing under `packages/motion-utils/src` and the git history on `warn-once.ts`, there's never been a plain `warning` export there. This PR also doesn't touch anything under `motion-utils` per the file list, so as it stands `import { warning } from "motion-utils"` has nothing to resolve to.

It also looks like `warnOnce`'s actual behavior wouldn't give the test at the bottom (`invalid physics warns rather than failing silently`) what it expects anyway, even if the name were swapped in as a quick fix. `warnOnce` dedupes by the exact message string process-wide via a module-level `Set`, and every `resolvePhysics` call in this file uses the same message text regardless of which key or value triggered it. The parameterized "does not produce NaN" tests earlier in the same file already exercise `stiffness: 0`-equivalent invalid values multiple times before that later test runs, so under real `warnOnce` semantics the message would already be marked as warned by the time that assertion runs, and `console.warn` wouldn't fire a second time for it. That reads like the intent here is a warning that fires on every occurrence, which isn't what either existing export in `motion-utils` does today.

PR #3761, opened 2026-06-17
Resolve selector/Element targets in animateView()

mattgperry:
## What

`animateView()` (the View Transitions integration in `motion-dom`) only ever animated the `root` layer and pre-named string layers. Targets other than `"root"` were never resolved to elements or assigned a `view-transition-name`, so selector/`Element` targets silently did nothing — the implementation carried three TODOs describing exactly this gap, with no E2E coverage.

This finishes non-root target resolution, adds the automatic `view-transition-name` management the docs already promise, and makes cross-aspect-ratio morphs look right out of the box.

## API

- **`.add(selector | Element)`** — the target API. Resolves matching elements and assigns each a script-visible `view-transition-name`, in two passes (before the update → old snapshot; after → new snapshot) via a per-transition `Map<Element, string>`, so a persistent element keeps its name and morphs as a single `group`. An element that **already** has an author-defined name is reused as-is (not auto-renamed); `auto`/`match-element` are overridden because their generated name isn't exposed to script (so it can't be targeted for WAAPI animations). A bare `.add()` **auto-enables** the layout/morph — no explicit `.layout()` needed.
- **`.add(from, to)`** — pairs two *different* elements into one shared-element morph: `from` is resolved in the old snapshot, `to` in the new, and both are forced onto one generated name so the browser tweens one into the other. **Symmetric** — reverse the args to morph back (open: `.add(card, ".modal")`, close: `.add(".modal", card)`). Crop applies across the pair (old corners → new), so a card → modal morph is rounded automatically with **no manual `view-transition-name` anywhere**.
- **`.crop()`** — **on by default**: clips the morph (`overflow: clip`) + `object-fit: cover` so cross-aspect-ratio morphs don't distort or overflow (the UA default does), and animates per-corner `border-radius` from the old element's measured radii to the new element's. `.crop(false)` opts a subject out.
- **`.layout(transition?)`** — sets the morph's transition (timing / `stagger` / `delay`). The morph is auto-enabled by `.add()`, so this is *only* a transition setter — no keyframes. On the implicit `root` subject it doubles as the opt-in for the whole-page crossfade (how the Netflix demo fades its backdrop/grid). `.enter(keyframes, transition?)` / `.exit(keyframes, transition?)` keep their keyframes — the new/old layer animations.
- **Per-element stagger** — a `delay` function resolves per resolved element across a selector.
- **Inferred enter/exit from-values** — `.exit({ opacity: 0 })` animates from an inferred `1` rather than applying instantly (falsy targets like `0` are no longer skipped).
- Removed `.get()`/`.new()`/`.old()` — not part of the documented surface.

Generated names are removed in `transition.finished.finally`. The three TODOs in `start.ts` are gone.

## Why not just rely on the platform?

Audited [CSS View Transitions L1](https://www.w3.org/TR/css-view-transitions-1/) and [L2](https://www.w3.org/TR/css-view-transitions-2/): neither offers a JS path for this, and `view-transition-name: auto`/`match-element` deliberately hides the generated name from script — which is exactly the handle our WAAPI engine needs to attach per-layer animations and read them back. So Motion must assign names it controls.

## Testing

- **Unit** (`assign-names.test.ts`): unique-name generation, same-element reuse, forced names (the new end of a pair), author-name skip, `auto`/`match-element` override, cleanup, and the no-`startViewTransition` fallback with an `Element` target. Full `motion-dom` suite green (482 passed).
- **E2E** (`tests/view/view-targets.spec.ts`, 15 specs × **Chromium + WebKit**): `Element` target animates its `new` layer; a selector matching 3 elements → 3 distinct named layers; an already-named element keeps its author name (not auto-renamed); a bare `.add()` auto-enables a group morph; a paired `.add(from, to)` morphs two different elements as one layer in **both directions** (old element hidden, and old element removed); per-element stagger (incl. default-option staggering of a browser-generated morph, and indexing from the new snapshot when `update()` replaces the nodes); overlapping `.add()` subjects on one element keep both buckets (`.a` enter + `.b` exit); default crop (clip + `object-fit` + animated radius); individual corner radii; `.crop(false)` opt-out; `.crossfade()`; and `.exit({ opacity: 0 })` animating from an inferred `1`.
- Visual demos in `dev/html/public/examples/` — the **Netflix card→modal** morph (paired `.add()`, manually verified open + close + a no-console-error smoke run), circular clip-path enter/leave, crop cover-vs-overflow explainer, stagger + auto-layout.

## Notes / follow-ups

- **Element-scoped capture (`animateView(element)` + `element.startViewTransition`) is deferred to a follow-up.** It's spec-aligned (Chrome/Edge 147+), but Playwright's pinned Chromium predates 147, so the subtree-scoped path can't be E2E-verified in CI and had accrued scoped-only morph/teardown bugs that need a real Chrome 147 dev loop. This PR is the fully-verified document-level surface.
- **Keyframes aren't a positional `.add()` argument** — its second argument is a *paired target* (above), not keyframes; keyframes stay on `.enter()/.exit()` (and `.layout()` takes only a transition).
- `dev/html/src/imports/view.js` is a small motion-dom-only fixture shim (the shared `inc.js` also imports `framer-motion`, which isn't needed here).


mattgperry:
Thanks for the review — addressed in `a3ce27a` and `e5e31104`:

**P1 — `match-element` not excluded (real bug, fixed).** Correct catch — `view-transition-name: match-element` generates a name the browser keeps internal, exactly like `auto`, so treating it as an author name produced a dead `::view-transition-new(match-element)` selector. Added it to the override condition and parametrized the unit test to cover both `auto` and `match-element`.

**P2 — fallback / `notifyReady` wiring (verified, no change needed).** Yes, the queue pipes it back: `queue.ts` `start()` does `startViewAnimation(builder).then((animation) => builder.notifyReady(animation))`, and the fallback resolves a `GroupAnimation([])` through that same path. The fallback test `await`s the builder — which can only resolve if `notifyReady` fires — and now also asserts the resolved value is defined, so a silent stall would fail the test.

**P2 — module-level `nameCount` (keeping, by design).** The global monotonic counter is intentional: names must be unique **across concurrent transitions** (the queue and element-scoped `startViewTransition` can overlap), and a per-transition counter reset to `0` would collide between two simultaneous transitions. It's a single integer (no memory growth), and the tests assert the `motion-view-<n>` shape + uniqueness via a `Set` rather than exact numbers, so there's no determinism issue.

Separately, this branch also migrates `turbo.json` from the pre-2.0 `pipeline` key to `tasks` — the repo pins turbo 2.9.14, so `turbo run` (and thus `yarn build` in the CircleCI `setup` job) was erroring on the old schema.

PR #3763, opened 2026-06-22
Enable WAAPI acceleration for backgroundColor and color

mattgperry:
Implements **plan 003** (`plans/003-color-waapi-acceleration.md`).

## Root cause
Color animations (`backgroundColor`, `color`, …) were deliberately excluded from WAAPI acceleration — a code comment deferred them "until we implement support for `linear()` easing". That blocker is resolved: Motion already generates `linear()` easing strings for unsupported easings/springs and wires them into the WAAPI path (`map-easing.ts:14-16`, `spring.ts:433`). As a result, plain color animations fell to the per-frame JS path, re-rendering the element's styles every frame on the main thread.

## Approach
Add `"backgroundColor"` and `"color"` (camelCase Motion value names) to `acceleratedValues`, replacing the dead dash-case `// "background-color"` comment (which never would have matched the camelCase name anyway). `supports/waapi.ts` needed no change — the existing `acceleratedValues.has(name)` branch covers it.

Honest framing: color is **paint-bound**, so this is a main-thread *offload*, not free compositor rendering. Chrome additionally composites solid-color `background-color` animations on top of that.

Scope kept minimal:
- The browser-only-color path (`oklch`/`lab`/`lch`/`color-mix`) is **unchanged** — that branch is correctness (the JS path can't parse those formats), not perf.
- Remaining colors (borderColor family, `fill`/`stroke`) intentionally left out; SVG is already excluded by the `HTMLElement` check.

## Tests
- **Unit** (the real regression gate, per CLAUDE.md): extended `supports/__tests__/waapi.test.ts` — standard-keyframe `backgroundColor`/`color` are now eligible; `onUpdate` kill-switch still applies; a color not in the set (`borderTopColor`) is ineligible with standard keyframes but eligible with oklch. The two new positive assertions were watched failing first.
- **E2E (Playwright, real Chromium)**: `animate-color-accelerated.html` asserts `getAnimations().length > 0` (it's now a WAAPI animation) and a strictly-between mid-animation colour; `animate-color-interrupt.html` asserts an interrupt lands on the new target with no jump and no console errors.

## Gates run
- `yarn build` ✅
- motion-dom jest — 475 pass ✅
- framer-motion `test-client` — 799 pass ✅
- Playwright animate suite (chromium) — 58 pass ✅
- Cypress color specs (`oklch-color-animation`, `animate-style`) on React 18 **and** 19 — 10/10 each ✅
- Spring-on-color STOP-condition check (manual throwaway fixture): valid colours mid-animation, no NaN, settles on target ✅
- `yarn lint`: fails only in pre-existing `dev/react` fixtures (zero diff vs `main`, untouched here); motion-dom is unlinted by `yarn lint`.

## Deviations from the plan
None. No files outside the in-scope list were modified.

## Notes for reviewers
- No change to the `hasBrowserOnlyColors` branch semantics.
- Safari de-accelerates `linear()` easing, so color animations with spring/custom easings run WAAPI-on-main-thread there — parity with today, worth a release-notes line.


mattgperry:
@copilot resolve the merge conflicts in this pull request

PR #3781, opened 2026-08-05
Fix SVG zero-duration opacity animations

### LINKED ISSUE #3779: [BUG] Animating SVG elements may leave inline styles and SVG attributes out-of-sync (regression in 12.43.0)
## Describe the bug

Since 12.43.0 where hardware acceleration for SVG elements was added, the SVG attributes and inline styles are not always in sync, and animating with duration 0 changes only the SVG attributes, but not inline styles.

## Reproduction

CodeSandbox: https://codesandbox.io/p/sandbox/motion-svg-duration-0-bug-qkst3g

1. Click toggle
2. Click toggle again

Expected: the button fades out, then reappears immediately.
Actual: the button fades out and never comes back.

What happens in the DOM, is that:

0. Initial state:  `<foreignObject opacity="1">`
1. Click toggle (non-zero duration): `<foreignObject opacity="0" style="opacity: 0;">`
2. Click toggle again (zero duration): `<foreignObject opacity="1" style="opacity: 0;">`

I.e., the `opacity` attribute is changed back to 1, but the inline style remains at opacity 0.

Downgrading to motion 12.42.x works.

## Workaround

My current workaround is to specify a minimal non-zero duration 0.001 instead of 0.

mattgperry:
## Summary
- render SVG opacity through CSS so WAAPI completion and instant JS updates write to the same channel
- add a Cypress regression based on the issue's `foreignObject` reproduction
- update SVG WAAPI assertions to reflect opacity's CSS rendering

Fixes #3779

## Test plan
- [x] `yarn build`
- [x] `yarn test` (all 7 tasks; 96 Framer Motion suites and 809 tests)
- [x] reproduced the stale `style.opacity = 0` before the fix in Chromium
- [x] verified fade-out and zero-duration restoration in Chromium with React 18 and React 19
- [x] targeted ESLint (no errors) and Prettier checks
- [ ] Cypress CLI: repository-pinned Cypress 4 Electron cannot start on macOS 26 (`bad option: --no-sandbox` / `--smoke-test`)
- [ ] repository-wide `yarn lint`: blocked by 52 pre-existing errors in unrelated dev examples/tests

### DIFF
--- a/dev/react/src/tests/waapi-svg-zero-duration.tsx
+++ b/dev/react/src/tests/waapi-svg-zero-duration.tsx
@@ -26,6 +26,24 @@ export const App = () => {
                 >
                     <div style={{ background: "red" }}>chip</div>
                 </motion.foreignObject>
+                <motion.rect
+                    id="transform-target"
+                    animate={{
+                        transform: hidden
+                            ? "translateX(50px)"
+                            : "translateX(0px)",
+                    }}
+                    height={40}
+                    initial={false}
+                    transition={
+                        hidden
+                            ? { duration: 0.3 }
+                            : { duration: 0.3, transform: { duration: 0 } }
+                    }
+                    width={40}
+                    x={120}
+                    y={10}
+                />
             </svg>
         </>
     )
--- a/dev/react/src/tests/waapi-svg-zero-duration.tsx
+++ b/dev/react/src/tests/waapi-svg-zero-duration.tsx
@@ -1,0 +1,32 @@
+import { motion } from "framer-motion"
+import { useState } from "react"
+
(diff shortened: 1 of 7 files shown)


mattgperry:
Follow-up verification correction: the Cypress launch failure was caused by `ELECTRON_RUN_AS_NODE` in the local Cursor environment, not by macOS/Cypress incompatibility. Running with that variable unset succeeds:

- React 18: `waapi-svg-zero-duration.ts` — 1 passing
- React 19: `waapi-svg-zero-duration.ts` — 1 passing

The unchecked Cypress item in the original PR description is therefore superseded by these passing runs.

PR #3782, opened 2026-08-05
Fix Safari opacity flash after WAAPI animations

### LINKED ISSUE #3778: [BUG] iOS Safari: one-frame flash of underlying opacity at WAAPI animation completion (whileInView opacity + transform)
## Describe the bug

On iOS Safari, a `motion.div` animating `opacity` + `y` via `whileInView` shows a **one-frame flash at its underlying inline `opacity: 0`** at the moment the animation completes — the fully-revealed element visibly vanishes for an instant and reappears. Desktop browsers are unaffected.

Mechanism (from reading `motion-dom` 12.38.0): `opacity` is in `acceleratedValues`, so it runs as a native WAAPI animation, while `y` runs on the JS frameloop. The element's inline style still carries the pre-animation `opacity: 0` (SSR-emitted; never overwritten while WAAPI composites on top). At completion, the finish handoff does `setStyle(..., opacity: 1)` followed by `animation.cancel()`. On WebKit the style commit and the cancel can land in different frames — for one frame the canceled animation no longer composites but the committed style isn't painted yet, so the element paints at the underlying inline `opacity: 0`.

## Environment

- framer-motion `12.38.0` (`motion-dom` 12.38.0), Next.js 16 App Router (SSR emits the initial hidden style inline)
- Reproduces: iPhone, iOS Safari (owner-observed consistently across ~25 `whileInView` instances on one page; each element flashes once, right after its own reveal completes)
- Not reproducible: desktop Chrome/Chromium (same page)

## Repro sketch

```tsx
<motion.div
  initial={{ opacity: 0, y: 20 }}
  whileInView={{ opacity: 1, y: 0 }}
  viewport={{ once: true, margin: "0px 0px -50px 0px" }}
  transition={{ type: "spring", stiffness: 300, damping: 30 }}
>
  {content}
</motion.div>
```

Scroll elements into view on an iPhone; watch each block flash once immediately after its reveal settles. (SSR/inline `style="opacity:0"` at first paint appears to be a precondition — the underlying value the flash reveals.)

## Evidence trail

- Verified the WAAPI routing at runtime: during reveal, `document.getAnimations()` shows a native `Animation` with opacity keyframes targeting the element; completed elements rest at inline `opacity: 1; transform: none`.
- **Workaround confirmed on device:** passing any `onUpdate` to the component (`supportsBrowserAnimation()` gates on `!onUpdate`, `waapi.mjs`) forces opacity onto the JS path — the flash disappears completely with no other change.
- Compositor-side mitigations (`will-change: opacity`, `backface-visibility: hidden` on the animated node) do NOT fix it — tested on device, both failed, which fits the handoff-ordering explanation rather than layer teardown.

## Suggested direction

Commit the final keyframe style and flush/await the paint before `cancel()` on WebKit — or use `commitStyles()`-style ordering so there is no frame where neither the animation nor the committed style holds the final opacity. (Noted the WAAPI-commit hardening in 12.34.5/12.35.x changelogs; 12.38.0 still reproduces.)

Happy to provide more detail; I can't capture iOS screen recordings from this environment, but the on-device observation protocol above reproduces it reliably.

mattgperry:
## Summary
- commit the finished WAAPI effect to the element before writing the final keyframe and cancelling the animation
- prevent iOS Safari from briefly revealing the underlying initial opacity during the compositor-to-inline-style handoff
- add a regression test that verifies `commitStyles()` runs before `cancel()`

Fixes #3778

## Test plan
- [x] `yarn build`
- [x] `yarn turbo run test --filter=motion-dom` (496 passed, 3 skipped)
- [x] ESLint on both changed files (0 errors; existing boundary-type warnings only)
- [ ] `yarn lint` (blocked by pre-existing `dev/react` lint debt: 52 errors and 25 warnings)

The visual flash is specific to iOS Safari and cannot be reproduced in JSDOM or desktop browser CI, so the regression gate tests the underlying WAAPI handoff ordering directly.

### DIFF
--- a/packages/motion-dom/src/animation/NativeAnimation.ts
+++ b/packages/motion-dom/src/animation/NativeAnimation.ts
@@ -106,6 +106,8 @@ export class NativeAnimation<T extends AnyResolvedKeyframe>
                     this.updateMotionValue(keyframe)
                 }
 
+                this.commitStyles()
+
                 /**
                  * If we can, we want to commit the final style as set by the user,
                  * rather than the computed keyframe value supplied by the animation.
--- a/packages/motion-dom/src/animation/__tests__/NativeAnimation.test.ts
+++ b/packages/motion-dom/src/animation/__tests__/NativeAnimation.test.ts
@@ -2,12 +2,10 @@ import { motionValue } from "../../value"
 import { NativeAnimationExtended } from "../NativeAnimationExtended"
 
 /**
- * Tests for the Firefox opacity bug (issue #3552) where the WAAPI animation's
- * onfinish handler needed to commit the final style to the element's inline
- * styles before cancelling the animation. Without this, there was a timing
- * window in Firefox where the WAAPI fill was removed (via cancel()) before
- * the scheduled render could apply the correct value, causing a visual flash
- * back to the initial value.
+ * Tests for WAAPI completion bugs where cancelling the animation before its
+ * final styles are committed can briefly reveal the underlying initial style.
+ * Issue #3552 needed the final inline style set synchronously for Firefox,
+ * while issue #3778 needs the WAAPI styles committed before cancel for Safari.
  */
 describe("NativeAnimation - onfinish style commit", () => {
     let mockAnimation: any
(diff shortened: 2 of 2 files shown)


mattgperry:
Closing until we find a reproduction

PR #3795, opened 2026-08-17
Add renderer/style-recalc benchmarks, findings doc and layout profiler

mattgperry:
## Summary

- `dev/html/public/benchmarks`: interactive benchmarks comparing renderer strategies (inline style writes, adopted stylesheet, registered CSS variables, WAAPI, CSS transitions, GSAP) in benign and hostile style-recalculation environments, plus `style-recalc-harness.html`, which isolates which selector/environment patterns (`[style*=…]` attribute selectors, inherited properties, `@container style()` queries) turn a single style-attribute write into a full-page recalculation.
- `FINDINGS.md`: measured results and methodology — what triggers recalculation bombs, renderer comparisons, and measurement pitfalls (vsync floor masking real cost, rAF throttling in occluded tabs, tween semantics differences between libraries).
- `dev/react/profile-layout.mjs`: Playwright/CDP V8 sampling profiler for the layout stress examples, reporting projection-attributed JS time and per-function self-time hotspots. Works with both the current minimal `recordStats` and newer reporter-returning versions.
- `layout-stress-deep` example: deep-tree (depth 30 × 40 chains) layout animation stress test exercising per-ancestor projection cost.

No library code changes — dev/test tooling only.

## Test plan

- [x] Benchmarks and harness run in Chrome (results documented in FINDINGS.md)
- [x] Profiler smoke-tested against `layout-stress-deep` and `layout-stress-transform` on main

### DIFF
--- a/dev/html/public/benchmarks/FINDINGS.md
+++ b/dev/html/public/benchmarks/FINDINGS.md
@@ -1,0 +1,239 @@
+# Renderer & Style Invalidation Findings
+
+Results from the renderer investigation (Aug 2026): can a new write mechanism
+avoid the "massive style recalculation" failure mode of per-frame
+`element.style` writes, and what should Motion's renderer(s) be?
+
+Benchmark pages (serve `dev/html` on port 8000):
+
+- `style-recalc-harness.html` — which environments make style writes expensive,
+  across write mechanisms (inline, CSS vars, registered vars, adopted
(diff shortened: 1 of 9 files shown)


mattgperry:
Moved to the private motiondivision/motion-research repo (benchmarks made standalone via pinned esm.sh imports, findings updated with per-item attribution corrections). Keeping research artifacts out of library history.

